PR #20, opened 2020-08-13
/client/r0/keys/query: Ensure 400 on non-spec input

Kegsay:
Synapse allows this test case due to lack of type checking but
Dendrite rejects it. This form (or a variant of it) is used by
iOS clients.


anoadragon453:
https://github.com/vector-im/element-ios/issues/3539 is the iOS bug for reference.

PR #23, opened 2020-09-03
Use --ci flag in Dendrite generate-config

neilalexander:
This uses the new `--ci` flag in matrix-org/dendrite@28d62244 to produce sane defaults for testing rather than having to `sed` them.


Kegsay (approved):
Well isn't that fancy! Love it!

PR #28, opened 2020-09-21
Add CA generation and mount it to /ca in containers

valkum:
This PR adds support for CA generation in complement.

The Outbound Federation server will create a CA cert either in `$PWD/ca` or if run in CI in the Volume mounted to `/ca`.
On container deploy, depending on CI or not, either `$PWD/ca` is bind mounted to `/ca` or the `/ca` Volume of complement is mounted to `/ca` in homeserver containers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,10 +32,36 @@ $ COMPLEMENT_BASE_IMAGE=complement-dendrite:latest go test -timeout 30s -run '^(
 - The homeserver needs to `200 OK` requests to `GET /_matrix/client/versions`.
 - The homeserver needs to manage its own storage within the image.
 - The homeserver needs to accept the server name given by the environment variable `SERVER_NAME` at runtime.
+- The homeserver can use the CA certificate mounted at /ca to create its own TLS cert (see [Complement PKI](README.md#complement-pki)).
 
 #### Why 'Complement'?
 
 Because **M**<sup>*C*</sup> = **1** - **M**
+
+#### Complement PKI
+
+As the Matrix federation protocol expects federation endpoints to be served with valid TLS certs,
+Complement will create a self-signed CA cert to use for creating valid TLS certs in homeserver containers.
+
+To enabled it pass `COMPLEMENT_CA=true` to complement or the docker container.
+If not used, the homeserver needs to not validate the cert when federating.
+To check whether complements runs in PKI mode, `COMPLEMENT_CA` is passed through to the homeserver containers.
+
+The public key to add to the trusted cert store (e.g., /etc/ca-certificates) is mounted at: `/ca/ca.crt`
+The private key to sign the created TLS cert is mounted at: `/ca/ca.key`
+
+For example, to sign your certificate for the homeserver, run at each container start (Ubuntu):
+```
+openssl genrsa -out $SERVER_NAME.key 2048
+openssl req -new -sha256 -key $SERVER_NAME.key -subj "/C=US/ST=CA/O=MyOrg, Inc./CN=$SERVER_NAME" -out $SERVER_NAME.csr
+openssl x509 -req -in $SERVER_NAME.csr -CA /ca/ca.crt -CAkey /ca/ca.key -CAcreateserial -out $SERVER_NAME.crt -days 1 -sha256
+```
+
+To add the CA cert to your trust store (Ubuntu):
+```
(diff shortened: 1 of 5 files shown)


valkum:
Not sure if the outbound federation endpoint creation is the best place to create the CA cert. There is probably a better place, right before jumping into blueprints right?

We should specify how the lifetime is for those certs.
The CA should be generated on the start of complement (valid enough until the end of the run), and probably deleted afterwards or overwritten the next time complement starts.

Can someone explain to me how the HS containers get executed?
For each blueprint one or multiple containers get created. Are they then commited?
If I get it right: 
If the container creates the local cert during RUN or ENTRYPOINT they would need to check whether the cert got already created during a previous run of that container (small optimization to not create a new cert everytime the container is run). Right?

I will add documentation when we got those specifics figured out.

Kegsay (approved):
LGTM

I was debating whether we should be gating this behind an env var but for now let's see how bad the performance hit is.

internal/docker/builder.go
```diff
@@ -300,6 +304,39 @@ func deployImage(docker *client.Client, imageID string, csPort int, containerNam
 		// change this to be  `host.docker.internal:host-gateway`
 		extraHosts = []string{HostnameRunningComplement + ":172.17.0.1"}
 	}
+
+	var caVolume map[string]struct{}
+	var caMount []mount.Mount
+	if os.Getenv("CI") == "true" {
```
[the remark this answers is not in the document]
valkum: Got time to finish this.
Changed and tested the implementation.
Added prose at appropriate places to explain what and why we are doing things.

valkum:
Some things changed while further testing this. Please reapprove, to be safe.
I fixed the logic behind mounting the Volume of the complement container when run in CI.
In addtion, I made sure the certs get created at the beginning of the tests, and we keep only one instance of them.

internal/docker/builder.go
```diff
@@ -291,6 +295,80 @@ func (d *Builder) deployBaseImage(blueprintName, hsName, contextStr, networkID s
...
+		for i := range container.Mounts {
+			if container.Mounts[i].Destination == "/ca" {
+				volumeName = container.Mounts[i].Name
+			}
+		}
+		if volumeName == "" {
+			// We did not find a volume. This container might be created without a volume,
+			// or CI=true is passed but we are not running in a container.
+			// todo: log that we do not provide a CA volume mount?
+			return map[string]struct{}{}, []mount.Mount{}, nil
```
Kegsay, line 334: `return nil, nil, nil` may be more appropriate?

internal/federation/server.go
```diff
@@ -195,6 +197,116 @@ func (s *Server) Listen() (cancel func()) {
 	}
 }
 
+// Get or create local CA cert. This is used to create the federation tls cert.
+// In addtion, it is passed to homesever containers to create tls certs
```
Kegsay, line 201:
```suggestion
// In addition, it is passed to homesever containers to create tls certs 
```

Kegsay (commented):
I'd like to see this be gated behind something like `COMPLEMENT_CA=1` because it adds a fair amount of extra complexity which isn't needed for most HSes which support disabling TLS validation. I'll run a few more tests with Dendrite and then see if there is a big perf hit or not.

valkum:
I assume that is enough to feature gate all of the PKI stuff.
Oh and I committed the README changes. Thought I included them in the last commit. My bad.

README.md
```diff
@@ -32,11 +32,37 @@ $ COMPLEMENT_BASE_IMAGE=complement-dendrite:latest go test -timeout 30s -run '^(
...
 #### Why 'Complement'?
 
 Because **M**<sup>*C*</sup> = **1** - **M**
 
+#### Complement PKI
+
+As the Matrix federation protocol expects federation endpoints to be served with valid TLS certs,
+Complement will create a self-signed CA cert to use for creating valid TLS certs in homeserver containers.
+
+To enabled it pass `COMPLEMENT_CA=true` to complement or the docker container.
```
Kegsay, line 46:
```suggestion
To enable it pass `COMPLEMENT_CA=true` to complement or the docker container.
```

Kegsay (approved):

Kegsay:
Fab, thanks! LGTM

PR #37, opened 2020-11-09
Add request query parameter map to instruction struct

anoadragon453:
This allows passing query parameters to a request while executing an instruction.

This will be useful for e.g sending `server_name` query parameters while attempting to join a room.

I've confirmed that Dendrite still passes all tests even with this change.

### DIFF
--- a/internal/instruction/runner.go
+++ b/internal/instruction/runner.go
@@ -109,11 +109,14 @@ func (r *Runner) next(instrs []instruction, hsURL string, i int) (*http.Request,
 		r.log("Stopping. Failed to form NewRequest for instruction: %s -- %+v \n", err, instr)
 		return nil, nil, 0
 	}
+	q := req.URL.Query()
 	if instr.accessToken != "" {
-		q := req.URL.Query()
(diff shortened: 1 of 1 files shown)


Kegsay:
I don't see the use case here. Instructions are for pre-configuring a HS, which allows referencing via homeserver name e.g `hs1`. If we want to test `?server_name=` then we'd need to write code for it, it wouldn't be done as a blueprint.

anoadragon453:
@Kegsay This came out of https://github.com/matrix-org/complement/issues/38 and wanting to stick a `server_name` on the `/join/!room:id` call that happens during a blueprint setup: https://github.com/matrix-org/complement/blob/cfb058d8d5f26289debd4c59d4f8ac2e55191657/internal/instruction/runner.go#L227-L234

If the homeserver doesn't support parsing room IDs for domains then this command will fail.

Perhaps there's a better way to achieve this though?

Kegsay:
Oh I see, is extracting the domain from the room ID non-spec then? I guess we'd need to add in some form of server name to the blueprint then yes.

Kegsay (approved):

PR #40, opened 2020-11-17
Add CSAPI.MustDoWithStatus(Raw) method

anoadragon453:
These methods are analogous with their `CSAPI.MustDo` equivalents other than that they require an expected status code to be specified.

This is a convenience to be used when making calls that we expect with return a non-2xx HTTP status code.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -166,7 +166,7 @@ func (c *CSAPI) MustDoWithStatusRaw(t *testing.T, method string, paths []string,
 		t.Fatalf("CSAPI.MustDoWithStatusRaw %s %s error: %s", method, strings.Join(paths, "/"), err)
 	}
 	if res.StatusCode != status {
-		t.Fatalf("CSAPI.MustDoWithStatusRaw %s %s returned HTTP %d", method, res.Request.URL.String(), res.StatusCode)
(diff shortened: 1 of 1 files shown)


internal/client/client.go
```diff
@@ -145,6 +145,32 @@ func (c *CSAPI) syncUntil(t *testing.T, since, key string, check func(gjson.Resu
...
+
+// MustDoWithStatusRaw is the same as MustDoRaw but fails the test if the response code does not match that provided
+func (c *CSAPI) MustDoWithStatusRaw(t *testing.T, method string, paths []string, body []byte, contentType string, query url.Values, status int) *http.Response {
+	t.Helper()
+	res, err := c.DoWithAuthRaw(t, method, paths, body, contentType, query)
+	if err != nil {
+		t.Fatalf("CSAPI.MustDoWithStatusRaw %s %s error: %s", method, strings.Join(paths, "/"), err)
+	}
+	if res.StatusCode != status {
+		t.Fatalf("CSAPI.MustDoWithStatusRaw %s %s returned HTTP %d", method, res.Request.URL.String(), res.StatusCode)
```
Kegsay, line 169: Please log what was expected as well.

Kegsay (approved):

PR #42, opened 2020-11-26
Abort test if the blueprint construction failed

### LINKED ISSUE #41: Error handling in constructHomeserver could be unified
Considering this code (with 2 places where an `err != nil` can happen):

https://github.com/matrix-org/complement/blob/9acaf3cd28e2d32024c80a3f04105e2cce7ff1e3/internal/docker/builder.go#L269-L284

If the deploy fails (first `err != nil`) it gets logged to the standard output via `log.Printf` (and `printLogs` is called explicitely).
If the runner fails, it gets logged to the Builder (and `printLogs` is called by the parent).

I think those 2 error logging should be unified: what do you think is the best approach?

I would recommend having some logging to the standard output, since it can impact the rest of the test.

For instance adding an `err` param to `printLogs`, which will get printed at the end of the server log if not `nil`, called inside `construct`.

I would be glad to make a PR for this.

---

Background info: I am trying to understand how #39 can be fixed, and I discovered that the Blueprints has an error (but only after spending a couple of hours and finally adding `COMPLEMENT_DEBUG=1` :)
```
2020/11/17 22:57:42 federation_one_to_one_room.hs2 : failed to run instructions: federation_one_to_one_room.hs2 : request http://localhost:32902/_matrix/client/r0/join/%21IiiKfQGoJGEqTPCTXN:hs1?access_token=MDAxMWxvY2F0aW9uIGhzMgowMDEzaWRlbnRpZmllciBrZXkKMDAxMGNpZCBnZW4gPSAxCjAwMWJjaWQgdXNlcl9pZCA9IEBib2I6aHMyCjAwMTZjaWQgdHlwZSA9IGFjY2VzcwowMDIxY2lkIG5vbmNlID0gbjNjbkNJeUs7bmxhMFBubAowMDJmc2lnbmF0dXJlIGjnOvc29ANDh6mwUukBi_1NocFirh6_gE1ZoogwQotxCg returned HTTP 404 Not Found : {"errcode":"M_UNKNOWN","error":"No known servers"}
```

oliverpool:
Closes #41

Since multiple errors can happen when constructing the blueprint, the `ConstructBlueprints` will log all of them and return the first.

### DIFF
--- a/internal/docker/builder.go
+++ b/internal/docker/builder.go
@@ -188,6 +188,7 @@ func (d *Builder) ConstructBlueprints(bs []b.Blueprint) error {
 	for i := 0; i < len(bs); i++ {
 		errs = append(errs, <-errc...)
 	}
+	close(errc)
 	if len(errs) > 0 {
 		for _, err := range errs {
 			d.log("could not construct blueprint: %s", err)
@@ -233,8 +234,10 @@ func (d *Builder) construct(bprint b.Blueprint) (errs []error) {
 		res := d.constructHomeserver(bprint.Name, runner, hs, networkID)
 		if res.err != nil {
 			errs = append(errs, res.err)
-			// print docker logs because something went wrong
-			printLogs(d.Docker, res.containerID, res.contextStr)
+			if res.containerID != "" {
+				// something went wrong, but we have a container which may have interesting logs
+				printLogs(d.Docker, res.containerID, res.contextStr)
+			}
 		}
 		// kill the container
 		defer func(r result) {
@@ -488,7 +491,7 @@ func createNetwork(docker *client.Client, blueprintName string) (networkID strin
 	}
 	if nw.Warning != "" {
 		if nw.ID == "" {
-			return "", fmt.Errorf("%s: fatal warning while creating docker network. %w", blueprintName, err)
+			return "", fmt.Errorf("%s: fatal warning while creating docker network. %s", blueprintName, nw.Warning)
 		}
 		log.Printf("WARNING: %s\n", nw.Warning)
 	}
--- a/internal/docker/builder.go
+++ b/internal/docker/builder.go
@@ -186,6 +186,9 @@ func (d *Builder) ConstructBlueprints(bs []b.Blueprint) error {
 	}
 	var errs []error
 	for i := 0; i < len(bs); i++ {
+		// the channel returns a slice of errors;
+		// spread and append them to the error slice
+		// (nothing will be appended if the slice is empty)
 		errs = append(errs, <-errc...)
 	}
 	close(errc)
--- a/internal/docker/builder.go
+++ b/internal/docker/builder.go
@@ -17,14 +17,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"log"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"os"
 	"path"
 	"runtime"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/docker/docker/api/types"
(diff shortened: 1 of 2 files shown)


internal/docker/builder.go
```diff
@@ -179,12 +178,23 @@ func (d *Builder) ConstructBlueprintsIfNotExist(bs []b.Blueprint) error {
...
 	for _, bprint := range bs {
-		go d.construct(bprint, &bpWg)
+		go (func(bprint b.Blueprint) {
+			errc <- d.construct(bprint)
+		})(bprint)
 	}
-	bpWg.Wait()
+	var errs []error
+	for i := 0; i < len(bs); i++ {
+		errs = append(errs, <-errc...)
```
Kegsay, line 189: We don't ever close `errc` so we'll leak resources. Call `close(errc)` after this `for` loop.

internal/docker/builder.go
```diff
@@ -179,12 +178,23 @@ func (d *Builder) ConstructBlueprintsIfNotExist(bs []b.Blueprint) error {
...
+	}
+	if len(errs) > 0 {
```
Kegsay, line 191: Is this really the right check? Won't `errs` just be `[nil, nil, nil, ...]` in the happy path?

internal/docker/builder.go
```diff
@@ -211,18 +221,18 @@ func (d *Builder) ConstructBlueprints(bs []b.Blueprint) error {
...
+	if err != nil {
+		return []error{err}
 	}
 
 	runner := instruction.NewRunner(bprint.Name, d.debugLogging)
 	results := make([]result, len(bprint.Homeservers))
 	for i, hs := range bprint.Homeservers {
 		res := d.constructHomeserver(bprint.Name, runner, hs, networkID)
-		if res.err != nil && res.containerID != "" {
+		if res.err != nil {
```
Kegsay, line 234: We still presumably want to check `if res.containerID != ""` before calling `printLogs` else if `constructHomeserver` failed before the container was made, there will be no logs to print.

internal/docker/builder.go
```diff
@@ -211,18 +221,18 @@ func (d *Builder) ConstructBlueprints(bs []b.Blueprint) error {
 }
 
 // construct all Homeservers sequentially then commits them
-func (d *Builder) construct(bprint b.Blueprint, bpWg *sync.WaitGroup) {
-	defer bpWg.Done()
-	networkID := createNetwork(d.Docker, bprint.Name)
-	if networkID == "" {
-		return
+func (d *Builder) construct(bprint b.Blueprint) (errs []error) {
```
Kegsay, line 224: I'd just have `(err error)` as the response and fail-fast and bail out of this function the moment anything goes wrong. We don't need to aggregate errors from different homeserver images.

internal/docker/builder.go
```diff
@@ -468,13 +484,18 @@ func createNetwork(docker *client.Client, blueprintName string) (networkID strin
 		},
 	})
 	if err != nil {
-		log.Printf("Failed to create docker network: %s\n", err)
-		return ""
+		return "", fmt.Errorf("%s: failed to create docker network. %w", blueprintName, err)
 	}
 	if nw.Warning != "" {
+		if nw.ID == "" {
+			return "", fmt.Errorf("%s: fatal warning while creating docker network. %w", blueprintName, err)
```
Kegsay, line 491: `err` will be `nil` here, you probably meant to log `nw.Warning` instead.

Kegsay (changes requested):
Few things.

internal/docker/builder.go
oliverpool: Done!

internal/docker/builder.go
oliverpool: Done!

internal/docker/builder.go
oliverpool: Good catch!

internal/docker/builder.go
oliverpool: I am appending `<-errc...` to `errs`, so if the returned error array (from the channel) is empty, nothing will be appended and the length of `errs` will not change.

internal/docker/builder.go
oliverpool: I think it is better to try everything and then informing the user of all the failures (instead to have only one failure at a time, which can be very time consuming to debug).

oliverpool: But if you prefer I can change this to fail early and ignore the rest.

Kegsay: I don't have strong opinions either way. I generally lean on the side of "what makes the code simpler" than "what presents the most information to the user" when it comes to internal guts. For tests themselves then it's the other way around. I don't mind.

internal/docker/builder.go
Kegsay: It's not super clear this is the case. Please add a comment to the append.

Kegsay (changes requested):
Please add the comment then LGTM! Thanks!

internal/docker/builder.go
oliverpool: I added a comment, is it explicit enough?

Kegsay (approved):

PR #44, opened 2020-11-27
Add tests for knocking

anoadragon453:
This PR adds tests for knocking. They are currently hidden behind a build tag. To run the tests, you'll need to add `-tags msc2403` to your command line args to include these tests. The build flag, as well as all of the `xyz.amorgan.knock` references should be remove/modified after MSC2403 makes it into a spec version.

* MSC: https://github.com/matrix-org/matrix-doc/pull/2403
* Synapse implementation: https://github.com/matrix-org/synapse/pull/6739


internal/client/client.go
```diff
@@ -101,6 +101,14 @@ func (c *CSAPI) SyncUntilTimelineHas(t *testing.T, roomID string, check func(gjs
 	c.syncUntil(t, "", "rooms.join."+gjsonEscape(roomID)+".timeline.events", check)
 }
 
+// SyncUntilKnockHas blocks and continually calls /sync until the `check` function returns true.
+// If the `check` function fails the test, the failing event will be automatically logged.
+// Will time out after CSAPI.SyncUntilTimeout.
+func (c *CSAPI) SyncUntilKnockHas(t *testing.T, roomID string, check func(gjson.Result) bool) {
```
Kegsay, line 107: I don't think knocking is _that_ common that it should pollute the `CSAPI` client. Move this to the test file please.

Kegsay: If you require `syncUntil` logic, make that public instead.

tests/knocking_test.go
```diff
@@ -0,0 +1,194 @@
+// +build knock
```
Kegsay, line 1: Prefer the MSC number please, as feature names are ambiguous (particularly if there are multiple competing MSCs for the same feature.

Kegsay: Also please add the MSC number to the filename, e.g like https://github.com/matrix-org/complement/blob/kegan/msc2836/tests/msc2836_test.go

anoadragon453: Good point, much more scalable indeed.

tests/msc2403_test.go
```diff
@@ -0,0 +1,241 @@
...
+					// We don't current define any required state event types to be sent
+					// If we've reached this point, then an entry for this room was found
+					return true
+				},
+			)
+		})
+	})
+
+	if !federation {
+		// Rescinding a knock over federation is currently not supported in Synapse
```
clokep, line 136: Isn't it not even defined in the spec?

tests/msc2403_test.go
```diff
@@ -0,0 +1,241 @@
...
+				"Seems like a trustworthy fellow",
+			},
+		)
+	})
+
+	t.Run("A user that is banned from a room cannot knock on it", func(t *testing.T) {
+		inRoomUser.MustDo(
+			t,
+			"POST",
+			[]string{"_matrix", "client", "r0", "rooms", roomID, "invite"},
```
clokep, line 194: Is this a copy and paste error? Shouldn't the user be being banned, not invited?

tests/msc2403_test.go
```diff
@@ -0,0 +1,241 @@
...
+				knockingUserUserID,
+				"I don't think so",
+			},
+		)
+
+		// Knock again
+		knockOnRoom(t, knockingUser, roomID, "Pleeease let me in?", []string{"hs1"})
+	})
+
+	t.Run("A user in the room can accept a knock", func(t *testing.T) {
```
clokep, line 175: I think a knock after this isn't allowed and should 403?

tests/msc2403_test.go
anoadragon453: *Rejecting* a knock over federation is currently not possible, but rescinding a knock over federation can work (implementations just need to figure out which homeserver to send the leave event through).

https://github.com/matrix-org/matrix-doc/pull/2403/files#diff-9dcacf9adb620c5346eba2f70f71bd50ed4f85a5e2844ac130e112165ce35c06R118-R119

tests/msc2403_test.go
anoadragon453: Err, yes. This is why it's handy to have different error codes per situation :(

tests/msc2403_test.go
anoadragon453: True! I'll stick in a test for that, thanks!

anoadragon453:
@Kegsay heads up I'm looking for a review from your side as well just to take a look over whether these tests were written in a way you think works with Complement :)

tests/msc2403_test.go
```diff
@@ -0,0 +1,281 @@
...
+			},
+		)
+	})
+
+	t.Run("A user cannot knock on a room they are already in", func(t *testing.T) {
+		reason := "I'm sticking my hand out the window and knocking again!"
+		knockOnRoomWithStatus(t, knockingUser, roomID, reason, []string{"hs1"}, 403)
+	})
+
+	t.Run("A user that is banned from a room cannot knock on it", func(t *testing.T) {
```
clokep, line 224: Is this really testing what we think it is or is it testing that the user is already in the room? (Or does banning also kick the user?)

tests/msc2403_test.go
```diff
@@ -0,0 +1,281 @@
...
+	charlieUserID := "@charlie:hs2"
+	charlie := deployment.Client(t, "hs2", charlieUserID)
+
+	// Create a room for alice and bob to test knocking with
+	roomIDOne := alice.CreateRoom(t, struct {
+		Preset      string `json:"preset"`
+		RoomVersion string `json:"room_version"`
+	}{
+		"private_chat",      // Set to private in order to get an invite-only room
+		"xyz.amorgan.knock", // Room version required for knocking. TODO: Remove when knocking is in a stable room version
```
clokep, line 44: What do we do when this is standardized? Do we just update these? Do we somehow let a server specify that they support the MSC vs. the standardized version?

clokep (approved):
I think the logic looks sound! I didn't closely look at the changes to the blueprint/client.

tests/msc2403_test.go
anoadragon453: Banning will kick a user, yes. It updates their membership to `ban`, removing them from the room.

tests/msc2403_test.go
anoadragon453: Once the MSC is merged into a released version of the spec, Complement should be updated to support that release. Part of that update will be stabilizing knocks.

The necessary steps for stabilizing will be:

* Updating instances of `xyz.amorgan.knock` to `knock`.
* Update endpoints from `/_matrix/client/unstable/xyz.amorgan.knock/...` to `/_matrix/client/rX/...`.
* Removing the `// +build ...` line at the top so the tests are run by default (they're currently opt in).
* Updating the name of the test file to `knock_test.go` or somesuch.
* Removing the conditional on rescinding knocks over federation. Synapse will need to support that if the spec says it's possible.

It might be a good idea to put the above list into the top of the file actually.

tests/msc2403_test.go
```diff
@@ -0,0 +1,281 @@
...
+	requestBody := struct {
+		Reason string `json:"reason"`
+	}{
+		// We specify a reason here instead of using the same one each time as implementations can
+		// cache responses to identical requests
+		reason,
+	}
+	b, err := json.Marshal(requestBody)
+	if err != nil {
+		t.Fatalf("knockOnRoomWithStatus failed to marshal JSON body: %s", err)
```
Kegsay, line 263: You can use `must.NotError(t, msg, err)`

tests/msc2403_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+	t.Run("Knocking on a room with a join rule other than 'knock' should fail", func(t *testing.T) {
+		knockOnRoomWithStatus(t, knockingUser, roomID, "Can I knock anyways?", []string{"hs1"}, 403)
+	})
+
+	t.Run("Change the join rule of a room from 'invite' to 'xyz.amorgan.knock'", func(t *testing.T) {
+		inRoomUser.MustDo(
+			t,
+			"PUT",
+			[]string{"_matrix", "client", "r0", "rooms", roomID, "state", "m.room.join_rules", ""},
```
Kegsay, line 74: Don't do this, use `CSAPI.SendEventSynced(t, roomID, b.Event)` to send state events into the room. This will wait for the response to come down `/sync` which is critical to reduce racey tests.

tests/msc2403_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+	t.Run("parallel", func(t *testing.T) {
+		t.Run("Users in the room see a user's membership update when they knock", func(t *testing.T) {
+			t.Parallel()
+			inRoomUser.SyncUntilTimelineHas(t, roomID, func(ev gjson.Result) bool {
+				if ev.Get("type").Str != "m.room.member" || ev.Get("sender").Str != knockingUserID {
+					return false
+				}
+				must.EqualStr(t, ev.Get("content").Get("reason").Str, testKnockReason, "incorrect reason for knock")
+				must.EqualStr(t, ev.Get("content").Get("membership").Str, "xyz.amorgan.knock", "incorrect membership for knocking user")
```
Kegsay, line 118: I would make `const constKnock = "xyz.amorgan.knock"` and use that, which will also help for reducing the number of changes when it lands.

tests/msc2403_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+	if !federation {
+		// Rescinding a knock over federation is currently not supported in Synapse
+		// TODO: Preventing another homeserver from running this test just because Synapse cannot is unfortunate
+		t.Run("A user that has knocked on a local room can rescind their knock and then knock again", func(t *testing.T) {
+			// Rescind knock
+			knockingUser.MustDo(
+				t,
+				"POST",
+				[]string{"_matrix", "client", "r0", "rooms", roomID, "leave"},
```
Kegsay, line 146: You should probably wait for the leave to come down sync.

tests/msc2403_test.go
```diff
@@ -0,0 +1,281 @@
...
+			knockOnRoom(t, knockingUser, roomID, "Let me in... again?", []string{"hs1"})
+		})
+	}
+
+	t.Run("A user in the room can reject a knock", func(t *testing.T) {
+		// Reject the knock
+		inRoomUser.MustDo(
+			t,
+			"POST",
+			[]string{"_matrix", "client", "r0", "rooms", roomID, "kick"},
```
Kegsay, line 164: Wait for it to come down sync before knocking again.

tests/msc2403_test.go
```diff
@@ -0,0 +1,281 @@
...
+	t.Run("A user cannot knock on a room they are already in", func(t *testing.T) {
+		reason := "I'm sticking my hand out the window and knocking again!"
+		knockOnRoomWithStatus(t, knockingUser, roomID, reason, []string{"hs1"}, 403)
+	})
+
+	t.Run("A user that is banned from a room cannot knock on it", func(t *testing.T) {
+		inRoomUser.MustDo(
+			t,
+			"POST",
+			[]string{"_matrix", "client", "r0", "rooms", roomID, "ban"},
```
Kegsay, line 228: Wait for it to come down sync.

Kegsay (changes requested):
Few things but broadly LGTM

tests/msc2403_test.go
clokep: I guess my broader question is -- does complement have a way for homeservers to say what version of the spec they support? Not expecting it here necessarily, just bringing it up for conversation!

And if it does, does it make sense for a homeserver to say it supports 0.6 + msc2403 even after 0.7 is released?

anoadragon453: @Kegsay I'd be interested in your answer here in case it differs from mine.

I don't think it does, and in doing so I think Complement would need to support multiple spec versions, which I don't think it's planning to. Mostly because new spec versions often require modifying existing tests instead of making a separate set of new ones.

Selecting a range of tests is done via build tags, thus I think this choice would be made in your individual CI script rather than in a config file in the server/complement (at least currently!).

clokep: Alright. I don't think we need to solve this here, was just a thought I had that!

anoadragon453:
@Kegsay quite a few chunky changes since your last review, all addressing your feedback. Details are in commit messages, but overall it was mostly a lot of code for syncing, part of which required some slight changes to client.go.

Kegsay (approved):
We'll probably want to factor out the leave synced stuff so we don't need to hit initial sync so much but for now this will do. LGTM!

PR #47, opened 2020-12-04
Add a handler for federation transactions

anoadragon453:
Add a handler which can accept a federation transaction, validate the PDUs and return errors in the case of failing to process a given PDU.

EDUs are accepted but we currently don't do anything with them.

This grew out of trying to fix `TestOutboundFederationIgnoresMissingEventWithBadJSONForRoomVersion6` for Synapse. The test would fail because as soon as Synapse joined the room, it would start trying to send presence EDUs to the new server. Rather than disabling presence, I thought it would be better to write a proper handler for it.

I'll use this handler for that test (and possibly others) in a separate PR.

### DIFF
--- a/internal/federation/handle.go
+++ b/internal/federation/handle.go
@@ -5,11 +5,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/util"
 )
 
 // MakeJoinRequestsHandler is the http.Handler implementation for the make_join part of
@@ -161,7 +163,7 @@ func HandleDirectoryLookups() func(*Server) {
 	}
 }
 
-// HandleEventRequests is an optino which will process GET /_matrix/federation/v1/event/{eventId} requests universally when requested.
+// HandleEventRequests is an option which will process GET /_matrix/federation/v1/event/{eventId} requests universally when requested.
 func HandleEventRequests() func(*Server) {
 	return func(srv *Server) {
 		srv.mux.Handle("/_matrix/federation/v1/event/{eventID}", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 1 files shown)


anoadragon453:
I've just realised that a proto version of this exists here: https://github.com/matrix-org/complement/blob/a7d6117bf325b61414481d20838cc4c18edc1a23/tests/federation_room_send_test.go#L45-L65, but perhaps this change is better still?

internal/federation/handle.go
```diff
@@ -277,3 +279,101 @@ func HandleMediaRequests(mediaIds map[string]func(w http.ResponseWriter)) func(*
...
+					RoomID string `json:"room_id"`
+				}
+				if err := json.Unmarshal(pdu, &header); err != nil {
+					log.Printf("Transaction '%s': Failed to extract room ID from event", transaction.TransactionID)
+					// We don't know the event ID at this point so we can't return the
+					// failure in the PDU results
+				}
+
+				// Retrieve the room version from the server
+				room := srv.rooms[header.RoomID]
```
Kegsay, line 347: Need to handle if `room == nil`

internal/federation/handle.go
```diff
@@ -277,3 +279,101 @@ func HandleMediaRequests(mediaIds map[string]func(w http.ResponseWriter)) func(*
...
+				roomVersion := gomatrixserverlib.RoomVersion(room.Version)
+
+				event, err := gomatrixserverlib.NewEventFromUntrustedJSON(pdu, roomVersion)
+				if err != nil {
+					// We were unable to verify or process this event.
+					// Add this PDU to the response along with an error message.
+					response.PDUs[event.EventID()] = gomatrixserverlib.PDUResult{
```
Kegsay, line 354: This isn't correct, we don't return errors here. The logic for "when do we tell the sender we encountered an error" is _horrible_ for `/send` - see https://github.com/matrix-org/dendrite/blob/master/federationapi/routing/send.go#L151 - for now I would say never return an error at all ever. This is just a helper function to populate the room timeline after all.

internal/federation/handle.go
```diff
@@ -277,3 +279,101 @@ func HandleMediaRequests(mediaIds map[string]func(w http.ResponseWriter)) func(*
...
+						Error: err.Error(),
+					}
+					continue
+				}
+
+				// Store this PDU in the room's timeline
+				room.Timeline = append(room.Timeline, &event)
```
Kegsay, line 361: Need to update state too. Just call https://github.com/matrix-org/complement/blob/master/internal/federation/server_room.go#L23

internal/federation/handle.go
```diff
@@ -277,3 +279,101 @@ func HandleMediaRequests(mediaIds map[string]func(w http.ResponseWriter)) func(*
 		mediamux.Handle("/r0/download/{origin}/{mediaId}", downloadFn).Methods("GET")
 	}
 }
+
+// HandleTransactionRequests is an option which will process GET /_matrix/federation/v1/send/{transactionID} requests universally when requested.
+func HandleTransactionRequests() func(*Server) {
```
Kegsay, line 284: Might be cute to have a callback function here which is passed with valid `gmsl.Event` structs, since it's fairly common to want to process stuff on `/send`. E.g:
```go
func HandleTransactionRequests(cb func(gmsl.Event)) func(*Server) {
    // when ready..
    cb(event)
}
```

Kegsay (changes requested):
Nil check is important.

internal/federation/handle.go
anoadragon453: This is what I was thinking, yeah. Then we could perhaps augment `TestOutboundFederationSend` to use this instead, while slimming down the test at the same time.

internal/federation/handle.go
anoadragon453: Will do. Could you explain why a room would be nil if we have the ID for it?

internal/federation/handle.go
anoadragon453: Shall I add another callback method for EDUs?

internal/federation/handle.go
anoadragon453: Oh goodness, OK. Didn't realise v6 had introduced that woe.

anoadragon453:
Realised there was a bug in that we were reading from `req.Body` which had already been read from during `gmsl.VerifyHTTPRequest`. So now we read from `fedReq.Content()` instead.

Additionally I added `nil` checks for the pdu and edu handler functions.

internal/federation/handle.go
Kegsay: Complement should not assume that HSes are going to send valid traffic to it. The room ID is untrusted input (from a random HS), so it's entirely possible that it is referring to a room which Complement is unaware of.

internal/federation/handle.go
```diff
@@ -277,3 +279,105 @@ func HandleMediaRequests(mediaIds map[string]func(w http.ResponseWriter)) func(*
...
+				w.Write(b)
+				return
+			}
+
+			// Unmarshal the request body into a transaction object
+			var transaction gomatrixserverlib.Transaction
+			err := json.Unmarshal(fedReq.Content(), &transaction)
+			if err != nil {
+				log.Printf("Unable to unmarshal transaction body bytes into Transaction object: %s", err.Error())
+				errResp := util.MessageResponse(400, err.Error())
```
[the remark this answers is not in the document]
Kegsay: Applies to `:319` and `:294` too.

internal/federation/handle.go
```diff
@@ -277,3 +279,105 @@ func HandleMediaRequests(mediaIds map[string]func(w http.ResponseWriter)) func(*
...
+			response.PDUs = make(map[string]gomatrixserverlib.PDUResult, 0)
+			for _, pdu := range transaction.PDUs {
+				var header struct {
+					RoomID string `json:"room_id"`
+				}
+				if err := json.Unmarshal(pdu, &header); err != nil {
+					log.Printf("Transaction '%s': Failed to extract room ID from event", transaction.TransactionID)
+					// We don't know the event ID at this point so we can't return the
+					// failure in the PDU results
+				}
```
Kegsay, line 337: Probably should `continue` rather than relying on the `continue` at `:342`.

internal/federation/handle.go
```diff
@@ -277,3 +279,105 @@ func HandleMediaRequests(mediaIds map[string]func(w http.ResponseWriter)) func(*
...
+
+			// Construct a response and fill as we process each PDU
+			response := gomatrixserverlib.RespSend{}
+			response.PDUs = make(map[string]gomatrixserverlib.PDUResult, 0)
+			for _, pdu := range transaction.PDUs {
+				var header struct {
+					RoomID string `json:"room_id"`
+				}
+				if err := json.Unmarshal(pdu, &header); err != nil {
+					log.Printf("Transaction '%s': Failed to extract room ID from event", transaction.TransactionID)
```
Kegsay, line 334: Log the error message please.

internal/federation/handle.go
```diff
@@ -277,3 +279,105 @@ func HandleMediaRequests(mediaIds map[string]func(w http.ResponseWriter)) func(*
...
+					continue
+				}
+				roomVersion := gomatrixserverlib.RoomVersion(room.Version)
+
+				event, err := gomatrixserverlib.NewEventFromUntrustedJSON(pdu, roomVersion)
+				if err != nil {
+					// We were unable to verify or process this event.
+					// Tell the sending server that we accepted it anyways, as we're just looking to populate the timeline here
+					log.Printf("Unable to process event '%s' in transaction '%s': %s", event.EventID(), transaction.TransactionID, err.Error())
+				}
```
Kegsay, line 351: Definitely need to `continue` here and not proceed else we'll panic on `:354`.

Kegsay (changes requested):
Logging needs to be more explicit, and the panic on error needs to be fixed, otherwise LGTM

internal/federation/handle.go
anoadragon453: Derp, good point.

internal/federation/handle.go
```diff
@@ -277,3 +279,128 @@ func HandleMediaRequests(mediaIds map[string]func(w http.ResponseWriter)) func(*
 		mediamux.Handle("/r0/download/{origin}/{mediaId}", downloadFn).Methods("GET")
 	}
 }
+
+// HandleTransactionRequests is an option which will process GET /_matrix/federation/v1/send/{transactionID} requests universally when requested.
+// eventCallback is a function that will be called and passed each event that is received in the transaction
```
Kegsay, line 284: Comments lie.

internal/federation/handle.go
```diff
@@ -277,3 +279,128 @@ func HandleMediaRequests(mediaIds map[string]func(w http.ResponseWriter)) func(*
...
+				return
+			}
+			transaction.TransactionID = gomatrixserverlib.TransactionID(transactionID)
+
+			// Transactions are limited in size; they can have at most 50 PDUs and 100 EDUs.
+			// https://matrix.org/docs/spec/server_server/latest#transactions
+			if len(transaction.PDUs) > 50 || len(transaction.EDUs) > 100 {
+				log.Printf(
+					"complement: Transaction '%s': Transaction too large. PDUs: %d/50, EDUs: %d/100",
+					transaction.TransactionID, (transaction.PDUs), len(transaction.EDUs),
```
Kegsay, line 330: Missing `len`

Kegsay (approved):

Kegsay (approved):

PR #48, opened 2020-12-07
Update tests to use the new federation transaction handler

anoadragon453:
Updates the following tests to use the new transaction handler introduced in #47:

* `TestOutboundFederationSend` - This test failed with Synapse with the inline transaction handler. Not entirely sure why, the handler function was just never called.
* `TestOutboundFederationIgnoresMissingEventWithBadJSONForRoomVersion6` - Synapse will attempt to send presence to the remote homeserver. The test would fail if not expecting this.

Based on ~~#47.~~

### DIFF
--- a/tests/federation_room_get_missing_events_test.go
+++ b/tests/federation_room_get_missing_events_test.go
@@ -34,6 +34,8 @@ func TestOutboundFederationIgnoresMissingEventWithBadJSONForRoomVersion6(t *test
 	srv := federation.NewServer(t, deployment,
 		federation.HandleKeyRequests(),
 		federation.HandleMakeSendJoinRequests(),
+		// Handle any transactions that the homeserver may send when connecting to another homeserver (such as presence)
+		federation.HandleTransactionRequests(nil, nil),
 	)
 	cancel := srv.Listen()
 	defer cancel()
--- a/tests/federation_room_send_test.go
+++ b/tests/federation_room_send_test.go
@@ -1,12 +1,11 @@
 package tests
 
 import (
-	"encoding/json"
-	"net/http"
 	"testing"
 	"time"
 
 	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/docker"
 	"github.com/matrix-org/complement/internal/federation"
 	"github.com/matrix-org/gomatrixserverlib"
 )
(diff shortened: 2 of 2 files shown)


tests/federation_room_get_missing_events_test.go
```diff
@@ -34,6 +34,8 @@ func TestOutboundFederationIgnoresMissingEventWithBadJSONForRoomVersion6(t *test
 	srv := federation.NewServer(t, deployment,
 		federation.HandleKeyRequests(),
 		federation.HandleMakeSendJoinRequests(),
+		// Handle any transactions that the homeserver may send when connecting to another homeserver (such as presence)
+		federation.HandleTransactionRequests(nil, nil),
```
Kegsay, line 38: Alternatively you can use `srv.UnexpectedRequestsAreErrors = false`

anoadragon453: True, though we may still want to fail on other types of erroneous requests? Not sure how stringent we want to be I suppose.

tests/federation_room_send_test.go
```diff
@@ -22,54 +21,47 @@ func TestOutboundFederationSend(t *testing.T) {
...
-		defer waiter.Finish()
-		var body gomatrixserverlib.Transaction
-		if err := json.NewDecoder(req.Body).Decode(&body); err != nil {
-			t.Errorf("failed to decode /send body: %s", err)
-			w.WriteHeader(500)
-			return
-		}
-		if len(body.PDUs) != 1 {
-			t.Fatalf("got %d pdus, want %d", len(body.PDUs), 1)
-		}
```
anoadragon453, line 56: We're losing this check in this change, though it's not as simple to perform this check as we only get the event instead of the transaction.

I'm wondering whether we should also pass the entire transaction to the callback in case it wants to check other parts of it?

tests/federation_room_get_missing_events_test.go
Kegsay: Yeah good point.

tests/federation_room_send_test.go
Kegsay: I'd leave it out, if people care about # events in a txn then they shouldn't use this helper.

Kegsay (approved):

Issue #49, opened 2020-12-07
TestMediaWithoutFileName test does not allow changes in content type between media upload and download


anoadragon453:
The tests:

* `Can download without a file name locally`
* `Can download without a file name over federation`

that are a part of `TestMediaWithoutFileName` both have a check for the `Content-Type` header value that was provided during initial upload of the test media. They ensure this value does not change between uploading and downloading media and its associated metadata.

Synapse will however [augment the content type](https://github.com/matrix-org/synapse/pull/7044/files) of text files in order to specify UTF-8 as their encoding. The reasoning behind this is at https://github.com/matrix-org/synapse/issues/7043.

While the spec doesn't [explicitly state](https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-media-r0-download-servername-mediaid) that the `Content-Type` response header when downloading a file needs to *exactly* match what was provided during upload, it does state that it should be the content type of the file, and so homeservers should at least try to get it right.

The original sytest this was taken from, `Can download without a file name locally`, did not check `Content-Header`. I personally think it is valuable to do so, but we may want to make an exception for this modification, as it does seem to be beneficial and not against the spec, however vague it might be.

richvdh:
>  both have a check for the request's Content-Type header value.

which request? The one Complement sends to the server to fetch the media? That seems to make no sense?

anoadragon453:
> > both have a check for the request's Content-Type header value.
> 
> which request? The one Complement sends to the server to fetch the media? That seems to make no sense?

No sorry, the `Content-Type` header of the initial upload media request.

Kegsay:
Maybe just check the mime-type on the Content-Type header and not a literal string match?

anoadragon453:
Right, looks like we can use [`http.DetectContentType`](https://golang.org/pkg/net/http/#DetectContentType) for this.

anoadragon453:
`http.DetectContentType` returns `text/plain; charset=utf-8` lol.

I'll just do a split on the first string before `;`.

turt2live:
fwiw this is a conversation that should happen in the spec at some point - I have fairly strong opinions that the server should not be doing certain things to the header (like interpreting the body to pick a different header) or changing charsets on people, however the spec does indeed do a poor job of explaining what the requirements are. In general, the server is expected to be largely a passthrough for the content-type header, but whether that makes sense or not is a spec concern.

[...]

PR #51, opened 2020-12-09
Check the mimetype, not the whole Content-Type header on media download

anoadragon453:
Fixes https://github.com/matrix-org/complement/issues/49

By checking just the mimetype instead of an exact string match, homeservers are allowed to add charsets and other directives if they choose. As long as the mimetype stays the same, we're happy.

A comment on the original issue states that the spec is a bit vague about this, and that ideally it would be more explicit about what should happen here. At the moment though this is all we're going to test as it matches the expected behaviour from homeservers.

cc @turt2live

### DIFF
--- a/tests/media_nofilename_test.go
+++ b/tests/media_nofilename_test.go
@@ -57,7 +57,11 @@ func TestMediaWithoutFileName(t *testing.T) {
 			b, ct := alice.DownloadContent(t, mxc)
 
 			// Check the Content-Type response header.
-			// Homeservers are free to add other directives. All we want to check is the mimetype.
+			// NOTSPEC: There is ambiguity over whether the homeserver is allowed to change the
+			// Content-Type header here from what is sent by the client during upload. Synapse does
+			// so, and there are benefits to doing so, but the spec needs to pick a side.
+			// For now, we're operating under the assumption that homeservers are free to add other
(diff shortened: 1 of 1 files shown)


tests/media_nofilename_test.go
```diff
@@ -52,8 +53,19 @@ func TestMediaWithoutFileName(t *testing.T) {
 			mxc := alice.UploadContent(t, file, fileName, contentType)
 			must.NotEqualStr(t, mxc, "", "did not return an MXC URI")
 			must.StartWithStr(t, mxc, "mxc://", "returned invalid MXC URI")
+
 			b, ct := alice.DownloadContent(t, mxc)
-			must.EqualStr(t, ct, contentType, "wrong Content-Type returned")
+
+			// Check the Content-Type response header.
+			// Homeservers are free to add other directives. All we want to check is the mimetype.
+			mimeType := strings.Split(ct, ";")[0]
```
Kegsay, line 61: Please add `NOTSPEC:` and explain the ambiguities that @turt2live pointed out.

Kegsay (approved):

tests/media_nofilename_test.go
anoadragon453: Oh ok. Should we add this wording to `ONBOARDING.md` or similar to standardise it?

PR #52, opened 2020-12-09
Add linting via golangci-lint

anoadragon453:
This PR adds a `.golangci.yml` file which the [golangci-lint](https://github.com/golangci/golangci-lint) tool can make use of to run linting over the codebase. This tool was chosen as it is can config and run a multitude of linters and is also the one that is used in [Dendrite](https://github.com/matrix-org/dendrite)'s codebase.

For the enabled/disabled list of linters, I started with those that Dendrite makes use of + new linters that have been introduced into golangci-lint since Dendrite's last `.golangci.yml` file update and disabled by unnecessary functionality for a test suite + recommendations by @Kegsay.

After that there's a bunch of commits fixing up all the leftover lint errors that were found.

Note that 1b1a5c5 is just moving config blocks around and doesn't perform any actual changes.

And finally one fixup commit, 2fe9efc, actually disables a check as a quick workaround and I'd prefer some input on it. If we decide we need a proper fix and it's big then we may want to take that to a separate PR.

Recommended to reviewed commit-by-commit.


internal/federation/server.go
```diff
@@ -98,7 +99,7 @@ func NewServer(t *testing.T, deployment *docker.Deployment, opts ...func(*Server
 	})
 
 	// generate certs and an http.Server
-	httpServer, certPath, keyPath, err := federationServer("name", srv.mux)
```
Kegsay, line 101: Please `if err != nil { t.Fatalf }` this.

internal/federation/server.go
```diff
@@ -200,7 +201,9 @@ func (s *Server) Listen() (cancel func()) {
 		defer wg.Done()
 		err := s.srv.ListenAndServeTLS(s.certPath, s.keyPath)
 		if err != nil && err != http.ErrServerClosed {
-			s.t.Fatalf("ListenFederationServer: ListenAndServeTLS failed: %s", err)
+			s.t.Logf("ListenFederationServer: ListenAndServeTLS failed: %s", err)
+			// Note that running s.t.FailNow is not allowed in a separate goroutine
+			// TODO: Figure out a way to fail the test here
```
Kegsay, line 206: We probably shouldn't fail the test in retrospect, as this error is from a cleanup process. If the server never listened in the first place (e.g conflicting port numbers) then this will result in `connection refused` errors in tests.

tests/federation_keys_test.go
```diff
@@ -71,7 +73,7 @@ func TestInboundFederationKeys(t *testing.T) {
 	if !sigBase64.Exists() {
 		t.Fatalf("missing signature for hs1.%s", keyID)
 	}
-	sigBytes, err := base64.RawStdEncoding.DecodeString(sigBase64.Str)
```
Kegsay, line 74: Please `if err != nil { t.Fatalf }` this error.

Kegsay (approved):

internal/federation/server.go
anoadragon453: Good point! I'll remove the TODO.

PR #53, opened 2020-12-16
Add find-lint.sh script

anoadragon453:
This PR adds a find-lint.sh script in a new `build/scripts/` directory, similar to [Dendrite](https://github.com/matrix-org/dendrite/tree/master/build/scripts).

I also realised during this PR that the `DENDRITE_LINT_CONCURRENCY` env var mentioned in Dendrite's find-lint.sh script doesn't actually get used anywhere :) 

Have fixed that in this one, and will PR a fix to Dendrite's after this one is approved.

I also had a couple other things to fix:

- We were accidentally setting the default lint concurrency to 4 CPUs, when really it should be number of available system CPUs.
- Some unlinted code got into master since the lint PR (#52) last had master merged into it. So I've fixed those up here as well.

The intention for this script is to be called by both developers and by CI. Reviewable commit-by-commit.

### DIFF
--- a/build/scripts/find-lint.sh
+++ b/build/scripts/find-lint.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+# Runs the linters against complement
+
+# The linters can take a lot of resources and are slow, so they can be
+# configured using the following environment variables:
+#
+# - `COMPLEMENT_LINT_CONCURRENCY` - number of concurrent linters to run,
+#   golangci-lint defaults this to NumCPU
+# - `GOGC` - how often to perform garbage collection during golangci-lint runs.
+#   Essentially a ratio of memory/speed. See https://golangci-lint.run/usage/performance/#memory-usage
+#   for more info.
+
+
+cd `dirname $0`/..
+
+args=""
+if [ ${1:-""} = "fast" ]
+then args="--fast"
+fi
+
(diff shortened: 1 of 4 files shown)


build/scripts/find-lint.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
Kegsay, line 1: Please always set `-eu` in your bash scripts.

`#!/bin/bash -eu`. This is because bash is a silly sausage. `-e` means if any pipeline ever ends with a non-zero exit code, terminate the script immediately. `-u` means treat unset local variables as errors (so calling `$FOO` without defining `FOO` first produces an error. These flags at least make bash a bit more sane like a normal programming language.

Kegsay: If you want to be particularly picky, you should also set `-o pipefail` to make any command in a pipeline return an error status. By default, only the last command will do this.

anoadragon453: Ah, I actually didn't set `-e` on purpose in case `golangci-lint run` failed and we exited the script without running the `mv` for go.mod and go.sum. However I guess the `|| exit_code=1` already ensures that that won't happen. For this reason I won't add `-o pipefail` though (although TIL!).

I can add `-e` though. I also omitted `-x` as it seemed a bit annoying to print out everything when running it as a dev.

Kegsay (approved):

PR #59, opened 2020-12-29
Handle transactions sent during UnverifiableEvents test

anoadragon453:
Once again Synapse likes to send presence when it joins a room. I wonder if we should perhaps make `federation.HandleTransactionRequests(nil, nil)` the default for new servers, with the ability to override it if desired?

The only thing I worry about there is that you may want a test that 404's on transactions.  Maybe at that point we could just extend HandleTransactionRequests to return custom responses.


Kegsay:
The entire point of `Handle...` functions is to be explicit about what the server is automatically handling so you aren't surprised by the behaviour of Complements federation server. I don't agree that automatically handling with optional override is desirable here.

Kegsay (approved):

PR #61, opened 2021-01-18
MSC2946

Kegsay:
CS API version.


neilalexander (approved):
Looks reasonable to me

PR #62, opened 2021-01-19
Add a new dockerfile for worker-mode synapse

anoadragon453:
This PR adds a new Dockerfile containing the configuration for a Synapse running in worker mode. Some additional files were required to support the image.

1. A caddy binary needed to be placed in the container for termination TLS requests on the federation port (thanks to Conduit for the idea!). The included file is a Caddy config json file.
2. A shared config file to hand to each Synapse worker was necessary. This file is subtley different from [dockerfiles/synapse/homeserver.yaml](https://github.com/matrix-org/complement/blob/master/dockerfiles/synapse/homeserver.yaml), though it may be nice to have both files be derived from a shared file in the future.

The intention of this is to add testing of Synapse in worker mode to Complement's CI. It is paired with https://github.com/matrix-org/synapse/pull/9162.

### DIFF
--- a/dockerfiles/SynapseWorkers.Dockerfile
+++ b/dockerfiles/SynapseWorkers.Dockerfile
@@ -46,12 +46,8 @@ ENTRYPOINT \
   SYNAPSE_REPORT_STATS=no \
   # Set postgres authentication details which will be placed in the homeserver config file
   POSTGRES_PASSWORD=somesecret POSTGRES_USER=postgres POSTGRES_HOST=localhost \
-  # Note: This list currently includes all worker types other than federation_sender, as
-  # Synapse fails to send federation transactions with it enabled.
-  # https://github.com/matrix-org/synapse/issues/9192
-  SYNAPSE_WORKERS=pusher,user_dir,media_repository,appservice,synchrotron,federation_reader,federation_inbound \
-  # To use all available workers:
-  #SYNAPSE_WORKERS=* \
+  # Use all available worker types
+  SYNAPSE_WORKERS=* \
   # Run the script that writes the necessary config files and starts supervisord, which in turn
   # starts everything else
   /configure_workers_and_start.py
--- a/dockerfiles/SynapseWorkers.Dockerfile
+++ b/dockerfiles/SynapseWorkers.Dockerfile
@@ -5,6 +5,12 @@ FROM matrixdotorg/synapse:workers
 # Tell Complement that we are using its custom CA
 ENV COMPLEMENT_CA=true
 
+# Download a caddy server to stand in front of nginx and terminate TLS using Complement's
+# custom CA.
+# We include this near the top of the file in order to cache the result.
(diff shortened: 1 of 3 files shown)


dockerfiles/SynapseWorkers.Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+  LC_COLLATE='C' \
+  LC_CTYPE='C' \
+  template=template0;\" | psql" && pg_ctlcluster 11 main stop
+
+# Modify the shared homeserver config with postgres support, certificate setup
+# and the disabling of rate-limiting
+COPY synapse/workers-shared.yaml /conf/workers/shared.yaml
+
+# Set up TLS certificates using the custom CA
+COPY keys/* /ca/
```
Kegsay, line 26: Not following this here. Complement will auto-mount a volume to `/ca` which will contain the certs. Why are we doing this copy from a `/keys` directory?

dockerfiles/SynapseWorkers.Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+# Generate a signing key
+RUN generate_signing_key.py -o /conf/server.signing.key
+
+WORKDIR /root
+
+# Download a caddy server to stand in front of nginx and terminate TLS using Complement's
+# custom CA
+RUN curl -OL "https://github.com/caddyserver/caddy/releases/download/v2.3.0/caddy_2.3.0_linux_amd64.tar.gz" && \
+  tar xzf caddy_2.3.0_linux_amd64.tar.gz && rm caddy_2.3.0_linux_amd64.tar.gz
```
Kegsay, line 39: I'd move this up near the top of the file to make better use of docker caching.

Kegsay (approved):

dockerfiles/SynapseWorkers.Dockerfile
anoadragon453: Good point. This was just copied over from the existing Synapse config, and can go. However I did notice that having `ENV COMPLEMENT_CA=true` in the dockerfile (at least in this setup) doesn't seem to goad Complement into running in CA mode.

Including the variable when running the dockerfile works as expected.

Kegsay: `COMPLEMENT_CA=true` is a config option for running Complement itself, it's driven by the end-user not by random dockerfiles. It's up to the dockerfile to look in `/ca` if it exists to know whether to use it or not.

anoadragon453: I see, I was confused by the "or the docker container" bit in the README then: https://github.com/matrix-org/complement#complement-pki.

Kegsay:
Is this ready to merge?

anoadragon453:
Yeah, sorry. I was just waiting on the CI stuff to get sorted before merging. But that might be a while and I want to put some more PRs up for fixing cert stuff in Complement that would require more changes here.

So let's just merge it now and do anything more in other PRs.

Issue #65, opened 2021-01-25
Consider removing COMPLEMENT_CA and having it always on


anoadragon453:
Before too many more homeserver dockerfiles are written, what do people think about removing the `COMPLEMENT_CA` environment variable, and making its behaviour the default? `/ca` will just be mounted in all docker containers, and homeserverd dockerfiles can choose to use the certificate and key in `/ca` or not.

Existing dockerfiles, such as the [Synapse Monolith](https://github.com/matrix-org/complement/blob/df150033eea5815c41111f700c5c00b9eb624f25/dockerfiles/Synapse.Dockerfile#L19) one will need to be updated, but I've been planning to adjust that one to use Complement's CA anyhow.

Kegsay:
It adds complexity to have it on which is why I wanted it to be opt-in originally. For now, my position remains the same.

richvdh:
> Existing dockerfiles, such as the Synapse Monolith one will need to be updated,

does it really though? Looks like it will just work; it'll just use the complement CA instead of its own.

in any case, +1. I feel like there should be one way to do this.

richvdh:
> It adds complexity to have it on

how so?

(@anoadragon453: what is the benefit of this approach over using a static CA cert as the monolith image does?)

Kegsay:
> how so?

Complement needs to generate the cert and add a volume to every docker container it runs. There's some magic to determine where to look/query depending on if Complement itself is running in docker https://github.com/matrix-org/complement/blob/master/internal/docker/builder.go#L350

The alternative is no volumes, and skipping TLS checks.

richvdh:
well, we've talked previously about how the magic `CI` flag is... magic. 

I'm still failing to grasp why setting up a CA is an *optional* feature. Either it's required to run complement's tests, so should be enabled by default; or it's not required, in which case we can just get rid of it?

anoadragon453:
> (@anoadragon453: what is the benefit of this approach over using a static CA cert as the monolith image does?)

Well, I was hoping it would allow us to set `federation_verify_certificates: true` in Synapse's config, although that didn't end up being the case: https://github.com/matrix-org/complement/blob/40d2a8d2a96f89e922a436336f068573d6472ddb/dockerfiles/synapse/workers-shared.yaml#L8-L11

anoadragon453:
Well the above is likely the case because I didn't follow [the instructions](https://github.com/matrix-org/complement#complement-pki) on using the certificate correctly.

[...]

PR #69, opened 2021-01-29
Fixes to Complement's Federation Certificate

anoadragon453:
This PR contains two fixes, separated by commit.

1. Add Subject information to Complement's CA and ephemeral certificates. Federation clients will not accept a certificate if they are unable to get the details of the CA that issued them. It looks for this information in the longterm CA cert.
2. The longterm cert would expire after 5 hours. It is not regenerated if it already exists. Federation clients will not accept expired certificates. I've updated the expiration to 10 years.

Because the longterm CA certificate is not regenerated if it already exists, people will need to clear the old certificates (located in `/tests/ca/`) to generate a new certificate.

Note: I think for 1. information is only required for the CA cert, but I've added the same to the ephemeral cert just in case.

### DIFF
--- a/internal/federation/server.go
+++ b/internal/federation/server.go
@@ -281,7 +281,8 @@ func GetOrCreateCaCert() (*x509.Certificate, *rsa.PrivateKey, error) {
 		}
 	}
 
-	certificateDuration := time.Hour * 5
+	// valid for 10 years
+	certificateDuration := time.Hour * 24 * 365 * 10
 	priv, err := rsa.GenerateKey(rand.Reader, 4096)
 	if err != nil {
 		return nil, nil, err
--- a/internal/federation/server.go
+++ b/internal/federation/server.go
@@ -6,6 +6,7 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
+	"crypto/x509/pkix"
 	"encoding/json"
 	"encoding/pem"
 	"errors"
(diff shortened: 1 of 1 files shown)


valkum:
Copy from the Complement Room:

> I think the better solution would be to check for the expiration date in the GetOrCreate function, and don't return the existing cert if it expires in less than 2 hours.
Long valid certs are considered bad in the security space form my understanding. While most problems with long running certs should not be encountered in this scenario (running mostly in CI). If you were to run Complement from your host and you want to enforce TLS checks, you would need to add the now 10year valid CA cert to your local trust store, which I find pretty rough. 
Furthermore, still while the chances are pretty low, sometime in the future no one has to debug a failing build because the CA cert expired, if we just check validity and renew if needed.
You can renew a cert to keep the same public key, and in my believe this would be the proper solution here.

> Oh and we need to make sure that we add the CA to the trust store in the complement container. That might be missing as well.
Reasoning for this: I rather like a secure by default design where Complement enforces TLS, to check whether a used container has a working TLS implementation and with that an option to disable this checks, than a insecure by default design. In my opinion it would be good to see if Homeservers correctly handle TLS validation. 
For this, there would be additional work needed to spin up a federation server in complement with a invalid cert to see if the Homeserver still communicates with Complement.

anoadragon453:
> If you were to run Complement from your host and you want to enforce TLS checks, you would need to add the now 10year valid CA cert to your local trust store, which I find pretty rough.

Perhaps a better solution may be to create a HTTP client that is used for federation requests to homeservers which contains the CA cert (see [tls.Config](https://golang.org/pkg/crypto/tls/#Config).RootCAs? That way we wouldn't be adding the cert to the host (and wouldn't be littering it with loads of expired certs).

> Reasoning for this: I rather like a secure by default design where Complement enforces TLS, to check whether a used container has a working TLS implementation and with that an option to disable this checks, than a insecure by default design. In my opinion it would be good to see if Homeservers correctly handle TLS validation.

Yeah, I agree. FYI see the discussion in https://github.com/matrix-org/complement/issues/65 regarding whether to remove the `COMPLEMENT_CA` environment variable. This could take the form of a separate env var, or piggyback off the existing one.

> For this, there would be additional work needed to spin up a federation server in complement with a invalid cert to see if the Homeserver still communicates with Complement.

This sounds worthwhile, though let's make a separate issue for that.

valkum:
> Perhaps a better solution may be to create a HTTP client that is used for federation requests to homeservers which contains the CA cert (see [tls.Config](https://golang.org/pkg/crypto/tls/#Config).RootCAs? That way we wouldn't be adding the cert to the host (and wouldn't be littering it with loads of expired certs).

That sounds reasonable and also helps with implementing an invalid federation endpoint for testing TLS verification.

Kegsay (approved):

Kegsay:
@anoadragon453 can you merge this please?

anoadragon453:
@Kegsay Apologies for the delay. I've been working on and off on some improvements to this PR as discussed above with @valkum. Mainly having Complement verify federation certificates and seeding Complement's federation client with the CA cert instead of needing to add it to the host.

Those changes can certainly come in a separate PR though.

PR #72, opened 2021-02-09
Test knock rooms appear in the public room directory

anoadragon453:
This tests the ["Changes regarding the Public Rooms Directory"](https://github.com/Sorunome/matrix-doc/blob/soru/knock/proposals/2403-knock.md#changes-regarding-the-public-rooms-directory) section of knocking MSC2403.

Essentially boiling down to:

* Knock rooms can be added to and appear in the public room directory.
* Clients see a `join_rule` entry in the room result to know whether to knock or not to knock. It really is the question.

### DIFF
--- a/tests/msc2403_test.go
+++ b/tests/msc2403_test.go
@@ -15,11 +15,13 @@ package tests
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/url"
 	"testing"
 
 	"github.com/matrix-org/complement/internal/b"
 	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/internal/match"
 	"github.com/matrix-org/complement/internal/must"
 	"github.com/tidwall/gjson"
 )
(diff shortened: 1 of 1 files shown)


tests/msc2403_test.go
```diff
@@ -342,3 +344,101 @@ func doInitialSync(t *testing.T, c *client.CSAPI) string {
 	since := client.GetJSONFieldStr(t, body, "next_batch")
 	return since
 }
+
+// knockRoomsInPublicRoomsDirectoryTest will create a knock room, attempt to publish it to the public rooms directory,
+// and then check that the room appears in the directory. The room's entry should also have a 'join_rule' field
+// representing a knock room. For sanity-checking, this test will also create a public room and ensure it has a
+// 'join_rule' representing a publicly-joinable room.
+func knockRoomsInPublicRoomsDirectoryTest(t *testing.T) {
```
Kegsay, line 352: Why isn't this a test?

Kegsay (approved):
I don't see how this test runs, but the test itself looks good to me.

Kegsay (approved):

PR #74, opened 2021-02-22
Add application service support to blueprints

MadLittleMods:
Add application service support to blueprints

 - Only supports defining a basic app service
 - In tests, you can use the app service `as_token` to interact with the homserver API
 - Does not support defining the `namespaces` part of the app service registration config
 - Does not support a mocked/stubbed actual app service that the homeserver calls back to at the `URL`

Split out from https://github.com/matrix-org/complement/pull/68

---

 - https://matrix.org/docs/guides/application-services
 - https://matrix.org/docs/spec/application_service/r0.1.2

### DIFF
--- a/internal/b/blueprints.go
+++ b/internal/b/blueprints.go
@@ -20,8 +20,6 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
-
-	"github.com/sirupsen/logrus"
 )
 
 // KnownBlueprints lists static blueprints
@@ -132,10 +130,6 @@ func Validate(bp Blueprint) (Blueprint, error) {
 			}
 		}
 	}
-
-	logrus.WithFields(logrus.Fields{
-		"bp": bp.Homeservers[0].ApplicationServices,
-	}).Error("after modfiying bp")
 
 	return bp, nil
 }
--- a/internal/b/blueprints.go
+++ b/internal/b/blueprints.go
@@ -83,12 +83,10 @@ type ApplicationService struct {
 }
 
 type Event struct {
-	Type           string
-	Sender         string
-	OriginServerTS uint64
-	StateKey       *string
-	PrevEvents     []string
-	Content        map[string]interface{}
+	Type     string
+	Sender   string
+	StateKey *string
+	Content  map[string]interface{}
 	// This field is ignored in blueprints as clients are unable to set it. Used with federation.Server
 	Unsigned map[string]interface{}
 }
--- a/internal/b/blueprints.go
+++ b/internal/b/blueprints.go
@@ -15,9 +15,13 @@
 package b
 
 import (
+	"crypto/rand"
+	"encoding/hex"
 	"fmt"
 	"strconv"
 	"strings"
+
+	"github.com/sirupsen/logrus"
 )
 
 // KnownBlueprints lists static blueprints
@@ -26,6 +30,7 @@ var KnownBlueprints = map[string]*Blueprint{
 	BlueprintAlice.Name:                       &BlueprintAlice,
 	BlueprintFederationOneToOneRoom.Name:      &BlueprintFederationOneToOneRoom,
 	BlueprintFederationTwoLocalOneRemote.Name: &BlueprintFederationTwoLocalOneRemote,
+	BlueprintHSWithApplicationService.Name:    &BlueprintHSWithApplicationService,
 	BlueprintOneToOneRoom.Name:                &BlueprintOneToOneRoom,
 	BlueprintPerfManyMessages.Name:            &BlueprintPerfManyMessages,
 	BlueprintPerfManyRooms.Name:               &BlueprintPerfManyRooms,
@@ -46,6 +51,8 @@ type Homeserver struct {
 	Users []User
 	// The list of rooms to create on this homeserver
 	Rooms []Room
+	// The list of application services to create on the homeserver
+	ApplicationServices []ApplicationService
 }
 
 type User struct {
@@ -68,11 +75,22 @@ type Room struct {
 	Events     []Event
 }
 
+type ApplicationService struct {
+	ID              string
+	HSToken         string
+	ASToken         string
+	URL             string
+	SenderLocalpart string
+	RateLimited     bool
+}
+
 type Event struct {
-	Type     string
-	Sender   string
-	StateKey *string
-	Content  map[string]interface{}
+	Type           string
+	Sender         string
+	OriginServerTS uint64
+	StateKey       *string
+	PrevEvents     []string
+	Content        map[string]interface{}
 	// This field is ignored in blueprints as clients are unable to set it. Used with federation.Server
 	Unsigned map[string]interface{}
 }
@@ -107,7 +125,18 @@ func Validate(bp Blueprint) (Blueprint, error) {
 				return bp, err
 			}
 		}
+		for i, as := range hs.ApplicationServices {
+			hs.ApplicationServices[i], err = normalizeApplicationService(as)
+			if err != nil {
+				return bp, err
+			}
+		}
 	}
+
+	logrus.WithFields(logrus.Fields{
+		"bp": bp.Homeservers[0].ApplicationServices,
+	}).Error("after modfiying bp")
+
 	return bp, nil
 }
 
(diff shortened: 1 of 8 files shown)


internal/docker/builder.go
```diff
@@ -371,51 +382,75 @@ func getCaVolume(docker *client.Client, ctx context.Context) (map[string]struct{
...
+		caMount = mount.Mount{
+			Type:   mount.TypeBind,
+			Source: path.Join(cwd, "ca"),
+			Target: "/ca",
 		}
 	}
 	return caVolume, caMount, nil
 }
 
+func getAppServiceVolume(docker *client.Client, ctx context.Context) (string, mount.Mount, error) {
```
MadLittleMods, line 418:
```
// getAppServiceVolume returns the correct mounts and volumes for providing the `/appservice` directory to homeserver containers
// containing application service registration files to be used by the homeserver 
```

internal/docker/builder.go
```diff
@@ -457,7 +509,32 @@ func deployImage(
...
+		err = tw.WriteHeader(&tar.Header{
+			Name: fmt.Sprintf("/appservices/%s.yaml", asID),
+			Mode: 0777,
+			Size: int64(len(registration)),
+		})
+		if err != nil {
+			return nil, fmt.Errorf("Failed to copy regstration to container: %v", err)
+		}
+		tw.Write([]byte(registration))
+		tw.Close()
```
MadLittleMods, line 528: This is from https://stackoverflow.com/a/52131297/796832

internal/b/hs_with_application_service.go
```diff
@@ -0,0 +1,25 @@
...
+			Users: []User{
+				{
+					Localpart:   "@alice",
+					DisplayName: "Alice",
+				},
+			},
+			ApplicationServices: []ApplicationService{
+				{
+					ID:              "my_as_id",
+					URL:             "http://localhost:9000",
```
Kegsay, line 18: What is the purpose of this URL? Does this need to be changed when run under CI?

internal/b/blueprints.go
```diff
@@ -68,11 +75,22 @@ type Room struct {
...
-	Type     string
-	Sender   string
-	StateKey *string
-	Content  map[string]interface{}
+	Type           string
+	Sender         string
+	OriginServerTS uint64
+	StateKey       *string
+	PrevEvents     []string
+	Content        map[string]interface{}
```
Kegsay, line 93: Please do not add these fields.

internal/b/blueprints.go
```diff
@@ -107,7 +125,18 @@ func Validate(bp Blueprint) (Blueprint, error) {
...
+			hs.ApplicationServices[i], err = normalizeApplicationService(as)
+			if err != nil {
+				return bp, err
+			}
+		}
 	}
+
+	logrus.WithFields(logrus.Fields{
+		"bp": bp.Homeservers[0].ApplicationServices,
+	}).Error("after modfiying bp")
```
Kegsay, line 138: Debug logging: please remove

internal/docker/builder.go
```diff
@@ -328,17 +337,19 @@ func (d *Builder) constructHomeserver(blueprintName string, runner *instruction.
...
+		d.Docker, d.BaseImage, d.CSAPIPort, fmt.Sprintf("complement_%s", contextStr), blueprintName, hs.Name, asIDToRegistrationMap, contextStr,
 		networkID, d.config.VersionCheckIterations,
 	)
 }
 
 // getCaVolume returns the correct mounts and volumes for providing a CA to homeserver containers.
-func getCaVolume(docker *client.Client, ctx context.Context) (map[string]struct{}, []mount.Mount, error) {
-	var caVolume map[string]struct{}
-	var caMount []mount.Mount
+func getCaVolume(docker *client.Client, ctx context.Context) (string, mount.Mount, error) {
```
Kegsay, line 350: Please document the return types.

internal/docker/builder.go
```diff
@@ -371,51 +382,77 @@ func getCaVolume(docker *client.Client, ctx context.Context) (map[string]struct{
...
+			Source: path.Join(cwd, "ca"),
+			Target: "/ca",
 		}
 	}
 	return caVolume, caMount, nil
 }
 
+// getAppServiceVolume returns the correct mounts and volumes for providing the `/appservice` directory to homeserver containers
+// containing application service registration files to be used by the homeserver
+func getAppServiceVolume(docker *client.Client, ctx context.Context) (string, mount.Mount, error) {
```
Kegsay, line 420: Please document the return types.

internal/docker/builder.go
```diff
@@ -371,51 +382,77 @@ func getCaVolume(docker *client.Client, ctx context.Context) (map[string]struct{
...
+}
+
+func generateASRegistrationYaml(as b.ApplicationService) string {
+	return fmt.Sprintf("id: %s\n", as.ID) +
+		fmt.Sprintf("hs_token: %s\n", as.HSToken) +
+		fmt.Sprintf("as_token: %s\n", as.ASToken) +
+		fmt.Sprintf("url: '%s'\n", as.URL) +
+		fmt.Sprintf("sender_localpart: %s\n", as.SenderLocalpart) +
+		fmt.Sprintf("rate_limited: %v\n", as.RateLimited) +
+		"namespaces:\n" +
```
Kegsay, line 443: You can use literal syntax via `

internal/docker/builder.go
Kegsay: Please chuck that link as a comment with a short prose to explain what this code is doing. It's unclear why we're doing this otherwise, but I do like that we can do this!

internal/docker/builder.go
```diff
@@ -457,7 +511,32 @@ func deployImage(
...
 	}
+
 	containerID := body.ID
+
+	// Create the application service files
+	for asID, registration := range asIDToRegistrationMap {
+		var buf bytes.Buffer
+		tw := tar.NewWriter(&buf)
+		err = tw.WriteHeader(&tar.Header{
+			Name: fmt.Sprintf("/appservices/%s.yaml", asID),
```
Kegsay, line 522: Please can we do filepath escapes here, as I would like to in the future allow untrusted Blueprint input.

Kegsay (approved):
Love it, just a few minor issues to resolve!

internal/b/hs_with_application_service.go
MadLittleMods: It's the URL that the homeserver will callback to when it has new data to pass through it: Homeserver gets a new message -> Sends it to the application service at that address.

We're not working with any of that side of application services for my tests in https://github.com/matrix-org/complement/pull/68

This URL would probably be best implemented by a stubbed HTTP server in a Complement test which can be figured out when someone needs it.

internal/b/blueprints.go
MadLittleMods: TODO: These are used in https://github.com/matrix-org/complement/pull/68

internal/docker/builder.go
```diff
@@ -426,26 +457,39 @@ func deployImage(
...
 		Image: imageID,
-		Env:   []string{"SERVER_NAME=" + hsName, "COMPLEMENT_CA=" + os.Getenv("COMPLEMENT_CA")},
+		Env:   env,
 		//Cmd:   d.ImageArgs,
 		Labels: map[string]string{
 			complementLabel:        contextStr,
 			"complement_blueprint": blueprintName,
 			"complement_hs_name":   hsName,
 		},
-		Volumes: caVolume,
```
MadLittleMods, line 444: Removing all of the `Volumes` config as it doesn't seem necessary and works without it. We use the `container.HostConfig,Mounts` already which covers this use case.

 - Seems like the `Volumes` syntax is used to create an anonymous volume, https://stackoverflow.com/a/58916037/796832
 - And lots of people not knowing what `Volumes` syntax is or what to do. Seems like Mounts is the thing to use
    - https://github.com/fsouza/go-dockerclient/issues/155
    - https://stackoverflow.com/questions/55718603/golang-docker-library-mounting-host-directory-volumes
    - https://stackoverflow.com/questions/48470194/defining-a-mount-point-for-volumes-in-golang-docker-sdk

internal/docker/builder.go
MadLittleMods: Refactored the previous `map[string]struct{}` and string `Volumes` stuff out of the code. See https://github.com/matrix-org/complement/pull/74#discussion_r580642237

internal/docker/builder.go
MadLittleMods: I assume you mean using `client.GjsonEscape`?

internal/docker/builder.go
MadLittleMods: String literal docs: https://golang.org/ref/spec#String_literals

---

Initially, I used a multiline backtick raw string but the `fmt.Sprintf` on it was pretty weird since I needed to make sure the order was correct over a list of 6 variables at the end. Seems pretty easy to get it wrong whenever touching it.

Also, since whitespaces matters, it looks out of place to have everything smashed against the beginning of the line and doesn't match the surrounding indentation of the code.

Is there a cleaner way to substitute variables (string interpolation)?

Kegsay: Hmm, good points. Nope, I think the combo of needing to inject vars and whitespace are good enough reasons not to use string literals.

internal/docker/builder.go
Kegsay: I mean path escaping so you can't have an `asID` like `../../../etc/passwd`

Kegsay:
LGTM

MadLittleMods:
Woot! Thanks for the review @Kegsay 🐊

PR #75, opened 2021-02-26
blueprints: Add a blueprint that creates E2EE capable users in a room

poljar:
This PR adds a blueprint that creates a bunch of E2EE capable users and joins them to the room, apologies for the dependency mess, I have no idea how to properly add deps in go.

Some thoughts for the future: It would be nice if we could post parameters to the blueprints that change the behavior of the blueprint slightly, e.g. I would like the number of users the E2EE blueprint creates to be dynamic, so one could start up a server with 100 users for some light and quick testing and one with 1000 for some more load.

 Creating 1000 users all the time would be annoying since it takes some time to create all the E2EE keys and membership events.


internal/b/perf_e2ee_room.go
```diff
@@ -0,0 +1,107 @@
...
+								"msgtype": "m.text",
+							},
+							Sender: "@bob",
+						},
+						{
+							Type: "m.room.encryption",
+							// Complement crashes because it doesn't
+							// know how to URL encode an empty string, Element seems
+							// to allow a state key with a space in it so use it instead
+							StateKey: Ptr(" "),
```
Kegsay, line 55: This is invalid, we should fix Complement to not crash.

internal/b/blueprints.go
```diff
@@ -53,6 +54,8 @@ type User struct {
 	DisplayName string
 	AvatarURL   string
 	AccountData AccountData
+	DeviceId    *string
+	E2E         bool
```
Kegsay, line 58: Please document what this will do if `true`. I feel it would probably be more suitable to have `OneTimeKeys int`, and set up device/OTKs if `> 0`.

Kegsay (changes requested):
Looks good, just a few small things:
 - Swap `E2E bool` for `OneTimeKeys int` and document it.
 - Fix the crash on zero-length state keys.

Kegsay (approved):

Kegsay:
#79

PR #80, opened 2021-03-05
Add account snapshot binary

Kegsay:
This PR:
 - Adds `./cmd/account-snapshot` - see the README.
 - Fixes some lint naming issues
 - Tweaks the instruction runner to `/login` if the user already exists in the blueprint (e.g logging in on a different device).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+sync_snapshot.json
 complement
 .idea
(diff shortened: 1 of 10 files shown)


neilalexander (approved):
I've looked through it and overall looks sane, but admittedly I haven't actually ran it on my machine.

PR #82, opened 2021-03-05
Add dendrite on postgres

Kegsay:


neilalexander (approved):
We should get this into the Dendrite pipeline too!

PR #83, opened 2021-03-06
Update Synapse IP range blacklist option

anoadragon453:
This option has been deprecated: https://github.com/matrix-org/synapse/blob/master/UPGRADE.rst#blacklisting-ip-ranges

It is recommended to replace it with the similarly named `ip_range_blacklist`.


Kegsay (approved):

clokep (approved):
Seems reasonable to me.

PR #90, opened 2021-03-31
feat: add login tests

meenal06:
Signed-off-by: meenal <meenal@Meenal-MacBook-Pro.local>

### DIFF
--- a/tests/apidoc_login_test.go
+++ b/tests/apidoc_login_test.go
@@ -1,0 +1,48 @@
+package tests
+
+import (
+	"encoding/json"
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
+	"github.com/tidwall/gjson"
+	"io/ioutil"
+	"testing"
+)
+
+func TestLogin(t *testing.T) {
+	deployment := Deploy(t, "login", b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	t.Run("parallel", func(t *testing.T) {
+		// sytest: POST /login can log in as a user
+		t.Run("POST /login can login as user", func(t *testing.T) {
+			t.Parallel()
+			res := unauthedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "register"}, json.RawMessage(`{
+				"auth": {
+					"type": "m.login.dummy"
+				},
+				"username": "post-login-user",
+				"password": "superuser"
+			}`))
+			body, _ := ioutil.ReadAll(res.Body)
+			userId := gjson.Get(string(body), "user_id")
+			res = unauthedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "login"}, json.RawMessage(`
+			{
+				"type": "m.login.password"
+				"identifier": {
+					"type": "m.id.user",
+					"user": "post-login-user",
+				}
+				"password": "superuser"
+			}`))
+
+			must.MatchResponse(t, res, match.HTTPResponse{
+				JSON: []match.JSON{
+					match.JSONKeyTypeEqual("access_token", gjson.String),
+					match.JSONKeyEqual("server_name", "hs1"),
+				},
+			})
+		})
+	})
+}
--- a/tests/apidoc_login_test.go
+++ b/tests/apidoc_login_test.go
@@ -46,7 +46,7 @@ func TestLogin(t *testing.T) {
 			must.MatchResponse(t, res, match.HTTPResponse{
 				JSON: []match.JSON{
 					match.JSONKeyTypeEqual("access_token", gjson.String),
-					match.JSONKeyEqual("server_name", "hs1"),
+					match.JSONKeyEqual("home_server", "hs1"),
 				},
 			})
 		})
--- a/tests/apidoc_login_test.go
+++ b/tests/apidoc_login_test.go
@@ -2,12 +2,13 @@ package tests
 
 import (
 	"encoding/json"
+	"testing"
+
+	"github.com/tidwall/gjson"
+
 	"github.com/matrix-org/complement/internal/b"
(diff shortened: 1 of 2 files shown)


tests/apidoc_login_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/tidwall/gjson"
+	"testing"
+)
+
+func TestLogin(t *testing.T) {
+	deployment := Deploy(t, "login", b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	t.Run("parallel", func(t *testing.T) {
+		t.Run("POST /login can login as user", func(t *testing.T) {
```
Kegsay, line 17:
```suggestion
		// sytest: POST /login can log in as a user
		t.Run("POST /login can login as user", func(t *testing.T) {
```
As you are converting a sytest, please add the comment of the form `// sytest: the test name`. More information on this is [here](https://github.com/matrix-org/complement/blob/master/ONBOARDING.md#how-should-i-name-the-test-files--test-functions).

tests/apidoc_login_test.go
```diff
@@ -0,0 +1,41 @@
...
+				"auth": {
+					"type": "m.login.dummy"
+				},
+				"username": "post-login-user",
+				"password": "superuser"
+			}`))
+			must.MatchResponse(t, res, match.HTTPResponse{
+				JSON: []match.JSON{
+					match.JSONKeyTypeEqual("access_token", gjson.String),
+				},
```
Kegsay, line 36: It's really important that the Complement tests we make are as thorough as the Sytests we're replacing. The login test we're replacing also checks the `home_server` - https://github.com/matrix-org/sytest/blob/develop/tests/10apidoc/02login.pl#L96 so please do the same. The homeserver should `== "hs1"`.

Kegsay (changes requested):
A few minor things, but overall this is _excellent_. Thank you!

tests/apidoc_login_test.go
meenal06: Hi @Kegsay , I have made the change

meenal06:
Hi @Kegsay , please take a look now, I have made the requested changes

tests/apidoc_login_test.go
```diff
@@ -0,0 +1,42 @@
...
+				"username": "post-login-user",
+				"password": "superuser"
+			}`))
+			res = unauthedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "login"}, json.RawMessage(`{
+				"auth": {
+					"type": "m.login.dummy"
+				},
+				"username": "post-login-user",
+				"password": "superuser"
```
callahad, line 32: I don't think this structure is quite correct. The [matching SyTest](https://github.com/matrix-org/sytest/blob/develop/tests/10apidoc/02login.pl#L70-L101) sets:

```perl
type     => "m.login.password",
identifier => {
   type => "m.id.user",
   user => $user_id,
},
password => $password,
```

tests/apidoc_login_test.go
```diff
@@ -0,0 +1,42 @@
...
+	deployment := Deploy(t, "login", b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	t.Run("parallel", func(t *testing.T) {
+		// sytest: POST /login can log in as a user
+		t.Run("POST /login can login as user", func(t *testing.T) {
+			t.Parallel()
+			res := unauthedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "register"}, json.RawMessage(`{
+				"auth": {
+					"type": "m.login.password"
```
callahad, line 22: I believe this should be `m.login.dummy`

callahad (changes requested):
Thank you so much! I've added two comments inline :)

tests/apidoc_login_test.go
meenal06: Thanks for the review @callahad , looks like I made switched the auth types in registration and login, let me fix them asap.

tests/apidoc_login_test.go
meenal06: Hi @callahad I have corrected the order and have even captured the user_id from the register response. But how can I pass that userId variable in the json raw message. I would really appreciate any help here 😕

callahad: Great, thank you!

> how can I pass that userId variable in the json raw message

In this case, hardcoding it to "post-login-user" is probably fine. In general, the easiest way is probably to break up the string and use `+` to concatenate things ([playground](https://play.golang.org/p/9pwWDdn_0Hi)). For example:

```go
`{"user": "` + userid + `"}`
```

You could also use `fmt.Sprintf` ([playground](https://play.golang.org/p/ST5YpRg3GG0)) to do something like:

```go
fmt.Sprintf(`{"user": "%s"}`, userid)
```

But again, for now, just hardcode it 😄 

All that said, I think the JSON structure is still a little off. Instead of sending JSON like:

```json
{
  "auth": {
    "type": "m.login.password"
  },
  "identifier": {
    "type": "m.id.user",
    "user": "post-login-user"
  },
  "username": "post-login-user",
  "password": "superuser"
}
```

I think you only need to send:

```json
{
  "type": "m.login.password",
  "identifier": {
    "type": "m.id.user",
    "user": "post-login-user",
  },
  "password": "superuser"
}
```

(Meaning: I think you can drop the `auth` and `username` keys altogether.)

meenal06: I tried using `f.Sprintf` but since the complete Raw Message parameter is under `` literal, the string concatenation was not working as expected. 
But as you said, I will hardcode it for now, thanks 😄

meenal06:
Hi @callahad , I have done the requested changes and have also squashed off my commits, please take a look!

meenal06:
Hi @callahad , finally green checks 🤩 , I guess this is good to be merged.

callahad:
Wooohooo! 🎉  Thank you for persevering through the very fussy golangci-lint checks! This is great!

callahad (approved):

PR #92, opened 2021-04-01
test: add test for request_encoding

meenal06:

This tests add the following sestets https://github.com/matrix-org/sytest/blob/develop/tests/10apidoc/01request-encoding.pl in Golang

### DIFF
--- a/tests/apidoc_request_encoding_test.go
+++ b/tests/apidoc_request_encoding_test.go
@@ -1,0 +1,29 @@
+
+package tests
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
+)
+
+func TestRequestEncodingFails(t *testing.T) {
+	deployment := Deploy(t, "request_encoding", b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	// sytest: POST rejects invalid utf-8 in JSON
+	t.Run("POST rejects invalid utf-8 in JSON", func(t *testing.T) {
+		res := unauthedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "register"}, json.RawMessage(`{
+			"test": "a` + string(0x81) + `"
+		}`))
+		must.MatchResponse(t, res, match.HTTPResponse{
+			StatusCode: 400,
+			JSON: []match.JSON{
+				match.JSONKeyEqual("errcode", "M_NOT_JSON"),
+			},
+		})
+	})
+}


meenal06:
Hi @callahad @Kegsay , please take a look at the following test I added.

tests/apidoc_request_encoding_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/matrix-org/complement/internal/must"
+)
+
+func TestRequestEncodingFails(t *testing.T) {
+	deployment := Deploy(t, "request_encoding", b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	// sytest: POST rejects invalid utf-8 in JSON
+	t.Run("POST /login can login as user", func(t *testing.T) {
+		res := unauthedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "login"}, json.RawMessage(`{
```
callahad, line 19: Whoops, I think these were copied over from your other PR 🙂  The title should match the Sytest title, and the POST route should be "r0/register" rather than "r0/login".

tests/apidoc_request_encoding_test.go
```diff
@@ -0,0 +1,29 @@
...
+			"test": "a' . chr(0x81) . '"
```
callahad, line 20: I don't think `"a' . chr(0x81) . '"` is valid Go, is it? I know that's what we're using in on the Perl side for SyTest, but I think Go uses different syntax and functions for this.

callahad (changes requested):
Thank you again 🙂 A few changes requested

tests/apidoc_request_encoding_test.go
meenal06: Ahh, my mistake here😓

tests/apidoc_request_encoding_test.go
meenal06: But we just need an invalid json over here right? Why do we need to think about syntax 🤔

tests/apidoc_request_encoding_test.go
callahad: No worries 😄 I think the title should also be "POST rejects invalid utf-8 in JSON", right?

tests/apidoc_request_encoding_test.go
callahad: You make a good argument 🙂 ... but hilariously moving this to Go _does_ create valid JSON, since it literally encodes to the string "`a' . chr(0x81) . '`" 🤣 

I think the intent is to somehow jam a [Unicode Control Character](https://www.fileformat.info/info/unicode/char/0081/index.htm) in there, so we'd need to do something like:

```
"test": "a` + string(0x81) + `"
```

tests/apidoc_request_encoding_test.go
meenal06: Yes, sorry for the mistake again,I will be more cautious next time😅

tests/apidoc_request_encoding_test.go
meenal06: Makes sense, sorry for the silly arguements😅 Amending it right away

meenal06:
Hi @callahad , please do take a look, I have done all requested changes and have squashed my commits 😄

callahad:
I accidentally merged this to master here while trying to test out CI on a different origin. My apologies ☹️ I'll finish reviewing and get this merged properly in just a sec.

meenal06:
Sure, no problem😅

PR #93, opened 2021-04-02
Test MSC3083 - joining a restricted room.

clokep:
This tests matrix-org/synapse#9735 by creating a room that requires membership of a space to join, then attempts to join it (which will fail), then joins the space, then joins the room (which now succeeds).

### DIFF
--- a/dockerfiles/synapse/homeserver.yaml
+++ b/dockerfiles/synapse/homeserver.yaml
@@ -105,3 +105,5 @@ AS_REGISTRATION_FILES
 experimental_features:
   # Enable knocking support
   msc2403_enabled: true
+  # Enable spaces support
+  spaces_enabled: true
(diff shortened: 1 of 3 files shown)


clokep:
i think this probably needs some testing of bad input data and invites.

callahad:
I accidentally merged this to master here while trying to test out CI on a different origin. My apologies ☹️ Will get sorted out later today / tonight.

callahad:
(I've reverted master to its previous state, but GitHub still shows this PR as merged despite no longer being in the DAG for origin/master. What a mess 😬)

clokep:
> (I've reverted master to its previous state, but GitHub still shows this PR as merged despite no longer being in the DAG for origin/master. What a mess 😬)

Yeah even reverting it doesn't let you re-open the same PR. I'll make a new one. 😢

PR #94, opened 2021-04-03
Use GitHub Actions for linting and testing

callahad:
We don't run the Complement Buildkite pipelines for external pull requests due to security concerns around passing the Docker socket through to the container.

...there's no such concern on GitHub's public build infrastructure :)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,19 +10,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-go@v2
-      - uses: actions/cache@v2
-        with:
-          path: |
-            ~/go/pkg/mod
-            ~/.cache/go-build
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-          restore-keys: |
-            ${{ runner.os }}-go-
       - run: sudo apt-get -qq install libolm-dev
-      - run: |
-          export PATH="$(go env GOPATH)/bin:$PATH"
-          ./build/scripts/build-and-lint.sh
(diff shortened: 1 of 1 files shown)


callahad:
Note how the linting and synapse checks run in less than half the time of the same steps on Buildkite :)

callahad:
I accidentally merged this to master here while trying to test out CI on a different origin. My apologies ☹️ I'll finish reviewing and get this merged properly in just a sec.

callahad:
Making the executive decision to merge this: it works, and it allows external contributors to see if their patches pass CI, which is really important with putting Complement forward as a potential Outreachy project

callahad:
Buildkite is failing for unrelated reasons; having connectivity issues to the Docker registery (`docker: Error response from daemon: Get https://registry-1.docker.io/v2/: net/http: request canceled (Client.Timeout exceeded while awaiting headers).`)

Kegsay:
I'm happy with this - but why bother using Buildkite at all at this point then?

callahad:
There are open questions about the suitability / scalability of GitHub Actions for broad use at Element. A few teams are experimenting with it and we should have a more solid recommendation by the end of the month.

The most critical concern right now is that we're limited to 20 concurrent jobs until we upgrade the `matrix-org` organization to GitHub's premium tier for non-profit organizations.

Kegsay:
Okay, well if we do upgrade let me know and I'll deprecate Buildkite for Complement in particular, given the security hole it needs in order to operate correctly.

callahad:
Will do! If we decide to stay with Buildkite, we could patch that security hole by running the host container under [sysbox](https://github.com/nestybox/sysbox), which would allow it to have its own, contained Docker daemon.

PR #95, opened 2021-04-03
tests: add tests for request_encoding

meenal06:


meenal06:
Hi @callahad , the server is giving me 401 unauthorised 🤔 , is there anything I am missing ?

Kegsay:
This test isn't exactly the same as the sytest. https://github.com/matrix-org/sytest/blob/develop/tests/10apidoc/01request-encoding.pl#L9 is the request body. You have `content` and `content_type` there which isn't right.

Furthermore, you want to use `\x81` instead of `fmt.Sprint(0x81)` else Go which just insert the decimal number for `0x81` instead of the literal byte, see https://play.golang.org/p/sIOpvdhzkb-

meenal06:
got it @Kegsay , making the requested changes

meenal06:
Hi @Kegsay , I have fixed the JSON and as well as from the logs I can see that the server is also returning 400 response code. But still tests are failing which I am not able to debug as I have compared the errcode to `M_NOT_JSON` as done in https://github.com/matrix-org/sytest/blob/7f4ce13e7dfb790d943b90de295b13b947bce0f9/tests/10apidoc/01request-encoding.pl#L21 . Can you please let me know what am I doing wrong here 😕 ?

tests/apidoc_request_encoding_test.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+func TestRequestEncodingFails(t *testing.T) {
+	deployment := Deploy(t, "request_encoding", b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	testString := `{ "test": "a' . ` + "\x81" + ` . '" }`
+	// sytest: POST rejects invalid utf-8 in JSON
+	t.Run("POST rejects invalid utf-8 in JSON", func(t *testing.T) {
+		res := unauthedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "register"}, json.RawMessage(testString))
```
Kegsay, line 20: `MustDo` will:

> // MustDo is the same as Do but fails the test if the response is not 2xx

We _want_ the test to return 400, so you cannot use `MustDo` here. Use `Do` instead.

Kegsay (changes requested):

tests/apidoc_request_encoding_test.go
kushthedude: Oops, Thanks for noticing, will fix this asap.

tests/apidoc_request_encoding_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
+)
+
+func TestRequestEncodingFails(t *testing.T) {
+	deployment := Deploy(t, "request_encoding", b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	testString := `{ "test": "a' . ` + "\x81" + ` . '" }`
```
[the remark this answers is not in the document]
meenal06: Oops. thanks for correcting !

meenal06:
Hi @Kegsay , I have fixed the tests please check them!

Kegsay (approved):
LGTM!

PR #98, opened 2021-04-05
Test MSC3083 - joining a restricted room.

clokep:
This tests matrix-org/synapse#9735 by creating a room that requires membership of a space to join, then attempts to join it (which will fail), then joins the space, then joins the room (which now succeeds).

See #93 for why this is being re-created.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -84,6 +84,23 @@ func (c *CSAPI) JoinRoom(t *testing.T, roomIDOrAlias string, serverNames []strin
 	return GetJSONFieldStr(t, body, "room_id")
 }
 
+// LeaveRoom joins the room ID, else fails the test.
+func (c *CSAPI) LeaveRoom(t *testing.T, roomID string) {
+	t.Helper()
+	// leave the room
+	c.MustDoRaw(t, "POST", []string{"_matrix", "client", "r0", "rooms", roomID, "leave"}, nil, "application/json", nil)
+}
+
+// InviteRoom joins the room ID, else fails the test.
+func (c *CSAPI) InviteRoom(t *testing.T, roomID string, userID string) {
+	t.Helper()
+	// Invite the user to the room
+	body := map[string]interface{}{
+    "user_id": userID,
+	}
+	c.MustDo(t, "POST", []string{"_matrix", "client", "r0", "rooms", roomID, "invite"}, body)
+}
+
 // SendEventSynced sends `e` into the room and waits for its event ID to come down /sync.
 // Returns the event ID of the sent event.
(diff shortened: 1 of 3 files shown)


internal/client/client.go
```diff
@@ -84,6 +84,23 @@ func (c *CSAPI) JoinRoom(t *testing.T, roomIDOrAlias string, serverNames []strin
...
 }
 
+// LeaveRoom joins the room ID, else fails the test.
+func (c *CSAPI) LeaveRoom(t *testing.T, roomID string) {
+	t.Helper()
+	// leave the room
+	c.MustDoRaw(t, "POST", []string{"_matrix", "client", "r0", "rooms", roomID, "leave"}, nil, "application/json", nil)
+}
+
+// InviteRoom joins the room ID, else fails the test.
```
Kegsay, line 94:
```suggestion
// InviteRoom invites userID to the room ID, else fails the test.
```

tests/msc3083_test.go
```diff
@@ -0,0 +1,132 @@
...
+	// Join the space, attempt to join the room again, which now should succeed.
+	bob.JoinRoom(t, space, []string{"hs1"})
+	bob.JoinRoom(t, room, []string{"hs1"})
+
+	// Leaving the room works and the user is unable to re-join.
+	bob.LeaveRoom(t, room)
+	bob.LeaveRoom(t, space)
+	FailJoinRoom(bob, t, room, "hs1")
+
+	// Invite the user and joining shuold work.
```
Kegsay, line 89:
```suggestion
	// Invite the user and joining should work.
```

tests/msc3083_test.go
```diff
@@ -0,0 +1,132 @@
...
+	bob := deployment.Client(t, "hs1", "@bob:hs1")
+	FailJoinRoom(bob, t, room, "hs1")
+
+	// Join the space, attempt to join the room again, which now should succeed.
+	bob.JoinRoom(t, space, []string{"hs1"})
+	bob.JoinRoom(t, room, []string{"hs1"})
+
+	// Leaving the room works and the user is unable to re-join.
+	bob.LeaveRoom(t, room)
+	bob.LeaveRoom(t, space)
```
Kegsay, line 86: What happens if you leave the space first?

Kegsay (approved):

tests/msc3083_test.go
clokep: Leaving the space first should be fine. The ordering of these two steps doesn't matter.

clokep:
Although...I now realize that these tests aren't actually running due to matrix-org/pipelines#139, although they won't run anyway until Synapse opts into them. Seems like a chicken-and-egg problem to me. 😢

PR #99, opened 2021-04-07
MSC3083: Add additional tests for joining over federation.

clokep:
This builds on #98 to add tests for matrix-org/synapse#9800.

The tests are pretty much the same, as #98, but the users are no longer on the same server.

### DIFF
--- a/tests/msc3083_test.go
+++ b/tests/msc3083_test.go
@@ -14,7 +14,7 @@ import (
 )
 
 var (
-	msc1772SpaceChildEventType  = "org.matrix.msc1772.space.child"
+	msc1772SpaceChildEventType = "org.matrix.msc1772.space.child"
 )
 
 func FailJoinRoom(c *client.CSAPI, t *testing.T, roomIDOrAlias string, serverName string, expectedHttpCode int) {
--- a/tests/msc3083_test.go
+++ b/tests/msc3083_test.go
@@ -149,8 +149,7 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
 
 	// Create a second user and attempt to join the room, it should fail.
 	bob := deployment.Client(t, "hs2", "@bob:hs2")
-	// Note that this is a 400 error because it comes back over federation.
-	FailJoinRoom(bob, t, room, "hs1", 400)
+	FailJoinRoom(bob, t, room, "hs1", 403)
 
 	// Join the space, attempt to join the room again, which now should succeed.
 	bob.JoinRoom(t, space, []string{"hs1"})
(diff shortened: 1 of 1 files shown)


tests/msc3083_test.go
```diff
@@ -127,5 +126,103 @@ func TestRestrictedRoomsLocalJoin(t *testing.T) {
...
+			},
+		},
+	})
+	alice.SendEventSynced(t, space, b.Event{
+		Type:     msc1772SpaceChildEventType,
+		StateKey: &room,
+		Content: map[string]interface{}{
+			"via": []string{"hs1"},
+		},
+	})
```
richvdh, line 170: is it worth factoring this out to a utility function?

tests/msc3083_test.go
```diff
@@ -127,5 +126,103 @@ func TestRestrictedRoomsLocalJoin(t *testing.T) {
...
+
+	// Create a second user and attempt to join the room, it should fail.
+	bob := deployment.Client(t, "hs2", "@bob:hs2")
+	// Note that this is a 400 error because it comes back over federation.
+	FailJoinRoom(bob, t, room, "hs1", 400)
```
richvdh, line 175: can we assert that a useful error message comes back to the client? I'm reluctant to add yet another cause of obscure "no known servers" error messages.

richvdh: (Indeed, maybe we should spec an error code as part of MSC3083)

tests/msc3083_test.go
```diff
@@ -127,5 +126,103 @@ func TestRestrictedRoomsLocalJoin(t *testing.T) {
...
+	// Join the space, attempt to join the room again, which now should succeed.
+	bob.JoinRoom(t, space, []string{"hs1"})
+	bob.JoinRoom(t, room, []string{"hs1"})
+
+	// Leaving the room works and the user is unable to re-join.
+	bob.LeaveRoom(t, room)
+	bob.LeaveRoom(t, space)
+	FailJoinRoom(bob, t, room, "hs1", 400)
+
+	// Invite the user and joining shuold work.
```
richvdh, line 186:
```suggestion
	// Invite the user and joining should work.
```

tests/msc3083_test.go
clokep: Yeah this part can, the parts with bob cannot since the HS differs between the two tests. :+1:

Kegsay (approved):
Merge when you're done with this @clokep - LGTM

tests/msc3083_test.go
clokep: I originally had this as a 403 also, but it seems Synapse doesn't properly return federation errors back to clients. I'll look further into it though.

clokep: I made this change in Synapse and updated the tests here. This also let us dramatically reduce the copied code between the tests!

I've also updated the spec to mention this error code.

richvdh (approved):
lgtm!

clokep:
This will also need the changes in matrix-org/synapse#9868 so that the returned error code is correct over federation.

clokep:
This seems to be failing since the latest Docker tag for Synapse doesn't have the required functionality (it will be in Synapse 1.33.0 which should RC this week).

clokep:
There's a new latest image, so hopefully this will pass now. 😄

PR #100, opened 2021-04-08
tests: POST /login returns same device_id

meenal06:

### DIFF
--- a/tests/apidoc_login_test.go
+++ b/tests/apidoc_login_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/tidwall/gjson"
 
 	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/client"
 	"github.com/matrix-org/complement/internal/match"
 	"github.com/matrix-org/complement/internal/must"
 )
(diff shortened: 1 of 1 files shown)


meenal06:
Hi @Kegsay @callahad , Please review the following tests. I have also created a new method to create dummy users to use them for login tests.

tests/apidoc_login_test.go
```diff
@@ -41,34 +42,76 @@ func TestLogin(t *testing.T) {
...
 					match.JSONKeyTypeEqual("access_token", gjson.String),
-					match.JSONKeyEqual("home_server", "hs1"),
+					match.JSONKeyEqual("device_id", deviceId),
 				},
 			})
 		})
 	})
 }
+
+func createDummyUser(t *testing.T, unauthedClient *client.CSAPI, userId string) error {
```
Kegsay, line 93: When specifying acronyms, please use upper case letters e.g `userID` not `userId`. This is what many Go developers expect to see, and some linters will fail code without this.

tests/apidoc_login_test.go
```diff
@@ -41,34 +42,76 @@ func TestLogin(t *testing.T) {
...
+	reqBody, err := json.Marshal(map[string]interface{}{
+		"auth": map[string]string{
+			"type": "m.login.dummy",
+		},
+		"username": userId,
+		"password": "superuser",
+	})
+	if err != nil {
+		t.Log("unable to marshal json")
+		return err
```
Kegsay, line 103: When making helper functions like this one, it helps to do as much as possible. Look at all the call sites for `createDummyUser` and see how they all then do `if err != nil` followed by failing the test. I would just change `createDummyUser` to not return `error` and instead do the check here and fail the test if there is an error. This reduces boilerplate required in the test.

Kegsay (approved):
Minor things, but otherwise looks good to me (LGTM)!

tests/apidoc_login_test.go
meenal06: Thanks for the tip, I will be making corrections right away!

meenal06:
Hi @Kegsay , I have the changes you requested please review them

tests/apidoc_login_test.go
```diff
@@ -41,34 +42,67 @@ func TestLogin(t *testing.T) {
...
-					match.JSONKeyTypeEqual("user_id", gjson.String),
+					match.JSONKeyEqual("home_server", "hs1"),
 				},
 			})
-			res = unauthedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "login"}, json.RawMessage(`{
+		})
+		// sytest: POST /login returns the same device_id as that in the request
+		t.Run("POST /login returns the same device_id as that in the request", func(t *testing.T) {
+			t.Parallel()
+			deviceId := "test_device_id"
```
Kegsay, line 65: Again, please use `deviceID`.

Kegsay (approved):

tests/apidoc_login_test.go
meenal06: Apologies, let me fix it quickly 😓

meenal06:
Fixed the deviceID @Kegsay , please review

Kegsay (approved):
LGTM!

PR #101, opened 2021-04-08
Update the Spaces Summary tests (MSC2946) to match the current spec

clokep:
This fixes matrix-org/synapse#9756 by updating the Complement tests to match the current specification. The main differences are:

* There's no longer a `limit` flag.
* The summary only travels "down" to children rooms from the given root.
* The path specifies the unstable identifier in it.

Note that this won't break Dendrite because Dendrite was not passing the tag to run these tests.

### DIFF
--- a/tests/msc2946_test.go
+++ b/tests/msc2946_test.go
@@ -171,7 +171,7 @@ func TestClientSpacesSummary(t *testing.T) {
 			ss2:  2, // ss1,r4
 			r4:   1, // ss2
 		}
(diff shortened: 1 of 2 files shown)


clokep:
Vaguely related to matrix-org/pipelines#139 in that these tests can now be enabled for Synapse.

richvdh (approved):
seems plausible to me

Kegsay (approved):

PR #102, opened 2021-04-12
tests: add all login tests

meenal06:

### DIFF
--- a/tests/apidoc_login_test.go
+++ b/tests/apidoc_login_test.go
@@ -47,7 +47,7 @@ func TestLogin(t *testing.T) {
 				"type": "m.login.password",
 				"identifier": {
 					"type": "m.id.user",
-					"user": "login_test_user"
+					"user": "@login_test_user:hs1"
 				},
 				"password": "superuser"
 			}`))
@@ -68,7 +68,7 @@ func TestLogin(t *testing.T) {
 				"type": "m.login.password",
 				"identifier": {
 					"type": "m.id.user",
-					"user": "device_id_test_user"
+					"user": "@device_id_test_user:hs1"
 				},
 				"password": "superuser",
 				"device_id": "`+deviceID+`"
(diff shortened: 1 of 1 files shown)


tests/apidoc_login_test.go
```diff
@@ -81,6 +82,70 @@ func TestLogin(t *testing.T) {
 				},
 			})
 		})
+		// sytest: POST /login can log in as a user with just the local part of the id
+		t.Run("POST /login can log in as a user with just the local part of the id", func(t *testing.T) {
+			t.Parallel()
+			createDummyUser(t, unauthedClient, "@test_user:matrix")
```
Kegsay, line 88: `createDummyUser` will pass whatever you give it as a `userID` into the request as `username`, which is treated as the user localpart. This means we don't need `@` or `:matrix` here.

tests/apidoc_login_test.go
```diff
@@ -81,6 +82,70 @@ func TestLogin(t *testing.T) {
...
+			reg := regexp.MustCompile("@([^:]*):")
+			localPart := reg.FindStringSubmatch("@test_user:matrix")
```
Kegsay, line 90: Please don't do this. I know the original perl code does this but it's unclear why we need to do this rather than just hard-coding the localpart to "test_user" in this case. I would just do that. The original perl code didn't do this because it dynamically made users so you didn't know at compile time what the user localpart was.

tests/apidoc_login_test.go
```diff
@@ -81,6 +82,70 @@ func TestLogin(t *testing.T) {
...
+			res := unauthedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "login"}, json.RawMessage(`{
+				"type": "m.login.password",
+				"identifier": {
+					"type": "m.id.user",
+					"user": "`+localPart[1]+`"
```
Kegsay, line 95: Just use `"user": "test_user"`

tests/apidoc_login_test.go
```diff
@@ -81,6 +82,70 @@ func TestLogin(t *testing.T) {
 				},
 			})
 		})
+		// sytest: POST /login can log in as a user with just the local part of the id
+		t.Run("POST /login can log in as a user with just the local part of the id", func(t *testing.T) {
```
Kegsay, line 86: We have actually been doing this the entire time! You should go back to the previous tests:
 - POST /login can login as user
 - POST /login returns the same device_id as that in the request

 and add `@` and `:hs1` to the `user` key. E.g:
```
"user": "device_id_test_user"
"user": "@device_id_test_user:hs1"
```

Sorry I didn't catch that earlier.

tests/apidoc_login_test.go
Kegsay: Also, `:matrix` wouldn't be valid. It has to be the domain of the server, which for these tests is `:hs1`.

tests/apidoc_login_test.go
```diff
@@ -81,6 +82,70 @@ func TestLogin(t *testing.T) {
...
+				},
+				"password": "wrong_password"
+			}`), nil)
+
+			if err != nil {
+				t.Fatalf("unable to make request to /login: %v", err)
+			}
+
+			must.MatchResponse(t, res, match.HTTPResponse{
+				JSON: []match.JSON{
```
Kegsay, line 144: Also check for 403 please.

Kegsay (changes requested):

tests/apidoc_login_test.go
meenal06: Hi @Kegsay I just modified the following test to not use createDummyUser and do the stuff we have been doing in the perl test
```go
			res := unauthedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "register"}, json.RawMessage(`{
				"auth": {
					"type": "m.login.dummy"
				},
				"username": "local-login-user",
				"password": "superuser"
			}`))
			body, _ := ioutil.ReadAll(res.Body)
			userID := gjson.Get(string(body), "user_id")

			reg := regexp.MustCompile("@([^:]*):")
			localPart := reg.FindStringSubmatch(userID.Str)
			res = unauthedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "login"}, json.RawMessage(`{
				"type": "m.login.password",
				"identifier": {
					"type": "m.id.user",
					"user": "`+localPart[1]+`"
				},
				"password": "superuser"
			}`))
```

I have got the userID from the response of the register request and then trimmed it as was done in sytest, and tests are passing now too. Does this looks fine 🤔 ?

tests/apidoc_login_test.go
meenal06: @Kegsay  should I return make the createDummyUser return the UserID from the server response and then use that userID as was done in the perl tests?

Kegsay: Yup that sounds good!

tests/apidoc_login_test.go
meenal06: I refactored the creatDummyUser to return userID so we may mimic the similar scenario of attests.

tests/apidoc_login_test.go
meenal06: done

tests/apidoc_login_test.go
meenal06: Since I have made the createDummyUser to return the complete userID, do you still want me to just use the localpart rather than trimming the returned userID?

meenal06:
Hi @Kegsay , I made all the changes you requested however I have restored the createDummyUser at original state as I encountered an failure. But I have made the changes as you requested thoroughly. Please take a look and review !

Kegsay (approved):

PR #103, opened 2021-04-13
tests: add version structure tests

meenal06:

### DIFF
--- a/tests/apidoc_version_test.go
+++ b/tests/apidoc_version_test.go
@@ -1,7 +1,6 @@
 package tests
 
 import (
-	"encoding/json"
 	"fmt"
 	"testing"
 
--- a/tests/apidoc_version_test.go
+++ b/tests/apidoc_version_test.go
@@ -18,7 +18,7 @@ func TestVersionStructure(t *testing.T) {
 	unauthedClient := deployment.Client(t, "hs1", "")
 	// sytest: Version responds 200 OK with valid structure
 	t.Run("Version responds 200 OK with valid structure", func(t *testing.T) {
-		res := unauthedClient.MustDo(t, "GET", []string{"_matrix", "client", "versions"}, json.RawMessage(`{}`))
+		res := unauthedClient.MustDo(t, "GET", []string{"_matrix", "client", "versions"}, nil)
 
 		must.MatchResponse(t, res, match.HTTPResponse{
 			JSON: []match.JSON{
--- a/tests/apidoc_version_test.go
+++ b/tests/apidoc_version_test.go
@@ -1,0 +1,35 @@
+package tests
+
+import (
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


meenal06:
Hi @callahad @Kegsay , please review the following PR. I didn't find a method in our utils to check if the key-type is an array or not, hence I ended up using `match.JSONArrayEach` with an empty function block, as the function has a pre-check for key to be an array type.

tests/apidoc_version_test.go
```diff
@@ -0,0 +1,31 @@
...
+	unauthedClient := deployment.Client(t, "hs1", "")
+	// sytest: Version responds 200 OK with valid structure
+	t.Run("Version responds 200 OK with valid structure", func(t *testing.T) {
+		res := unauthedClient.MustDo(t, "GET", []string{"_matrix", "client", "versions"}, json.RawMessage(`{}`))
+
+		must.MatchResponse(t, res, match.HTTPResponse{
+			JSON: []match.JSON{
+				match.JSONKeyPresent("versions"),
+				match.JSONArrayEach("versions", func(val gjson.Result) error {
+					return nil
```
Kegsay, line 26: You can do more here. The original tests don't assert anything on the elements, but we can at least make sure they are strings by doing:
```go
match.JSONArrayEach("versions", func(val gjson.Result) error {
    if val.Type != gjson.String {
        return fmt.Errorf("value isn't string: %s", val.Raw)
    }
    return nil
}
```

This uses the `gjson` library which we use a lot in Complement - see the functions/fields you can use [here](https://pkg.go.dev/github.com/tidwall/gjson#Result).

meenal06:
Hi @Kegsay I have done the requested changes, please review 😄 !

tests/apidoc_version_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/matrix-org/complement/internal/must"
+)
+
+func TestVersionStructure(t *testing.T) {
+	deployment := Deploy(t, "test_version", b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	// sytest: Version responds 200 OK with valid structure
+	t.Run("Version responds 200 OK with valid structure", func(t *testing.T) {
+		res := unauthedClient.MustDo(t, "GET", []string{"_matrix", "client", "versions"}, json.RawMessage(`{}`))
```
Kegsay, line 21: GET requests will never have a body, so please replace `json.RawMessage(`{}`)` with `nil`.

meenal06: Done!

meenal06:
@Kegsay I did the requested change, please do take a look again!

Kegsay (approved):

PR #104, opened 2021-04-15
tests: add user_presence tests

meenal06:

### DIFF
--- a/tests/apidoc_login_test.go
+++ b/tests/apidoc_login_test.go
@@ -42,7 +42,7 @@ func TestLogin(t *testing.T) {
 		// sytest: POST /login can log in as a user
 		t.Run("POST /login can login as user", func(t *testing.T) {
 			t.Parallel()
-			createDummyUser(t, unauthedClient, "login_test_user")
+			CreateDummyUser(t, unauthedClient, "login_test_user")
 			res := unauthedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "login"}, json.RawMessage(`{
 				"type": "m.login.password",
 				"identifier": {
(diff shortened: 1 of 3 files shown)


meenal06:
Hi @Kegsay  , so to fix the 401, I created an unauthedClient and then made a user through that client. After than I created an authedClient with the userID of the user created from the unauthedClient, but I am getting error that the user was not found. How should I go ahead and approach the following tests 😕 ?

meenal06:
Hi @Kegsay , I have fixed the tests and created authedClient as it was specified in the onboarding docs, Please do take a look. But I still have the previous doubt that why did my earlier method didn't work and what if we want to create a new user which we may need to use with an authedClient?

tests/apidoc_login_test.go
```diff
@@ -149,7 +149,7 @@ func TestLogin(t *testing.T) {
 	})
 }
 
-func createDummyUser(t *testing.T, unauthedClient *client.CSAPI, userID string) {
+func CreateDummyUser(t *testing.T, unauthedClient *client.CSAPI, userID string) {
```
Kegsay, line 152: Please keep this private. You don't need it for the new tests.

tests/apidoc_presence_test.go
```diff
@@ -0,0 +1,26 @@
+package tests
```
Kegsay, line 1: Please add a blacklist in for Dendrite as it does not implement presence yet. E.g https://github.com/matrix-org/complement/blob/33f712f8b1c138210ee9514c32f383f10937bb7a/tests/user_query_keys_test.go#L1 - this comment is read by Go, and it has to be the first line in the file. Replace `synapse` with `dendrite`.

Kegsay: You will also need to add this tag to the Github Actions file - https://github.com/matrix-org/complement/blob/09ac587f5a1c96f056b8723fc1316d1e816c228f/.github/workflows/ci.yaml#L29 - add in `dendrite_blacklist`.

Doing this should make CI pass.

Kegsay (changes requested):

meenal06:
Hi @Kegsay , please review I have made all the changes you requested 😄 !

Kegsay (approved):

PR #105, opened 2021-04-23
Update the Synapse Worker configuration with new base image name and worker configuration

anoadragon453:
This PR updates Complement's Synapse worker-flavoured docker image in a few ways:

* Updates the base image name from `synapse:workers` to `synapse-workers`, as it's built from a separate Dockerfile from the base `matrixdotorg/synapse` image.
* Update the `SYNAPSE_WORKERS` env var name to `SYNAPSE_WORKER_TYPES`
* Switches to an explicit list of worker names. We dropped the `*` functionality after adding sharding capabilities. The worker configuration mirrors [that of sytest](https://github.com/matrix-org/sytest/blob/7d297956d6c5a177d2ea492ce6561b2e2500b291/lib/SyTest/Homeserver/Synapse.pm#L657-L1051).

These changes grew out of [this PR on Synapse](https://github.com/matrix-org/synapse/pull/9162/) which changed the base image.

I'd ideally like feedback from the Synapse team on the chosen worker configuration.

### DIFF
--- a/dockerfiles/SynapseWorkers.Dockerfile
+++ b/dockerfiles/SynapseWorkers.Dockerfile
@@ -47,7 +47,7 @@ ENTRYPOINT \
   # Set postgres authentication details which will be placed in the homeserver config file
   POSTGRES_PASSWORD=somesecret POSTGRES_USER=postgres POSTGRES_HOST=localhost \
   # Specify the workers to test with
-  SYNAPSE_WORKERS=\
+  SYNAPSE_WORKER_TYPES=\
     event_persister, \
     event_persister, \
     background_worker, \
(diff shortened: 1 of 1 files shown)


erikjohnston (approved):
I think this broadly matches current sytest config

anoadragon453:
After looking at the diff again, I don't think this really needs a review from a Complement perspective...

Kegsay:
@anoadragon453 is this ready to merge?

anoadragon453:
Yes apologies for the delay. I just wanted to test the whole thing end-to-end before merging but got hung up on the `/proc/1/cpuset` bug.

I don't think these changes are going to be modified any further however, so we can merge this bit.

PR #107, opened 2021-04-30
Test against stable MSC1772 spaces identifiers

clokep:
~~Based on #99~~

Blocked on FCP matrix-org/matrix-doc#1772.

### DIFF
--- a/tests/msc2946_test.go
+++ b/tests/msc2946_test.go
@@ -13,8 +13,8 @@ import (
 )
 
 var (
-	spaceChildEventType  = "org.matrix.msc1772.space.child"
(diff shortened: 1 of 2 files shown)


clokep:
Annoyingly this will fail until there's a stable release of Synapse that supports this. 😢

Kegsay:
This is presumably because Complement doesn't support same named branches when checking out branches to test under CI?

clokep:
> This is presumably because Complement doesn't support same named branches when checking out branches to test under CI?

Yes, and uses synapse stable instead of develop.

erikjohnston (approved):

PR #108, opened 2021-05-07
Use GET instead of POST for spaces summary.

clokep:
Per changes in matrix-org/matrix-doc#2946.

### DIFF
--- a/tests/msc2946_test.go
+++ b/tests/msc2946_test.go
@@ -4,6 +4,7 @@ package tests
 
 import (
 	"fmt"
+	"net/url"
 	"testing"
 
 	"github.com/matrix-org/complement/internal/b"
@@ -168,7 +169,7 @@ func TestClientSpacesSummary(t *testing.T) {
 	// - Rooms are returned correctly along with the custom fields `room_type`.
 	// - Events are returned correctly.
 	t.Run("query whole graph", func(t *testing.T) {
-		res := alice.MustDo(t, "POST", []string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", root, "spaces"}, map[string]interface{}{})
+		res := alice.MustDo(t, "GET", []string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", root, "spaces"}, map[string]interface{}{})
 		must.MatchResponse(t, res, match.HTTPResponse{
 			JSON: []match.JSON{
 				match.JSONCheckOff("rooms", []interface{}{
@@ -207,9 +208,16 @@ func TestClientSpacesSummary(t *testing.T) {
 		// should omit either R3 or R4 if we start from SS1 because we only return 1 link per room which will be:
 		// SS1 -> SS2
 		// SS2 -> R3,R4 (but only 1 is allowed)
-		res := alice.MustDo(t, "POST", []string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", ss1, "spaces"}, map[string]interface{}{
-			"max_rooms_per_space": 1,
-		})
+		query := make(url.Values, 1)
+		query.Set("max_rooms_per_space", "1")
+		res := alice.MustDoRaw(
+			t,
+			"GET",
+			[]string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", ss1, "spaces"},
+			nil,
+			"application/json",
+			query,
+		)
 		wantItems := []interface{}{
 			ss1ToSS2,
 			ss2ToR3, ss2ToR4, // one of these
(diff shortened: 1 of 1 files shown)


tests/msc2946_test.go
```diff
@@ -207,9 +208,16 @@ func TestClientSpacesSummary(t *testing.T) {
 		// should omit either R3 or R4 if we start from SS1 because we only return 1 link per room which will be:
 		// SS1 -> SS2
 		// SS2 -> R3,R4 (but only 1 is allowed)
-		res := alice.MustDo(t, "POST", []string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", ss1, "spaces"}, map[string]interface{}{
-			"max_rooms_per_space": 1,
-		})
+		query := make(url.Values, 1)
+		query.Set("max_rooms_per_space", "1")
+		res := alice.MustDoRaw(
```
richvdh, line 213: why does this one need a `MustDoRaw` ?

tests/msc2946_test.go
```diff
@@ -168,7 +169,7 @@ func TestClientSpacesSummary(t *testing.T) {
 	// - Rooms are returned correctly along with the custom fields `room_type`.
 	// - Events are returned correctly.
 	t.Run("query whole graph", func(t *testing.T) {
-		res := alice.MustDo(t, "POST", []string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", root, "spaces"}, map[string]interface{}{})
+		res := alice.MustDo(t, "GET", []string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", root, "spaces"}, map[string]interface{}{})
```
richvdh, line 172: isn't this going to try and send `{}` as the body of the request? that seems pretty wrong to me?

clokep: Yeah, it should pass `nil` I suppose. Good call!

tests/msc2946_test.go
clokep: `MustDo` doesn't seem to accept query parameters.

richvdh: oh yes. That's odd, given `Do` does. fair enough.

clokep: I can attempt to pipe it all through if you'd prefer, not sure if we were trying to keep those APIs extremely simple or not (maybe @Kegsay has a thought)?

tests/msc2946_test.go
```diff
@@ -168,7 +169,7 @@ func TestClientSpacesSummary(t *testing.T) {
 	// - Rooms are returned correctly along with the custom fields `room_type`.
 	// - Events are returned correctly.
 	t.Run("query whole graph", func(t *testing.T) {
-		res := alice.MustDo(t, "POST", []string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", root, "spaces"}, map[string]interface{}{})
+		res := alice.MustDo(t, "GET", []string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", root, "spaces"}, nil)
```
richvdh, line 172: I think this makes it send `null` as the GET body, which I'm not sure is much better :/

richvdh (approved):
sending bodies with GETs seems pretty icky, but maybe it's the best you can do without changing half of Complement.

tests/msc2946_test.go
Kegsay: Having seen more people use the client, I would press for functional options at this point. That would avoid the API mismatches that we have here between `MustDo` and `Do`, and make it clearer on the caller side what checks you're doing e.g:
```go
alice.MustDo(t, "POST", []string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", ss1, "spaces"}, map[string]interface{}{
	"max_rooms_per_space": 1,
}, client.WithQueries(query), client.WithReturnStatusCode(200))
```
etc. Exactly what shape this will take I don't know yet - some are assertions and some modiy the request. Should they all be functional options entirely or will that get annoying doing things like:
```go
alice.Do(t,
	client.WithMethod("POST"),
	client.WithPath([]string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", ss1, "spaces"}),
	client.WithBody(map[string]interface{}{
		"max_rooms_per_space": 1,
	}),
	client.WithQueries(query),
	client.WithAuth(),
)
```
This would also make it clearer for cases where you aren't sending a body (as you don't need to awkwardly set `nil` as the arg, you just omit `client.WithBody(..)`. The downside is that the API then isn't guiding the caller into what bits are required vs optional, so maybe a hybrid where method/path are required, but then it's all optional functions?

tests/msc2946_test.go
clokep: I think it causes it to not be sent at all, but I'm not 100% sure on that.

clokep: I think d36faaaa9f16d3f93e46a4bed8a6c95a0b256768 fixes this.

PR #109, opened 2021-05-13
Add a test for restricted rooms appearing in the spaces summary (MSC3083).

clokep:
This adds some additional tests for [MSC3083](matrix-org/matrix-doc#3083) for ensuring that the space summary works for restricted rooms.

See matrix-org/synapse#9922 for the Synapse implementation.

### DIFF
--- a/tests/msc3083_test.go
+++ b/tests/msc3083_test.go
@@ -165,7 +165,6 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
 
 // Request the room summary and ensure the expected rooms are in the response.
 func RequestAndAssertSummary(t *testing.T, user *client.CSAPI, space string, expected_rooms []interface{}) {
-	// The room appears for no one at first since hs2 doesn't know about who is in ss1.
 	res := user.MustDo(t, "POST", []string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", space, "spaces"}, map[string]interface{}{})
 	must.MatchResponse(t, res, match.HTTPResponse{
 		JSON: []match.JSON{
--- a/tests/msc3083_test.go
+++ b/tests/msc3083_test.go
@@ -252,7 +252,9 @@ func TestRestrictedRoomsSpacesSummary(t *testing.T) {
 // Tests that MSC2946 works over federation for a restricted room.
 //
 // Create a space with a room in it that has join rules restricted to membership
-// in that space. The space and room are on different homeservers.
+// in that space. The space and room are on different homeservers. While generating
+// the summary of space hs1 needs to ask hs2 to generate the summary for room since
+// it is not participating in the room.
 //
 // The user should be unable to see the room in the spaces summary unless they
 // are a member of the space.
--- a/tests/msc3083_test.go
+++ b/tests/msc3083_test.go
@@ -258,7 +258,7 @@ func TestRestrictedRoomsSpacesSummary(t *testing.T) {
 // are a member of the space.
 //
 // This tests the interactions over federation where the space and room are on
-// difference homeservers, which might not have the proper information needed to
+// different homeservers, and one might not have the proper information needed to
 // decide if a user is in a room.
 func TestRestrictedRoomsSpacesSummaryFederation(t *testing.T) {
 	deployment := Deploy(t, b.BlueprintFederationTwoLocalOneRemote)
@@ -321,7 +321,7 @@ func TestRestrictedRoomsSpacesSummaryFederation(t *testing.T) {
 	requestAndAssertSummary(t, bob, space, []interface{}{space})
 
 	// charlie joins the space and now hs2 knows that alice is in the space (and
-	// can join room).
+	// can join the room).
 	charlie.JoinRoom(t, space, []string{"hs1"})
 
 	// The restricted room should appear for alice (who is in the space).
--- a/tests/msc3083_test.go
+++ b/tests/msc3083_test.go
@@ -163,6 +163,19 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
 	CheckRestrictedRoom(t, alice, bob, space, room)
 }
 
+// Request the room summary and ensure the expected rooms are in the response.
+func RequestAndAssertSummary(t *testing.T, user *client.CSAPI, space string, expected_rooms []interface{}) {
+	// The room appears for no one at first since hs2 doesn't know about who is in ss1.
+	res := user.MustDo(t, "POST", []string{"_matrix", "client", "unstable", "org.matrix.msc2946", "rooms", space, "spaces"}, map[string]interface{}{})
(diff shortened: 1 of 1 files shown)


tests/msc3083_test.go
```diff
@@ -159,3 +162,159 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
 	// Execute the checks.
 	CheckRestrictedRoom(t, alice, bob, space, room)
 }
+
+// Request the room summary and ensure the expected rooms are in the response.
+func RequestAndAssertSummary(t *testing.T, user *client.CSAPI, space string, expected_rooms []interface{}) {
```
Kegsay, line 167: Keep these private please, there is no need to make it public.

tests/msc3083_test.go
```diff
@@ -159,3 +162,159 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
...
+	// The room appears for no one at first since hs2 doesn't know about who is in ss1.
```
Kegsay, line 168: This comment lies, move it to where you call `RequestAndAssertSummary`

clokep: Good catch! Thanks. :)

tests/msc3083_test.go
clokep: Thanks! I also made the other helpers in this file private.

clokep:
As with other PRs, we'll need to wait for a release before this will pass CI.

tests/msc3083_test.go
```diff
@@ -163,3 +164,160 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
...
+// are a member of the space.
+func TestRestrictedRoomsSpacesSummaryFederation(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintFederationTwoLocalOneRemote)
+	defer deployment.Destroy(t)
+
+	// Create the rooms
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+	bob := deployment.Client(t, "hs1", "@bob:hs1")
+	space := alice.CreateRoom(t, map[string]interface{}{
+		"preset": "public_chat",
```
anoadragon453, line 267: Perhaps name this one `Space` to be consistent with the above test?

tests/msc3083_test.go
```diff
@@ -163,3 +164,160 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
...
+	// about the room? but it works for testing)
+	alice.SendEventSynced(t, space, b.Event{
+		Type:     spaceChildEventType,
+		StateKey: &room,
+		Content: map[string]interface{}{
+			"via": []string{"hs2"},
+		},
+	})
+
+	// The room appears for no one at first since hs2 doesn't know about who is in ss1.
```
anoadragon453, line 313: What does "ss" stand for?

tests/msc3083_test.go
```diff
@@ -163,3 +164,160 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
...
+	requestAndAssertSummary(t, alice, space, []interface{}{space})
+	requestAndAssertSummary(t, bob, space, []interface{}{space})
+
+	// Charlie joins ss1 and now hs2 knows that alice is in it.
+	charlie.JoinRoom(t, space, []string{"hs1"})
+
+	// The restricted room should appear for alice (who is in the space).
```
anoadragon453, line 320: This is the case as hs1 has to ask hs2 if Alice can view the room, and hs2 will only answer positively once hs2 knows that Alice is in the space (once charlie:hs2 has joined it)?

It'd be useful to explain the dependency chain for Alice being able to see the room, as on first glance Alice is in the space, so you'd assume that she'd immediately be able to see the room.

The docstring for this function simplifies this to:

```golang
// The user should be unable to see the room in the spaces summary unless they
// are a member of the space.
```

but it'd be useful to mention how multiple homeservers play into it without diving into the MSC.

anoadragon453 (commented):
The tests look correct from a functional standpoint. I just had a couple questions as someone who's been outside the world of Spaces implementation up until recently.

tests/msc3083_test.go
clokep: "sub-space", but it really just matches the room name.

clokep: Ah this comment is stale from an old version of the code!

tests/msc3083_test.go
clokep: @anoadragon453 I reworded some of these comments (and expanded the docstring). I hope it is clearer now, but please let me know if it can be improved!

tests/msc3083_test.go
```diff
@@ -256,6 +256,10 @@ func TestRestrictedRoomsSpacesSummary(t *testing.T) {
 //
 // The user should be unable to see the room in the spaces summary unless they
 // are a member of the space.
+//
+// This tests the interactions over federation where the space and room are on
+// difference homeservers, which might not have the proper information needed to
```
anoadragon453, line 261:
```suggestion
// different homeservers, and one might not have the proper information needed to
```

tests/msc3083_test.go
```diff
@@ -310,11 +315,13 @@ func TestRestrictedRoomsSpacesSummaryFederation(t *testing.T) {
...
 
-	// The room appears for no one at first since hs2 doesn't know about who is in ss1.
+	// The room appears for neither alice or bob initially. Although alice is in
+	// the space and should be able to access the room, hs2 doesn't know this!
 	requestAndAssertSummary(t, alice, space, []interface{}{space})
 	requestAndAssertSummary(t, bob, space, []interface{}{space})
 
-	// Charlie joins ss1 and now hs2 knows that alice is in it.
+	// charlie joins the space and now hs2 knows that alice is in the space (and
+	// can join room).
```
anoadragon453, line 324:
```suggestion
	// can join the room).
```

anoadragon453, line 324: My main question revolves around why hs2 needs to know alice can join the room if we're asking hs1 for the space summary. hs1 knows alice can join the room already.

My assumption is that hs1 needs to ask hs2 this question, but it'd be good to mention that.

anoadragon453 (commented):
Definitely better! One last question (and a couple corrections).

tests/msc3083_test.go
clokep: Ah, hs1 is asking hs2 about "room" while generating the summary. I'll add this.

clokep: I updated the docstring which talks about the setup of space/room, I hope that's OK. The above is true for every time we attempt to summarize so that seemed like the best spot.

anoadragon453: Thanks! And aha, I think my mental model had Alice creating the room, not Charlie, that's why I was confused about why hs1 didn't know about the status of the room.

That realisation and the additional notes make things fairly clear now, thanks!

anoadragon453 (approved):

PR #111, opened 2021-05-14
Speed up Complement runs

Kegsay:
Every `Deploy()` call used to create a network and the deploy the containers.
It took 4s per network and 4s for containers, so 8s for single HSes, 12s for
federated deployments. Both the networks and containers got torn down at the
end of the test via `Destroy()`.

We don't need to re-create networks all the time. It is sufficient to just
make the network once (similar to how we execute blueprints once per run)
and then clean everything up at the *end of all the tests*. This PR does
just that.

Back of the envelope calculations:
```
Number of Deploy() calls: 23
Number of unique blueprints (hence networks): 4
Number of repeat Deploy() calls: 23-4 = 19

Savings per full Complement run: 19*4 = 76s
```

Also added in logs for `Deploy()` calls to keep an eye on any regressions.

### DIFF
--- a/internal/docker/builder.go
+++ b/internal/docker/builder.go
@@ -617,12 +617,14 @@ func deployImage(
 // CreateNetworkIfNotExists creates a docker network and returns its id.
 // ID is guaranteed not to be empty when err == nil
 func CreateNetworkIfNotExists(docker *client.Client, blueprintName string) (networkID string, err error) {
+	// check if a network already exists for this blueprint
 	nws, err := docker.NetworkList(context.Background(), types.NetworkListOptions{
 		Filters: label("complement_blueprint=" + blueprintName),
 	})
 	if err != nil {
 		return "", fmt.Errorf("%s: failed to list networks. %w", blueprintName, err)
 	}
+	// return the existing network
 	if len(nws) > 0 {
 		return nws[0].ID, nil
 	}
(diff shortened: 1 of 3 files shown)


internal/docker/builder.go
```diff
@@ -612,9 +614,18 @@ func deployImage(
...
+func CreateNetworkIfNotExists(docker *client.Client, blueprintName string) (networkID string, err error) {
+	nws, err := docker.NetworkList(context.Background(), types.NetworkListOptions{
+		Filters: label("complement_blueprint=" + blueprintName),
+	})
+	if err != nil {
+		return "", fmt.Errorf("%s: failed to list networks. %w", blueprintName, err)
+	}
+	if len(nws) > 0 {
+		return nws[0].ID, nil
+	}
```
anoadragon453, line 628: I think just a comment here explaining what this section is doing could go a long way.

I'm also not the biggest fan of abbreviated variable names, though I see that's consistent with the existing code.

anoadragon453 (approved):
lgtm other than some tiny code quality nits. Cool to see timing statistics added as well :)

internal/docker/builder.go
Kegsay: I shall point you to https://github.com/golang/go/wiki/CodeReviewComments#variable-names for why we do this.

anoadragon453 (approved):

PR #113, opened 2021-05-14
Use atomic ints for namespaces

Kegsay:
It's always been a weird quirk that you need to specify a unique namespace string when
deploying containers. This is for parallel tests such that if 2 tests deploy
BlueprintAlice they get unique container names.

This PR changes the API so that we just increment an atomic
counter and use that as the namespace.


anoadragon453 (approved):
Great QoL improvement, thanks!

PR #116, opened 2021-05-21
Fixes for running Synapse in worker mode.

clokep:
See matrix-org/synapse#10039.


dockerfiles/synapse/workers-shared.yaml
```diff
@@ -64,3 +64,5 @@ federation_rr_transactions_per_room_per_second: 9999
 experimental_features:
   # Enable knocking support
   msc2403_enabled: true
+  # Enable spaces support
+  spaces_enabled: true
```
clokep, line 68: No really related, but it makes it consistent with the standard synapse config.

anoadragon453 (approved):

PR #117, opened 2021-05-24
feat: add tests for password change

meenal06:

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -15,8 +15,6 @@ import (
 	"github.com/tidwall/gjson"
 
 	"github.com/matrix-org/complement/internal/b"
-	"github.com/matrix-org/complement/internal/match"
-	"github.com/matrix-org/complement/internal/must"
 )
 
 // RequestOpt is a functional option which will modify an outgoing HTTP request.
@@ -189,18 +187,14 @@ func (c *CSAPI) RegisterUser(t *testing.T, localpart, password string) (userID, 
 		"password": password,
 	}
 	res := c.MustDo(t, "POST", []string{"_matrix", "client", "r0", "register"}, reqBody)
-	must.MatchResponse(t, res, match.HTTPResponse{
-		JSON: []match.JSON{
-			match.JSONKeyTypeEqual("access_token", gjson.String),
-			match.JSONKeyTypeEqual("user_id", gjson.String),
-		},
-	})
+
 	body, err := ioutil.ReadAll(res.Body)
 	if err != nil {
 		t.Fatalf("unable to read response body: %v", err)
 	}
-	userID = gjson.Get(string(body), "user_id").String()
-	accessToken = gjson.Get(string(body), "access_token").String()
+
+	userID = gjson.GetBytes(body, "user_id").Str
+	accessToken = gjson.GetBytes(body, "access_token").Str
 	return userID, accessToken
 }
 
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -181,15 +181,12 @@ func (c *CSAPI) SyncUntil(t *testing.T, since, key string, check func(gjson.Resu
 // return user ID & access token, and fail the test on network error
 func (c *CSAPI) RegisterUser(t *testing.T, localpart, password string) (userID, accessToken string) {
 	t.Helper()
-	reqBody, err := json.Marshal(map[string]interface{}{
+	reqBody := map[string]interface{}{
 		"auth": map[string]string{
 			"type": "m.login.dummy",
 		},
 		"username": localpart,
 		"password": password,
-	})
-	if err != nil {
-		t.Fatalf("unable to marshal json: %v", err)
 	}
 	res := c.MustDo(t, "POST", []string{"_matrix", "client", "r0", "register"}, reqBody)
 	must.MatchResponse(t, res, match.HTTPResponse{
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -15,6 +15,8 @@ import (
 	"github.com/tidwall/gjson"
 
 	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
 )
 
 // RequestOpt is a functional option which will modify an outgoing HTTP request.
@@ -173,6 +175,36 @@ func (c *CSAPI) SyncUntil(t *testing.T, since, key string, check func(gjson.Resu
 			}
 		}
 	}
+}
+
+//RegisterUser will register the user with given parameters and
+// return user ID & access token, and fail the test on network error
+func (c *CSAPI) RegisterUser(t *testing.T, localpart, password string) (userID, accessToken string) {
+	t.Helper()
+	reqBody, err := json.Marshal(map[string]interface{}{
+		"auth": map[string]string{
+			"type": "m.login.dummy",
+		},
+		"username": localpart,
+		"password": password,
+	})
+	if err != nil {
+		t.Fatalf("unable to marshal json: %v", err)
+	}
+	res := c.MustDo(t, "POST", []string{"_matrix", "client", "r0", "register"}, reqBody)
+	must.MatchResponse(t, res, match.HTTPResponse{
+		JSON: []match.JSON{
+			match.JSONKeyTypeEqual("access_token", gjson.String),
+			match.JSONKeyTypeEqual("user_id", gjson.String),
+		},
+	})
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		t.Fatalf("unable to read response body: %v", err)
+	}
+	userID = gjson.Get(string(body), "user_id").String()
+	accessToken = gjson.Get(string(body), "access_token").String()
+	return userID, accessToken
 }
 
 // MustDo will do the HTTP request and fail the test if the response is not 2xx
--- a/tests/account_change_password_test.go
+++ b/tests/account_change_password_test.go
@@ -3,8 +3,6 @@ package tests
 import (
 	"testing"
 
-	"encoding/json"
-
 	"github.com/matrix-org/complement/internal/b"
 	"github.com/matrix-org/complement/internal/client"
 	"github.com/matrix-org/complement/internal/match"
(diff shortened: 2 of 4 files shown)


tests/account_change_password_test.go
```diff
@@ -0,0 +1,46 @@
...
+				"auth": {
+					"type": "m.login.password",
+					"user": "@cant_login_with_old_password:hs1",
+					"password": "superuser"
+				},
+				"new_password": "`+newPassword+`"
+			}`)))
+
+		must.MatchResponse(t, res, match.HTTPResponse{
+			StatusCode: 200,
```
meenal06, line 32: Hi @Kegsay , I did similar things to what you said & what is done in the sytest, I created an user & then provided auth-credentials in the request body of `POST account/password`. I am still getting an 401 in place of 200, should I pass access-key in the auth body too 😕 ?

tests/apidoc_login_test.go
```diff
@@ -85,7 +85,7 @@ func TestLogin(t *testing.T) {
 		t.Run("POST /login can log in as a user with just the local part of the id", func(t *testing.T) {
 			t.Parallel()
 
-			createDummyUser(t, unauthedClient, "local-login-user")
+			CreateDummyUser(t, unauthedClient, "local-login-user")
```
Kegsay, line 88: We keep using this function, so let's move it to the main bit of Complement. Please move this to the `CSAPI` and `Deployment` structs. Something along the lines of:

```go
func (c *CSAPI) RegisterUser(t *testing.T, localpart, password string) (userID string, accessToken string) {
    // does the register request with the parameters given
    // returns the full user ID and access token, fails the test on network error
}
```
Which is then used in `Deployment`:
```go
func (d *Deployment) RegisterUser(t *testing.T, hsName, localpart, password string) *client.CSAPI {
    t.Helper()
    dep, ok := d.HS[hsName]
    if !ok {
       t.Fatalf("Deployment.Client - HS name '%s' not found", hsName)
       return nil
    }
    client := &client.CSAPI{
		BaseURL:          dep.BaseURL,
		Client:           client.NewLoggedClient(t, nil),
		SyncUntilTimeout: 5 * time.Second,
		Debug:            d.Deployer.debugLogging,
    }
    userID, accessToken := client.RegisterUser(t, localpart, password)

    // remember the token so subsequent calls to deployment.Client return the user
    dep.AccessTokens[userID] = accessToken

    client.UserID = userID
    client.AccessToken = accessToken
    return client
}
```
Which can then be used in tests:
```go
deployment := Deploy(t, b.BlueprintAlice)
charles := deployment.RegisterUser(t, "hs1", "charles", "mypassword")
```

tests/account_change_password_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+func TestChangePassword(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	newPassword := "new_superuser"
+	// sytest: After changing password, can't log in with old password
+	t.Run("After changing password, can't log in with old password", func(t *testing.T) {
+		CreateDummyUser(t, unauthedClient, "cant_login_with_old_password")
+		res := unauthedClient.DoFunc(t, "POST", []string{"_matrix", "client", "r0", "account", "password"}, client.WithRawBody(json.RawMessage(`{
```
Kegsay, line 22: I wouldn't use `client.WithRawBody` when you are just sending JSON. Use `client.WithJSONBody` instead, which accepts an `interface{}`. There's a few reasons for this:
 - It's easier to add variables to the JSON body, things like `newPassword` here.
 - It's easier to read and less error prone, as the compiler will complain if you miss a `,` but this isn't true for the raw form.
 - It will encode special characters correctly so it remains valid JSON.

```go
client.WithJSONBody(map[string]interface{} {
    "auth": map[string]interface{}{
        "type": "m.login.password",
        "user": "@cant_login_with_old_password:hs1",
        "password": "superuser",
    },
    "new_password": newPassword,
})

File: tests/account_change_password_test.go (hunk 1 above)
Comment (Kegsay) [reply tests/account_change_password_test.go:32]: Please always paste the error output from the CI run so it's easier for me to help. The way I do it is to:
 - Click the red X on your commit, it'll scroll to a part of the logs. I don't want the entire log output.
 - Scroll up until you start to see homeserver logs. There are start and end markers to make this clear:
```
2021/05/24 19:55:30 ============================================
2021/05/24 19:55:30 e3026ed6adb1614ebe0f46e1891380aca613caaf08975a3f5cdcba8dd52809b6 : Server logs:

[...]

2021/05/24 19:55:30 ============== e3026ed6adb1614ebe0f46e1891380aca613caaf08975a3f5cdcba8dd52809b6 : END LOGS ==============
```
- Go to the logs just before the start of the homeserver logs and grab that error.
```
=== CONT  TestChangePassword/After_changing_password,_can't_log_in_with_old_password
    account_change_password_test.go:31: MatchResponse got status 401 want 200 - http://172.17.0.1:49158/_matrix/client/r0/account/password => {"session":"ptLNVKZxYEzyqrJIgwtrhRZN","flows":[{"stages":["m.login.email.identity"]}],"params":{},"completed":["m.login.password"]}
2021/05/24 19:55:30 ============================================
2021/05/24 19:55:30 e3026ed6adb1614ebe0f46e1891380aca613caaf08975a3f5cdcba8dd52809b6 : Server logs:
```

Complement is generally pretty good with logging responses, and this immediately tells me that your request is okay (you can see `"completed":["m.login.password"]`) but for some reason it is asking for email authentication(!!!). I'll raise this with @anoadragon453

Comment (anoadragon453) [reply tests/account_change_password_test.go:32]: Sorry for the confusion (it even took me a while to figure out what's going on here)! Basically Synapse will only validate your account through the usual User Interactive Authentication flows for your account (such as `m.login.password`) if you provide an `access_token`query parameter in the request to `account/password`.

I've filed an issue [here](https://github.com/matrix-org/synapse/issues/10070) in an attempt to make things clearer for the future. But for right now to use this endpoint, you'll annoyingly need to:

1. Call `/login` with the username and password to retrieve an access token.
2. Provide that access token as a query parameter in the request to `account/password`.
3. Now you should be able to pass User-Interactive Authentication by the completing the `m.login.password` stage.

Hope that makes sense!

File: tests/apidoc_login_test.go (hunk 2 above)
Comment (meenal06) [reply tests/apidoc_login_test.go]: Hi @Kegsay , I did the changes and registered the method RegisterUser with deployment, however I am getting 400 whenever the client is making the POST request to the register endpoint. I have pushed the modified code, (do ignore the commented part, I will restore them later as soon as my util function gets fixed) Please do take a look at the helper functions registered!

File: tests/account_change_password_test.go (hunk 3 above)
Comment (meenal06) [reply tests/account_change_password_test.go]: Done the change!

File: tests/apidoc_login_test.go (hunk 2 above)
Comment (meenal06) [reply tests/apidoc_login_test.go]: Done 😄 !

File: tests/account_change_password_test.go (hunk 1 above)
Comment (meenal06) [reply tests/account_change_password_test.go:35]: Hi @anoadragon453 @Kegsay , I added the acces_token param in the account/password request endpoint. However I am getting similar error logs as above in the updated request too. Is there anything I have missed 😕 ?

File: tests/account_change_password_test.go [hunk 4]
```diff
@@ -0,0 +1,48 @@
...
+	newPassword := "new_superuser"
+	// sytest: After changing password, can't log in with old password
+	t.Run("After changing password, can't log in with old password", func(t *testing.T) {
+		passwordClient := deployment.RegisterUser(t, "hs1", "cant_login_with_old_password", "superuser")
+		reqBody := client.WithJSONBody(t, map[string]interface{}{
+			"auth": map[string]interface{}{
+				"type":         "m.login.password",
+				"user":         "@cant_login_with_old_password:hs1",
+				"password":     "superuser",
+				"access_token": passwordClient.AccessToken,
```
Kegsay, line 27: This isn't how you set an access token on a request. It's done internally by `deployment.Client` or `deployment.RegisterUser`.

tests/account_change_password_test.go
```diff
@@ -0,0 +1,48 @@
...
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	newPassword := "new_superuser"
+	// sytest: After changing password, can't log in with old password
+	t.Run("After changing password, can't log in with old password", func(t *testing.T) {
+		passwordClient := deployment.RegisterUser(t, "hs1", "cant_login_with_old_password", "superuser")
+		reqBody := client.WithJSONBody(t, map[string]interface{}{
+			"auth": map[string]interface{}{
+				"type":         "m.login.password",
+				"user":         "@cant_login_with_old_password:hs1",
```
Kegsay, line 25: Prefer using `passwordClient.UserID` here so we don't have to keep updating these strings if the user ID changes.

internal/client/client.go
```diff
@@ -175,6 +177,33 @@ func (c *CSAPI) SyncUntil(t *testing.T, since, key string, check func(gjson.Resu
...
+	t.Helper()
+	reqBody := map[string]interface{}{
+		"auth": map[string]string{
+			"type": "m.login.dummy",
+		},
+		"username": localpart,
+		"password": password,
+	}
+	res := c.MustDo(t, "POST", []string{"_matrix", "client", "r0", "register"}, reqBody)
+	must.MatchResponse(t, res, match.HTTPResponse{
```
Kegsay, line 192: This is a helper function. You don't need to do assertions here (stuff in the `must` or `match` packages). Note that in all the other functions here we don't do this either. This is because these helper functions are not the main point of the test, and are just a bit of code to help set up the test. It's important to make it clear how the test is set up, and having assertions here can make things harder to read.

internal/client/client.go
```diff
@@ -175,6 +177,33 @@ func (c *CSAPI) SyncUntil(t *testing.T, since, key string, check func(gjson.Resu
...
+		JSON: []match.JSON{
+			match.JSONKeyTypeEqual("access_token", gjson.String),
+			match.JSONKeyTypeEqual("user_id", gjson.String),
+		},
+	})
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		t.Fatalf("unable to read response body: %v", err)
+	}
+	userID = gjson.Get(string(body), "user_id").String()
```
Kegsay, line 202: Use `gjson.GetBytes(body, "user_id").Str` instead. `GetBytes` saves a typecast to `string()` and `.Str` won't coerce the value to be a string if it isn't one (e.g if it was a boolean value, `String()=="false"` but `.Str == ""`.

tests/account_change_password_test.go
Kegsay: As per #complement, I see no issues with the code and don't understand why it's 401ing you still. Run it against Dendrite and if it passes then add this test to synapse blacklist and move on.

anoadragon453: It appears Synapse isn't seeing the `Authorization` header for whatever reason. Do double-check it's being sent with `COMPLEMENT_DEBUG=1`, otherwise I'm also at a loss.

meenal06: Fixed, dendrite logs were more accurate and showed that the error was at the client side which was sending the access_token.

meenal06: However this brings an weird behaviour of complement. user_id & access_token were not getting assigned to the client in registerUser helper, moreover the code was getting stuck at this line https://github.com/matrix-org/complement/blob/cd4a83f40ff66a3c137d923855cec56e84a6c06b/internal/client/client.go#L192 & all the statements ahead of this line were not getting executed.
When I removed these lines 
```go
         JSON: []match.JSON{			
          match.JSONKeyTypeEqual("access_token", gjson.String),			
          match.JSONKeyTypeEqual("user_id", gjson.String),		
       },	
``` 
from my helper function, the tests started passing. Any idea why the error was occuring 😕

Kegsay: The problem is that `must.MatchResponse` will consume the response body, so when you try to read from it later on it won't work. https://github.com/matrix-org/complement/blob/master/internal/must/must.go#L72

Kegsay (approved):

PR #120, opened 2021-06-03
feat: add more change_password_tests

meenal06:

### DIFF
--- a/tests/account_change_password_test.go
+++ b/tests/account_change_password_test.go
@@ -12,12 +12,14 @@ import (
 func TestChangePassword(t *testing.T) {
 	deployment := Deploy(t, b.BlueprintAlice)
 	defer deployment.Destroy(t)
-	passwordClient := deployment.RegisterUser(t, "hs1", "test_change_password_user", "superuser")
+	oldPassword := "superuser"
+	newPassword := "my_new_password"
+	passwordClient := deployment.RegisterUser(t, "hs1", "test_change_password_user", oldPassword)
 	unauthedClient := deployment.Client(t, "hs1", "")
 	// sytest: After changing password, can't log in with old password
 	t.Run("After changing password, can't log in with old password", func(t *testing.T) {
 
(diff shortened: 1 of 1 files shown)


tests/account_change_password_test.go
```diff
@@ -12,39 +12,62 @@ import (
...
 			StatusCode: 403,
 			JSON: []match.JSON{
 				match.JSONKeyEqual("errcode", "M_FORBIDDEN"),
 			},
 		})
 	})
+	// sytest: After changing password, can log in with old password
+	t.Run("After changing password, can log in with old password", func(t *testing.T) {
+
+		changePassword(passwordClient, "my_new_password", "new_password", t)
```
meenal06, line 41: @Kegsay I have initialised both clients outside the tests & then I am using the same authedClient for each & every test iteration. However the issue over here is each call to changePassword in the tests changes the old_password to new_password, then I need to refer the new_password rather than the original password in the next iteration of test. So is it okay as of now or should initialise new passwordClient in each of the test case?

tests/account_change_password_test.go
```diff
@@ -12,39 +12,62 @@ import (
...
+		res := unauthedClient.DoFunc(t, "POST", []string{"_matrix", "client", "r0", "login"}, reqBody)
 		must.MatchResponse(t, res, match.HTTPResponse{
 			StatusCode: 403,
 			JSON: []match.JSON{
 				match.JSONKeyEqual("errcode", "M_FORBIDDEN"),
 			},
 		})
 	})
+	// sytest: After changing password, can log in with old password
+	t.Run("After changing password, can log in with old password", func(t *testing.T) {
```
Kegsay, line 39: This doesn't make sense. Shouldn't this be `After changing password, can log in with new password`?

tests/account_change_password_test.go
Kegsay: Don't bother changing the password again, just try logging in with the `newPassword`.

Kegsay (changes requested):

PR #122, opened 2021-06-03
feat: implement all filter tests

meenal06:

### DIFF
--- a/tests/sync_filter_test.go
+++ b/tests/sync_filter_test.go
@@ -29,7 +29,7 @@ func TestSyncFilter(t *testing.T) {
 		if err != nil {
 			t.Fatalf("failed to marshal JSON request body: %s", err)
 		}
-		createFilter(authedClient, reqBody, t, "@alice:hs1")
+		createFilter(t, authedClient, reqBody, "@alice:hs1")
 	})
 	t.Run("Can download filter", func(t *testing.T) {
 		reqBody, err := json.Marshal(map[string]interface{}{
@@ -42,7 +42,7 @@ func TestSyncFilter(t *testing.T) {
 		if err != nil {
 			t.Fatalf("failed to marshal JSON request body: %s", err)
 		}
-		filterID := createFilter(authedClient, reqBody, t, "@alice:hs1")
+		filterID := createFilter(t, authedClient, reqBody, "@alice:hs1")
 		res := authedClient.MustDo(t, "GET", []string{"_matrix", "client", "r0", "user", "@alice:hs1", "filter", filterID}, nil)
 		must.MatchResponse(t, res, match.HTTPResponse{
 			JSON: []match.JSON{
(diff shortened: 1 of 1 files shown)


tests/sync_filter_test.go
```diff
@@ -2,35 +2,47 @@ package tests
...
-		},
+		})
+		if err != nil {
+			t.Fatalf("failed to marshal JSON request body: %s", err)
+		}
+		create_filter(authedClient, reqBody, t, "@alice:hs1")
 	})
+}
+
+func create_filter(authedClient *client.CSAPI, reqBody []byte, t *testing.T, userID string) string {
```
Kegsay, line 34: please don't use `_` in function names. I don't see why this function needs to be separate, where else do we call it?

Kegsay (commented):
You haven't added any new tests here, are you missing a new file?

tests/sync_filter_test.go
```diff
@@ -14,23 +18,54 @@ func TestSyncFilter(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("failed to marshal JSON request body: %s", err)
+		}
+		filterID := createFilter(authedClient, reqBody, t, "@alice:hs1")
+		res := authedClient.MustDo(t, "GET", []string{"_matrix", "client", "r0", "user", "@alice:hs1", "filter", filterID}, nil)
 		must.MatchResponse(t, res, match.HTTPResponse{
 			JSON: []match.JSON{
-				match.JSONKeyPresent("filter_id"),
+				match.JSONKeyPresent("room"),
+				match.JSONKeyEqual("room.timeline.limit", "10"),
```
meenal06, line 50: @Kegsay I am getting error
```sh
=== CONT  TestSyncFilter/Can_download_filter
    sync_filter_test.go:47: MatchResponse key 'room.timeline.limit' got '10' want '10' - http://localhost:55093/_matrix/client/r0/user/@alice:hs1/filter/1 => {"event_format":"client","presence":{"limit":20},"account_data":{"limit":20},"room":{"ephemeral":{"limit":20},"state":{"limit":2147483647},"timeline":{"limit":10},"account_data":{"limit":20}}}
```,
I even tried to remove 10 from the double quotes but still the error persisted. Any idea where I am going wrong?

Comment (kegsay) [reply tests/sync_filter_test.go:50]: Interesting point. So when Go deserialises bytes into JSON, it converts all numbers into `float64`:

> float64, for JSON numbers

https://golang.org/pkg/encoding/json/#Unmarshal

When you use `JSONKeyEqual` it will type-check and value-check the value, in this case `10`. Referring to a number directly in Go code defaults to type `int` as per the [Go specification](https://golang.org/ref/spec#Integer_literals). This is why removing the double quotes doesn't work. Instead, type-cast `10` to float64 e.g `float64(10)`.

File: tests/sync_filter_test.go [hunk 3]
```diff
@@ -14,23 +18,55 @@ func TestSyncFilter(t *testing.T) {
...
+		}
+		createFilter(t, authedClient, reqBody, "@alice:hs1")
 	})
-	if err != nil {
-		t.Fatalf("failed to marshal JSON request body: %s", err)
-	}
-	t.Run("Can create filter", func(t *testing.T) {
-		res := authedClient.MustDo(t, "POST", []string{"_matrix", "client", "r0", "user", "@alice:hs1", "filter"}, json.RawMessage(reqBody))
-
+	t.Run("Can download filter", func(t *testing.T) {
```
kegsay, line 34: This test is missing a `// sytest` comment.

tests/sync_filter_test.go
meenal06: Thanks! Got it!

kegsay (approved):

PR #125, opened 2021-06-11
Use stable prefixes / endpoints for knocking.

clokep:
Updates knocking tests to use stable:

* Prefixes
* Room versions
* Endpoints

I chose not to remove the build flag since it allows servers to opt into it being enabled.


clokep:
See matrix-org/synapse#10167 for the corresponding Synapse PR.

anoadragon453 (approved):
> I chose not to remove the build flag since it allows servers to opt into it being enabled.

As the MSC is merged to the spec now, we should probably remove this build tag (and update the file name) and have homeserver implementations blacklist the tests instead.

That said, it should probably happen in a separate PR.

PR #126, opened 2021-06-12
feat: session related change-password tests

meenal06:

### DIFF
--- a/tests/account_change_password_test.go
+++ b/tests/account_change_password_test.go
@@ -51,6 +51,16 @@ func TestChangePassword(t *testing.T) {
 		res := unauthedClient.DoFunc(t, "POST", []string{"_matrix", "client", "r0", "login"}, reqBody)
 		must.MatchResponse(t, res, match.HTTPResponse{
 			StatusCode: 200,
+			JSON: []match.JSON{
+				match.JSONKeyEqual("user_id", passwordClient.UserID),
+			},
+		})
+	})
+	// sytest: After changing password, existing session still works
+	t.Run("After changing password, existing session still works", func(t *testing.T) {
+		res := passwordClient.DoFunc(t, "GET", []string{"_matrix", "client", "r0", "account", "whoami"})
+		must.MatchResponse(t, res, match.HTTPResponse{
+			StatusCode: 200,
 		})
 	})
 }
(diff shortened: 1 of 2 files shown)


meenal06:
@kegsay Can you please tell me how can I get multiple sessions for a user l=https://github.com/matrix-org/sytest/blob/eedac420ae8c810f48200e7a4ba6a3bc85958f35/tests/14account/01change-password.pl#L93 like done for this one?

kegsay:
You need to call `/login` again. See https://matrix.org/docs/spec/client_server/r0.6.0#post-matrix-client-r0-login

meenal06:
> You

But isnt it trying to start an another session? We need to see if any other session already started should be invalidated except for the existing ones which was used to change password 🤔

kegsay:
> trying to start an another session

This is what you asked for:

> Can you please tell me how can I get multiple sessions for a user

If you want to check session invalidation AFTER changing your password, then make the 2nd session before changing your password.

meenal06:
@kegsay Please review the updated PR.

tests/account_change_password_test.go
```diff
@@ -72,3 +118,24 @@ func changePassword(t *testing.T, passwordClient *client.CSAPI, oldPassword stri
 		StatusCode: 200,
 	})
 }
+
+func createSession(t *testing.T, unauthedClient *client.CSAPI, userID, password string) *client.CSAPI {
+	authedClient := unauthedClient
```
kegsay, line 123: Make a new client instead of accepting an unauthed client. You're passing this client around by pointer so it could become very confusing and is likely not what you intended:
```go
unauthedClient := deployment.Client(t, "hs1", "")
sessionTest := createSession(t, unauthedClient, "test_change_password_user", "superuser")
// sessionTest == unauthedClient
```
so any requests using unauthedClient *will have auth*. To fix this, replace `unauthedClient` with `deployment` from the function signature and instead do `unauthedClient := deployment.Client(t, "hs1", "")`

kegsay (commented):
Other than that, this is great, thank you! You might want to `go run sytest_coverage.go -v` and update the README to get a feel for progress :)

tests/account_change_password_test.go
meenal06: Got it 👍

kegsay (approved):

PR #130, opened 2021-06-19
feat: add display name test

meenal06:

### DIFF
--- a/tests/apidoc_profile_displayname_test.go
+++ b/tests/apidoc_profile_displayname_test.go
@@ -1,0 +1,46 @@
+package tests
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
+)
+
+func TestProfileDisplayName(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	authedClient := deployment.RegisterUser(t, "hs1", "display_name_user", "superuser")
+	displayName := "my_display_name"
+	// sytest: PUT /profile/:user_id/displayname sets my name
+	t.Run("PUT /profile/:user_id/displayname sets my name", func(t *testing.T) {
+		setDisplayName(t, authedClient, displayName)
+		res := unauthedClient.DoFunc(t, "GET", []string{"_matrix", "client", "r0", "profile", authedClient.UserID, "displayname"})
+		must.MatchResponse(t, res, match.HTTPResponse{
+			StatusCode: 200,
+			JSON: []match.JSON{
+				match.JSONKeyEqual("displayname", displayName),
+			},
+		})
+	})
+	// sytest: GET /profile/:user_id/displayname publicly accessible
+	t.Run("GET /profile/:user_id/displayname publicly accessible", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


meenal06:
@kegsay please review

tests/apidoc_profile_displayname_test.go
```diff
@@ -0,0 +1,46 @@
...
+		must.MatchResponse(t, res, match.HTTPResponse{
+			StatusCode: 200,
+			JSON: []match.JSON{
+				match.JSONKeyEqual("displayname", displayName),
+			},
+		})
+	})
+}
+
+func setDisplayName(t *testing.T, authedClient *client.CSAPI, displayName string) {
```
kegsay, line 42: Having a function which just calls another function is called a proxy function. They are generally bad because:
 - It doubles the API surface of things you need to remember as you can do the same thing in multiple ways.
 - It adds extra code needlessly.

Please just directly call `MustDoFunc`, and use `client.WithJSONBody` not `json.RawMessage`.

tests/apidoc_profile_displayname_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
+)
+
+func TestProfileDisplayName(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	authedClient := deployment.RegisterUser(t, "hs1", "display_name_user", "superuser")
```
kegsay, line 17: Why? We aren't testing registration/login anymore where we need to do this. You have a built-in "alice" client available, use it.

tests/apidoc_profile_displayname_test.go
```diff
@@ -0,0 +1,46 @@
...
+		must.MatchResponse(t, res, match.HTTPResponse{
+			StatusCode: 200,
+			JSON: []match.JSON{
+				match.JSONKeyEqual("displayname", displayName),
+			},
+		})
+	})
+	// sytest: GET /profile/:user_id/displayname publicly accessible
+	t.Run("GET /profile/:user_id/displayname publicly accessible", func(t *testing.T) {
+		res := unauthedClient.DoFunc(t, "GET", []string{"_matrix", "client", "r0", "profile", "@display_name_user:hs1", "displayname"})
```
kegsay, line 32: Just use `authedClient.UserID` instead of literally `"@display_name_user:hs1"` for consistency with the previous test. The only change should be that you're using an `unauthedClient`.

kegsay (changes requested):

meenal06:
@kegsay All the requested changes are addressed over here

tests/apidoc_profile_displayname_test.go
```diff
@@ -0,0 +1,42 @@
...
+	authedClient := deployment.Client(t, "hs1", "@alice:hs1")
+	displayName := "my_display_name"
+	// sytest: PUT /profile/:user_id/displayname sets my name
+	t.Run("PUT /profile/:user_id/displayname sets my name", func(t *testing.T) {
+		reqBody := client.WithJSONBody(t, map[string]interface{}{
+			"displayname": displayName,
+		})
+		_ = authedClient.MustDoFunc(t, "PUT", []string{"_matrix", "client", "r0", "profile", authedClient.UserID, "displayname"}, reqBody)
+		res := unauthedClient.DoFunc(t, "GET", []string{"_matrix", "client", "r0", "profile", authedClient.UserID, "displayname"})
+		must.MatchResponse(t, res, match.HTTPResponse{
```
kegsay, line 25: Surely these two lines are doing the test below? We don't need them.

meenal06: Fixed

kegsay (approved):

PR #131, opened 2021-06-21
Test that knocks and invite rejections are sent out correctly over federation

richvdh:
This is a regression test for https://github.com/matrix-org/synapse/issues/10222, and also extends the knock tests to ensure we don't do the same thing there.

This needed a bit of groundwork in the test library, as well as requiring that gomatrixserverlib support room v7, so is dependent on https://github.com/matrix-org/gomatrixserverlib/pull/263.


richvdh:
this is failing in GHA because `matrix-synapse:latest` doesn't yet include https://github.com/matrix-org/synapse/pull/10223.

internal/federation/handle.go
```diff
@@ -128,6 +128,52 @@ func HandleMakeSendJoinRequests() func(*Server) {
...
+				return
+			}
+
+			if inviteCallback != nil {
+				inviteCallback(inviteRequest.Event())
+			}
+
+			// Send the response
+			res := map[string]interface{}{
+				"event": inviteRequest.Event(),
```
kegsay, line 168: Shouldn't we be signing this event or something?

richvdh: I guess so. The spec says we should, though synapse doesn't enforce it.

richvdh: ... but it's a real pita to do so because gomatrixserverlib doesn't seem to expose the methods needed to do so?

tests/federation_room_invite_test.go
```diff
@@ -0,0 +1,78 @@
...
+	deployment := Deploy(t, b.BlueprintFederationTwoLocalOneRemote)
+	defer deployment.Destroy(t)
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+	charlie := deployment.Client(t, "hs2", "@charlie:hs2")
+
+	waiter := NewWaiter()
+	srv := federation.NewServer(t, deployment,
+		federation.HandleKeyRequests(),
+		federation.HandleTransactionRequests(func(ev *gomatrixserverlib.Event) {
+			defer waiter.Finish()
```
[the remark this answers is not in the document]
richvdh: I guess so. Done.

tests/msc2403_test.go
```diff
@@ -40,6 +45,20 @@ func TestKnocking(t *testing.T) {
 	charlieUserID := "@charlie:hs2"
 	charlie := deployment.Client(t, "hs2", charlieUserID)
 
+	// Create a server to observe
```
anoadragon453, line 48: Since it might not be immediately obvious, I might just state here that the purpose of this third server is to ensure that knocking events are observed by all homeservers in the room, not just those are knocking and sending the knock event into the room on behalf of the knocking server.

internal/federation/handle.go
anoadragon453: I think `inviteRequest.Event().Sign()` may work? https://github.com/matrix-org/gomatrixserverlib/blob/45159d0c62d0295dddeed9b951fd6ccbba1ca238/event.go#L558-L574

anoadragon453 (approved):

tests/msc2403_test.go
kegsay: It was clear to me through the use of the word "observe" in this comment.

tests/federation_room_invite_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	srv := federation.NewServer(t, deployment,
+		federation.HandleKeyRequests(),
+		federation.HandleTransactionRequests(func(ev *gomatrixserverlib.Event) {
+			sk := "<nil>"
+			if ev.StateKey() != nil {
+				sk = *ev.StateKey()
+			}
+			t.Logf("Received PDU %s/%s", ev.Type(), sk)
+			if waiter != nil && ev.Type() == "m.room.member" && ev.StateKey() != nil && *ev.StateKey() == charlie.UserID {
```
kegsay, line 37: You can use `ev.StateKeyEquals(charlie.UserID)` which does the state key nil check for you: https://pkg.go.dev/github.com/matrix-org/gomatrixserverlib#Event.StateKeyEquals

tests/federation_room_invite_test.go
```diff
@@ -0,0 +1,83 @@
...
+	// Charlie rejects the invite; Delia should see the rejection.
+	waiter = NewWaiter()
+	charlie.LeaveRoom(t, roomID)
+	waiter.Wait(t, 5*time.Second)
+	if err := checkMembershipForUser(room, charlie.UserID, "leave"); err != nil {
+		t.Errorf("Membership state for charlie after reject: %v", err)
+	}
+}
+
+func checkMembershipForUser(room *federation.ServerRoom, userID, wantMembership string) (err error) {
```
kegsay, line 68: This is a sufficiently useful and common function it should probably live on `federation.ServerRoom`. Filed https://github.com/matrix-org/complement/issues/137

kegsay (commented):
There's a bunch of functions on [Event](https://pkg.go.dev/github.com/matrix-org/gomatrixserverlib#Event) which are generally useful, specifically `Sign()` and `StateKeyEquals()` for this PR. For `Sign()`, the complement server exposes the fields you need https://github.com/matrix-org/complement/blob/63b6c63d9e070313a0c0b1248bb3d478fb765a3e/internal/federation/server.go#L38

richvdh:
> this is failing in GHA because `matrix-synapse:latest` doesn't yet include [matrix-org/synapse#10223](https://github.com/matrix-org/synapse/pull/10223).

This is still the case: matrix-org/synapse#10223 is in Synapse 1.38.0 which is still in RC. I've confirmed the tests pass against 1.38.0rc1.

tests/federation_room_invite_test.go
richvdh: I'm happy to move it, but presumably you'd prefer I did this as a separate PR?

richvdh:
@anoadragon453 @kegsay thanks for the feedback - PTAL?

tests/federation_room_invite_test.go
kegsay: Yup don't worry about it now, I've filed #137 for it.

kegsay (approved):

PR #132, opened 2021-06-23
Test validation on `send_{join,knock,leave}`

richvdh:
Add some tests that try sending things other than proper events via `send_{join,knock,leave}`

### DIFF
--- a/internal/federation/server.go
+++ b/internal/federation/server.go
@@ -155,6 +155,23 @@ func (s *Server) FederationClient(deployment *docker.Deployment) *gomatrixserver
 	return f
 }
 
+// SendFederationRequest signs and sends an arbitrary federation request from this server.
+//
+// The requests will be routed according to the deployment map in `deployment`.
+func (s *Server) SendFederationRequest(deployment *docker.Deployment, req gomatrixserverlib.FederationRequest, resBody interface{}) error {
+	if err := req.Sign(gomatrixserverlib.ServerName(s.ServerName), s.KeyID, s.Priv); err != nil {
+		return err
+	}
+
+	httpReq, err := req.HTTPRequest()
+	if err != nil {
+		return err
+	}
+
+	httpClient := gomatrixserverlib.NewClient(gomatrixserverlib.WithTransport(&docker.RoundTripper{Deployment: deployment}))
+	return httpClient.DoRequestAndParseResponse(context.Background(), httpReq, resBody)
+}
+
 // MustCreateEvent will create and sign a new latest event for the given room.
 // It does not insert this event into the room however. See ServerRoom.AddEvent for that.
 func (s *Server) MustCreateEvent(t *testing.T, room *ServerRoom, ev b.Event) *gomatrixserverlib.Event {
--- a/tests/federation_room_join_test.go
+++ b/tests/federation_room_join_test.go
@@ -2,10 +2,12 @@ package tests
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
 	"testing"
 
+	"github.com/matrix-org/gomatrix"
 	"github.com/matrix-org/gomatrixserverlib"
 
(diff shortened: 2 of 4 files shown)


richvdh:
based on #131

internal/federation/server.go
```diff
@@ -481,3 +525,12 @@ func (f *basicKeyFetcher) FetchKeys(
 func (f *basicKeyFetcher) FetcherName() string {
 	return "basicKeyFetcher"
 }
+
+// SupportedRoomVersions is a convenience method which returns a list of the room versions supported by gomatrixserverlib.
+func SupportedRoomVersions() []gomatrixserverlib.RoomVersion {
+	supportedRoomVersions := make([]gomatrixserverlib.RoomVersion, 0, 10)
```
babolivier, line 531: Why a capacity of 10 ooc?

tests/federation_room_join_test.go
```diff
@@ -192,3 +194,124 @@ func TestJoinFederatedRoomWithUnverifiableEvents(t *testing.T) {
...
+			Content: map[string]interface{}{"body": "bzz"},
+		})
+		assertRequestFails(t, event)
+	})
+	t.Run("non-state membership event", func(t *testing.T) {
+		event := srv.MustCreateEvent(t, room, b.Event{
+			Type:    "m.room.message",
+			Sender:  charlie,
+			Content: map[string]interface{}{"body": "bzz"},
+		})
```
babolivier, line 286: This doesn't look like a non-state membership event?

tests/msc2403_test.go
```diff
@@ -426,5 +465,13 @@ func publishAndCheckRoomJoinRule(t *testing.T, c *client.CSAPI, roomID, expected
 	if !roomFound {
 		t.Fatalf("Room was not present in public room directory response")
 	}
+}
 
+// TestCannotSendNonKnockViaSendKnock checks that we cannot submit anything via /send_knock except a knock
+func TestCannotSendNonKnockViaSendKnock(t *testing.T) {
+	testValidationForSendMembershipEndpoint(t, "knock", "/_matrix/federation/v2/send_leave",
```
babolivier, line 472: It looks like the expected membership and the path aren't passed in the right order.

tests/federation_room_join_test.go
```diff
@@ -192,3 +194,124 @@ func TestJoinFederatedRoomWithUnverifiableEvents(t *testing.T) {
...
+// This test checks that we cannot submit anything via /v2/send_leave except a leave.
+func TestCannotSendNonLeaveViaSendLeaveV2(t *testing.T) {
+	testValidationForSendMembershipEndpoint(t, "/_matrix/federation/v2/send_leave", "leave", nil)
+}
+
+// testValidationForSendMembershipEndpoint attempts to submit a range of events via the given endpoint
+// and checks that they are all rejected.
+func testValidationForSendMembershipEndpoint(t *testing.T, baseApiPath, expectedMembership string, createRoomOpts map[string]interface{}) {
+	if createRoomOpts == nil {
+		createRoomOpts = map[string]interface{}{}
```
babolivier, line 222:
```suggestion
		createRoomOpts = make(map[string]interface{})
```

Not a functional change but this is more idiomatic when you're not shoving content inside of the map when instantiating it.

babolivier (changes requested):

internal/federation/server.go
richvdh: good question. mostly because I forgot that capacity was optional.

tests/federation_room_join_test.go
richvdh: true!

internal/federation/server.go
richvdh: ... though this change has come from #131 rather than here.

erikjohnston (approved):

PR #133, opened 2021-06-23
Allow running against older Synapse releases

richvdh:
This is super useful if you're trying to bisect for where a bug got introduced (or fixed, in my case...)

### DIFF
--- a/dockerfiles/synapse/start.sh
+++ b/dockerfiles/synapse/start.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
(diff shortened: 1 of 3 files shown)


dockerfiles/synapse/start.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
richvdh, line 1: older versions of the Synapse docker image didn't include a /bin/bash, and this script works fine with regular sh, so I'm changing to that.

anoadragon453 (approved):
Quite a useful addition -- thank you!

PR #134, opened 2021-06-23
Check for unauthed events via send_join

richvdh:
Based on #131

### DIFF
--- a/tests/federation_room_join_test.go
+++ b/tests/federation_room_join_test.go
@@ -1,11 +1,14 @@
 package tests
 
 import (
+	"context"
 	"encoding/json"
 	"net/http"
 	"net/url"
 	"testing"
+	"time"
 
+	"github.com/matrix-org/gomatrix"
 	"github.com/matrix-org/gomatrixserverlib"
 
 	"github.com/tidwall/sjson"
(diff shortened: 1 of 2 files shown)


richvdh:
This will fail until https://github.com/matrix-org/synapse/pull/10243 lands.

erikjohnston (approved):

tests/federation_room_join_test.go
```diff
@@ -192,3 +195,70 @@ func TestJoinFederatedRoomWithUnverifiableEvents(t *testing.T) {
...
+		Sender:   alice.UserID,
+		StateKey: &charlie,
+		Content: map[string]interface{}{
+			"membership": "ban",
+		},
+	})
+
+	// charlie sends a make_join for a different user
+	makeJoinResp, err := fedClient.MakeJoin(context.Background(), "hs1", roomID, srv.UserID("charlie2"), federation.SupportedRoomVersions())
+	must.NotError(t, "MakeJoin", err)
```
kegsay, line 230: So Dendrite will fail the test here with `M_FORBIDDEN: eventauth: "@charlie2:host.docker.internal" is not allowed to change their membership from "leave" to "join"` which feels like reasonable behaviour. Annoyingly, Complement CI seems to not be running Dendrite tests through to completion causing this not to be picked up.

Currently Complement tests now fail on the Dendrite repo as a result.

richvdh: I think this should work? it should be a public room so `make_join` should work?

kegsay: The root cause here was that `/createRoom` with `{}` defaults to private rooms as per the spec:

> A public visibility indicates that the room will be shown in the published room list. A private visibility will hide the room from the published room list. Rooms default to private visibility if this key is not included. 

But synapse defaults to public it seems.

richvdh: fixed by #158

PR #123, opened 2021-06-10
Avoid unnecessary work in the spaces summary.

clokep:
Updates tests for matrix-org/synapse#10085.

PR #138, opened 2021-06-25
Do not recurse into non-spaces in the spaces summary.

clokep:
This is a subset of #123 and corresponds to matrix-org/synapse#10256, which is the straightforward change to only recurse into spaces, not the bit about trimming unnecessary events.

### DIFF
--- a/tests/msc2946_test.go
+++ b/tests/msc2946_test.go
@@ -31,8 +31,8 @@ func eventKey(srcRoomID, dstRoomID, evType string) string {
 // _____|________
 // |    |       |
 // R1  SS1      R2
-//      |
-//     SS2
+//      |       |
+//     SS2      R5
 //      |________
 //      |       |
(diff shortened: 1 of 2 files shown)


tests/msc2946_test.go
```diff
@@ -134,6 +143,14 @@ func TestClientSpacesSummary(t *testing.T) {
 			"via": []string{"hs1"},
 		},
 	})
+	// Note that this link gets ignored since R2 is not a space.
```
babolivier, line 146: Where do we verify that assumption?

babolivier (approved):
lgtm otherwise

tests/msc2946_test.go
clokep: Near lines 213 - 219 below we do a `match.JSONCheckOff` call against `events` and do not include this.

PR #140, opened 2021-06-28
Added Apple Silicon instructions to README

Azrenbeth:
I ran into some trouble getting Complement to run on an M1 Mac. It wasn't able to find libolm - which was installed using brew. Thanks to @tulir, this fixed it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ or on Mac:
 ```
 brew install libolm
 ```
-If you are on an Apple Silicon Mac then you'll need to set these environment variables too:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,6 +21,12 @@ or on Mac:
 ```
 brew install libolm
 ```
+If you are on an Apple Silicon Mac then you'll need to set these environment variables too:
```
kegsay, line 24:
```suggestion
If you are on an Apple Silicon Mac then you'll need to set these environment variables too so Go can find `libolm`:
```

kegsay (approved):

PR #143, opened 2021-07-01
Bomb out sooner if we cannot start a federation server

richvdh:
If we can't start our test federation server, there's no point going on.

(I seem to keep forgetting to shut down the thing I have on port 8448 and then wondering why my test is failing.)

### DIFF
--- a/internal/federation/server.go
+++ b/internal/federation/server.go
@@ -205,11 +205,18 @@ func (s *Server) Mux() *mux.Router {
 func (s *Server) Listen() (cancel func()) {
 	var wg sync.WaitGroup
 	wg.Add(1)
+
+	ln, err := net.Listen("tcp", s.srv.Addr)
+	if err != nil {
+		s.t.Fatalf("ListenFederationServer: net.Listen failed: %s", err)
+	}
+
 	go func() {
+		defer ln.Close()
 		defer wg.Done()
-		err := s.srv.ListenAndServeTLS(s.certPath, s.keyPath)
+		err := s.srv.ServeTLS(ln, s.certPath, s.keyPath)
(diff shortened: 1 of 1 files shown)


kegsay (commented):
What's the end-user visible change here? Is the stack trace better or something?

I'm happy with the change but I don't understand how this improves things, given we _should_ be failing pretty quickly anyway before (just waiting for a goroutine to spin up).

richvdh:
before:

```
--- FAIL: TestOutboundFederationIgnoresMissingEventWithBadJSONForRoomVersion6 (10.16s)
    federation_room_get_missing_events_test.go:33: Deploy times: 5.361170194s blueprints, 2.936517777s containers
    server.go:212: ListenFederationServer: ListenAndServeTLS failed: listen tcp :8448: bind: address already in use
    client.go:353: POST /_matrix/client/r0/join/#flibble:host.docker.internal => 500 Internal Server Error (416.598489ms)
    federation_room_get_missing_events_test.go:50: CSAPI.MustDoFunc response return non-2xx code: 500 Internal Server Error
```

after:

```
--- FAIL: TestOutboundFederationIgnoresMissingEventWithBadJSONForRoomVersion6 (9.89s)
    federation_room_get_missing_events_test.go:35: Deploy times: 4.741593323s blueprints, 3.096835435s containers
    server.go:211: ListenFederationServer: net.Listen failed: listen tcp :8448: bind: address already in use
```

In the first case, I always start out by thinking I've broken something so that it returns a 500 error.

kegsay:
Right, I see now. This looks fab, thank you!

kegsay (approved):

PR #145, opened 2021-07-01
Updates to tests for restricted rooms based on changes to MSC3083.

clokep:
Corresponds to matrix-org/synapse#10254:

* Updates the room version identifier.
* Additional tests for the authorised server changes in the MSC.


tests/msc3083_test.go
```diff
@@ -181,6 +184,130 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
...
+			"via": []string{"hs2"},
+		},
+	})
+
+	// Invite alice manually and accept it.
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+	charlie.InviteRoom(t, room, alice.UserID)
+	alice.JoinRoom(t, room, []string{"hs2"})
+
+	// Alice should not be able to issue invites.
```
babolivier, line 238: I assume that's because the default PLs in this room prevents her from doing so but could you add that info (or the real reason if I'm mistaken) to the comment? Since we're not _explicitly_ preventing her from inviting in the previous steps.

tests/msc3083_test.go
```diff
@@ -181,6 +184,130 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
...
+		client.WithJSONBody(t, body),
+	)
+	must.MatchResponse(t, res, match.HTTPResponse{
+		StatusCode: 403,
+	})
+
+	// Bob cannot join the room.
+	failJoinRoom(t, bob, room, "hs1")
+
+	// Join the space, joining the room should work, but will bounce through hs2.
```
babolivier, line 256: Sorry I'm not clear on whether it's just the room join or both the room and space joins that are expected to bounce through h2?

babolivier (approved):
A few more details needed in comments but otherwise lgtm

tests/msc3083_test.go
clokep: Yes, I'm just confirming that the user cannot issue invites as expected! I'll make it clearer.

tests/msc3083_test.go
clokep: Only the join to the room needs to go through hs2, I'll clarify!

clokep:
Thanks! I also updated the overall comment for the test since it was quite confusing.

PR #146, opened 2021-07-03
Add tests for clients sending `m.room.power_levels` events

aaronraimist:
Tests for https://github.com/matrix-org/synapse/pull/10232

### DIFF
--- a/internal/b/alice_and_bob.go
+++ b/internal/b/alice_and_bob.go
@@ -1,0 +1,21 @@
+package b
+
+// BlueprintAliceAndBob is a two user homeserver
+var BlueprintAliceAndBob = MustValidate(Blueprint{
+	Name: "alice_and_bob",
+	Homeservers: []Homeserver{
+		{
+			Name: "hs1",
+			Users: []User{
+				{
+					Localpart:   "@alice",
+					DisplayName: "Alice",
+				},
+				{
+					Localpart:   "@bob",
+					DisplayName: "Bob",
+				},
+			},
+		},
+	},
+})
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -112,12 +112,7 @@ func (c *CSAPI) InviteRoom(t *testing.T, roomID string, userID string) {
 // Returns the event ID of the sent event.
 func (c *CSAPI) SendEventSynced(t *testing.T, roomID string, e b.Event) string {
 	t.Helper()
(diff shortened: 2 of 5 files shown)


internal/b/alice_and_bob.go
```diff
@@ -0,0 +1,21 @@
+package b
+
+// BlueprintAliceAndBob is a two user homeserver
+var BlueprintAliceAndBob = MustValidate(Blueprint{
```
kegsay, line 4: Use `BlueprintOneToOneRoom` instead. https://github.com/matrix-org/complement/blob/master/internal/b/one_to_one_room.go#L4

internal/must/must.go
```diff
@@ -71,30 +71,30 @@ func MatchResponse(t *testing.T, res *http.Response, m match.HTTPResponse) []byt
...
 	}
 	if m.JSON != nil {
 		if !gjson.ValidBytes(body) {
-			t.Fatalf("MatchResponse response body is not valid JSON - %s", contextStr)
+			t.Errorf("MatchResponse response body is not valid JSON - %s", contextStr)
 		}
 		for _, jm := range m.JSON {
 			if err = jm(body); err != nil {
-				t.Fatalf("MatchResponse %s - %s", err, contextStr)
+				t.Errorf("MatchResponse %s - %s", err, contextStr)
```
kegsay, line 97: Please do not do this. It is intended for test assertions in this package to fail the test immediately and not to continue executing.

internal/client/client.go
```diff
@@ -127,6 +122,18 @@ func (c *CSAPI) SendEventSynced(t *testing.T, roomID string, e b.Event) string {
 	return eventID
 }
 
+// SendEventSynced sends `e` into the room.
+func (c *CSAPI) SendEvent(t *testing.T, roomID string, e b.Event) *http.Response {
```
kegsay, line 126: Please move this into a helper function inside `tests/client_power_levels_test.go` and not inside the Complement `CSAPI` client struct.

The reason for this is because `CSAPI` should be reserved for functions which are commonly used across many tests, and this function is primarily only useful to check error conditions when sending events. In addition, by having both forms sit alongside each other in `CSAPI`, it confuses test authors who may prefer to use this function over `SendEventSynced` which would end up being disasterous as `SendEventSynced` is the only form which guards against race conditions when processing new events.

PR #149, opened 2021-07-05
Fix race in MSC3083 TestRestrictedRoomsRemoteJoin

erikjohnston:
This was broken by matrix-org/synapse#10272

### DIFF
--- a/tests/msc3083_test.go
+++ b/tests/msc3083_test.go
@@ -95,7 +95,8 @@ func checkRestrictedRoom(t *testing.T, alice *client.CSAPI, bob *client.CSAPI, s
 	bob.LeaveRoom(t, space)
 
 	// Wait until Alice sees Bob leave the space. This ensures that Alice's HS
-	// has processed the leave before Bob tries rejoining.
+	// has processed the leave before Bob tries rejoining, so that it rejects his
+	// attempt to join the room.
 	alice.SyncUntilTimelineHas(t, space, func(ev gjson.Result) bool {
 		if ev.Get("type").Str != "m.room.member" || ev.Get("sender").Str != bob.UserID {
 			return false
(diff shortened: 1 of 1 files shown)


tests/msc3083_test.go
```diff
@@ -93,6 +93,15 @@ func checkRestrictedRoom(t *testing.T, alice *client.CSAPI, bob *client.CSAPI, s
 	// Leaving the room works and the user is unable to re-join.
 	bob.LeaveRoom(t, room)
 	bob.LeaveRoom(t, space)
+
+	alice.SyncUntilTimelineHas(t, space, func(ev gjson.Result) bool {
```
kegsay, line 97: I wonder if having `LeaveRoom` is a mistake, and we should always be doing the synced form..

erikjohnston: Maybe, though the subtlety here is that we need to wait for it to be processed on the *remote* server, rather than the local one.

richvdh, line 97:
```suggestion
	// wait until alice sees one of the leave events
	alice.SyncUntilTimelineHas(t, space, func(ev gjson.Result) bool {
```

richvdh (commented):
it kinda looks fine, but I don't quite follow why alice needs to see the leave for the test to continue?

erikjohnston:
> it kinda looks fine, but I don't quite follow why alice needs to see the leave for the test to continue?

I've added a comment, but basically Bob would leave the space and then try to join the original room again before the remote server saw him leave the space, so the remote server accepted the join when we were expecting it to fail.

tests/msc3083_test.go
```diff
@@ -93,6 +93,17 @@ func checkRestrictedRoom(t *testing.T, alice *client.CSAPI, bob *client.CSAPI, s
 	// Leaving the room works and the user is unable to re-join.
 	bob.LeaveRoom(t, room)
 	bob.LeaveRoom(t, space)
+
+	// Wait until Alice sees Bob leave the space. This ensures that Alice's HS
+	// has processed the leave before Bob tries rejoining.
```
richvdh, line 98:
```suggestion
	// Wait until Alice sees Bob leave the space. This ensures that Alice's HS
	// has processed the leave before Bob tries rejoining, so that it rejects his
	// attempt to join the room.
```

richvdh (approved):
lgtm

PR #151, opened 2021-07-08
Link to instructions on how to build SynapseWorker image

anoadragon453:
The comment was old (back when this file was in the Synapse repo) and not very helpful for actually building the image. The linked doc (which granted is hard to find on its own currently) should be helpful for those trying to use this.


kegsay (approved):
Better than nothing, thanks! Why does it have to be so complicated again?

PR #152, opened 2021-07-08
Fix 'Outbound federation will ignore a missing event with bad JSON for room version 6'

richvdh:
This test got blacklisted for the release of Synapse 1.37.1.

We can't assume that homeservers will process inbound federation events synchronously.

In this particular case we replace the check for errors in /send response with a check that if we send a new event that references the expected rejected event we'll see a request for that event come in.

c.f. matrix-org/synapse#10275 and matrix-org/sytest#1061

### DIFF
--- a/tests/federation_room_get_missing_events_test.go
+++ b/tests/federation_room_get_missing_events_test.go
@@ -1,5 +1,3 @@
-// +build !synapse_blacklist
-
 package tests
 
 import (
--- a/tests/federation_room_get_missing_events_test.go
+++ b/tests/federation_room_get_missing_events_test.go
@@ -168,9 +168,8 @@ func TestOutboundFederationIgnoresMissingEventWithBadJSONForRoomVersion6(t *test
 			},
 		})
 		defer waiter.Finish()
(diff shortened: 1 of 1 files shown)


tests/federation_room_get_missing_events_test.go
```diff
@@ -131,9 +141,46 @@ func TestOutboundFederationIgnoresMissingEventWithBadJSONForRoomVersion6(t *test
...
+		must.MatchRequest(t, req, match.HTTPRequest{
+			JSON: []match.JSON{
+				match.JSONKeyEqual("earliest_events", []interface{}{latestEvent.EventID()}),
+				match.JSONKeyEqual("latest_events", []interface{}{message3.EventID()}),
+			},
+		})
+		defer waiter.Finish()
+		// we don't really care what we return here.
+
+		// return the bad event, which should result in the transaction failing.
```
erikjohnston, line 173: We don't seem to be doing that?

richvdh: good spot...

erikjohnston (approved):

PR #156, opened 2021-07-11
 feat: add room-state tests

meenal06:


meenal06:
@kegsay please review the PR!

meenal06:
@kegsay ping

tests/apidoc_room_state_test.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		t.Fatalf("unable to read response body: %v", err)
+	}
+
+	roomID = gjson.GetBytes(body, "room_id").Str
+	roomAlias = gjson.GetBytes(body, "room_alias").Str
+	return roomID, roomAlias
+}
+
+func generateRoomAlias() (string, error) {
```
kegsay, line 99: Why are we doing this? It seems a little silly to randomly generate an alias every time. It hurts reproducibility of tests as each test run now has different values, and there is no improvement to clarity using `generateRoomAlias()` over a hardcoded value of say `"test_alias"`. Removing this also means you don't need to check for `error`.

Instead, just add an argument to `roomFixture` which is `roomAliasName string` and use that.

tests/apidoc_room_state_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+func TestRoomState(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	authedClient := deployment.Client(t, "hs1", "@alice:hs1")
+	t.Run("Parallel", func(t *testing.T) {
+		// sytest: GET /rooms/:room_id/state/m.room.member/:user_id fetches my membership
+		t.Run("GET /rooms/:room_id/state/m.room.member/:user_id fetches my membership", func(t *testing.T) {
+			t.Parallel()
+			roomID, _ := roomFixture(t, authedClient)
```
kegsay, line 26: I don't understand why we need `roomFixture` when `roomID := authedClient.CreateRoom(t, map[string]interface{}{})` is not only clearer but doesn't need `_` on the alias.

tests/apidoc_room_state_test.go
```diff
@@ -0,0 +1,111 @@
...
+					match.JSONKeyPresent("events_default"),
+					match.JSONKeyPresent("users"),
+					match.JSONKeyPresent("events"),
+				},
+			})
+		})
+	})
+}
+
+func roomFixture(t *testing.T, authedClient *client.CSAPI) (roomID string, roomAlias string) {
```
kegsay, line 75: I am not convinced we need this function at all, remove it please.

kegsay (changes requested):

tests/apidoc_room_state_test.go
meenal06: I didn't know the existence of CreateRoom until my last PR, apologies for this. I will clean up all unnecessary code

kegsay (approved):
LGTM! Thanks!

PR #158, opened 2021-07-13
Tweak ban test to make it clearer

kegsay:
This test still fails on Dendrite because it actually doesn't do this check.


richvdh (approved):
lgtm

PR #159, opened 2021-07-13
Build containers with the tag of their HS name

kegsay:
Else if they run at the same time in CI they cause problems and cancel
later test runs. cc @callahad


neilalexander (approved):

kegsay:
Looks like the root cause was actually https://github.com/matrix-org/complement/commit/d756e54b9cd7ea33f3fb1996a4578a20af45839a

PR #160, opened 2021-07-14
Handle failures in bestEffort mode without panicking

kegsay:
https://github.com/matrix-org/complement/pull/160/files?w=1

### DIFF
--- a/internal/instruction/runner.go
+++ b/internal/instruction/runner.go
@@ -138,33 +138,36 @@ func (r *Runner) runInstructionSet(hs b.Homeserver, hsURL string, instrs []instr
 				return err
 			}
 		}
-		if i < 100 || i%200 == 0 {
-			r.log("%s [%d/%d] %s => HTTP %s\n", contextStr, i, len(instrs), req.URL.String(), res.Status)
-		}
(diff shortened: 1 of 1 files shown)


internal/instruction/runner.go
```diff
@@ -138,33 +138,36 @@ func (r *Runner) runInstructionSet(hs b.Homeserver, hsURL string, instrs []instr
...
+				if err != nil {
+					return err
+				}
 			}
-		}
-		if instr.storeResponse != nil {
-			if !gjson.ValidBytes(body) {
-				err = isFatalErr(fmt.Errorf("%s : cannot storeResponse as response is not valid JSON. Body: %s", contextStr, string(body)))
+			if res.StatusCode < 200 || res.StatusCode >= 300 {
+				r.log("INSTRUCTION: %+v\n", instr)
```
neilalexander, line 154: Is this meant to be here still?

neilalexander (approved):

internal/instruction/runner.go
kegsay: Yes, if `COMPLEMENT_DEBUG` is set it's useful data to have.

PR #163, opened 2021-07-16
Add option to always print server logs

MadLittleMods:
Add option to always print server logs.

When working on https://github.com/matrix-org/complement/pull/68, I would just put a `true` in their manually but seems useful to have an actual option to use.


kegsay (approved):

MadLittleMods:
Thanks for taking a look @kegsay 🐈

PR #164, opened 2021-07-16
Document how to hook up Element to the resultant homeservers from Complement

MadLittleMods:
Document how to hook up Element to the resultant homeservers from running Complement tests.

Spawned from https://github.com/matrix-org/complement/pull/68


kegsay (approved):
You'll still hit the `go test` timeout, but removal of the `deployment.Destroy` line alone should be enough.

MadLittleMods:
Thanks for taking a look and merging @kegsay 🐚

PR #165, opened 2021-07-16
Document how to access the database for a homeserver after Complement runs

MadLittleMods:
Document how to access the database for a homeserver after Complement runs

Spawned from https://github.com/matrix-org/complement/pull/68

### DIFF
--- a/ONBOARDING.md
+++ b/ONBOARDING.md
@@ -188,6 +188,25 @@ For Goland:
  * Under "Run"->"Edit Configurations..."->"Templates"->"Go Test", add `COMPLEMENT_BASE_IMAGE=complement-dendrite:latest`
  * Then you can right-click on any test file or test case and "Run <test name>".
 
+
+
+### Access database for homeserver after Complement test runs
+
+
+For Synapse:
+
+```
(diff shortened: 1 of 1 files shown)


kegsay (commented):
Unsure about this one. It's too specific to a specific HS impl and DB impl.

MadLittleMods:
It's using the database that's defined in [`dockerfiles/synapse/homeserver.yaml`](https://github.com/matrix-org/complement/blob/3eaa98939578f56fa2cd1511556d110dfd8eaf1d/dockerfiles/synapse/homeserver.yaml#L31-L38).

I wanted to add one for Dendrite too but I haven't done it yet to be sure.


> **Synapse:**
> ...
>
> **Dendrite:**
> ...

kegsay:
I don't feel this is the right place for this information. This is too Synapse specific, and then SQLite specific. It should live in the Synapse repo.

MadLittleMods:
Moved this info to https://github.com/matrix-org/synapse/pull/10483 ⏩ 

Thanks for reviewing this @kegsay 🦢

PR #166, opened 2021-07-16
Always specify a preset

kegsay:
Synapse and Dendrite disagree what the join rules are if this is omitted.
Dendrite follows the spec, Synapse doesn't. Nontheless, we should be
specific in tests.

Also, actually fix the /v1 tests to unmarshal into the right data structure.
You can't use `map[string]interface{}` for v1 due to it sending back an
array of `[200, {}]` instead.

### DIFF
--- a/tests/federation_room_join_test.go
+++ b/tests/federation_room_join_test.go
@@ -326,7 +326,7 @@ func testValidationForSendMembershipEndpoint(t *testing.T, baseApiPath, expected
 			return
 		}
 
-		var res map[string]interface{}
+		var res interface{}
 		err := srv.SendFederationRequest(deployment, req, &res)
 		if err == nil {
 			t.Errorf("send request returned 200")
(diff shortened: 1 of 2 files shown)


tests/federation_room_join_test.go
```diff
@@ -290,7 +290,9 @@ func TestCannotSendNonLeaveViaSendLeaveV2(t *testing.T) {
 // and checks that they are all rejected.
 func testValidationForSendMembershipEndpoint(t *testing.T, baseApiPath, expectedMembership string, createRoomOpts map[string]interface{}) {
 	if createRoomOpts == nil {
-		createRoomOpts = make(map[string]interface{})
+		createRoomOpts = map[string]interface{}{
+			"preset": "public_chat",
+		}
```
richvdh, line 295: we'll need to change https://github.com/matrix-org/complement/blob/bdf545bf03b78b442e539ac78192001037e29aa9/tests/msc2403_test.go#L476 too, I think?

richvdh: because that test sets an explicit createRoomOpts, so it will otherwise create a room with no preset (and hence that will suffer the same problem as you're trying to solve here.)

kegsay: Oh I see, yes of course. Sorry I thought you meant putting the room version in the create opts.

PR #168, opened 2021-07-20
Adapt MSC2716 implementation to use chunk events

MadLittleMods:
Adapt MSC2716 implementation to use chunk events

See https://github.com/matrix-org/synapse/pull/10432

## Todo:

 - [x] Wait for https://github.com/matrix-org/synapse/pull/10432 to merge


kegsay (approved):
I assume this chunk event is made in Synapse and not by the client.

MadLittleMods:
> I assume this chunk event is made in Synapse and not by the client.

Correct, it's automatically added when using the `/batch_send` endpoint 👍

MadLittleMods:
Thanks for the review @kegsay 🐄

PR #171, opened 2021-07-22
Move CSAPI only tests to tests/csapi

### LINKED ISSUE #119: Parallelise tests by default

kegsay:
This allows them to run in parallel with federation tests when using `go test ./...`

Add a copy of `main_test.go` with the `PackageNamespace` set to `csapi`. This will be factored out in another PR.

Effectively fixes #119

On my machine before, a run of Complement on Dendrite would take ~360s. Now it takes 240s:
```
ok  	github.com/matrix-org/complement/tests/csapi	148.571s
ok  	github.com/matrix-org/complement/tests	239.968s
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,6 +40,8 @@ jobs:
         - /var/run/docker.sock:/var/run/docker.sock
 
     steps:
+      - name: Get number of CPU cores
+        uses: SimenB/github-actions-cpu-cores@v1
       - uses: actions/checkout@v2
       - run: docker build -t homeserver -f dockerfiles/${{ matrix.homeserver }}.Dockerfile dockerfiles/
       - run: go test -v -tags "${{ matrix.tags }}" ./tests/...
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,10 +40,9 @@ jobs:
         - /var/run/docker.sock:/var/run/docker.sock
 
     steps:
-      - name: Get number of CPU cores
-        uses: SimenB/github-actions-cpu-cores@v1
+      - run: lscpu
       - uses: actions/checkout@v2
       - run: docker build -t homeserver -f dockerfiles/${{ matrix.homeserver }}.Dockerfile dockerfiles/
-      - run: go test -v -tags "${{ matrix.tags }}" ./tests/...
+      - run: go test -p 2 -v -tags "${{ matrix.tags }}" ./tests/...
         env:
           COMPLEMENT_BASE_IMAGE: homeserver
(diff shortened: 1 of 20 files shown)


neilalexander (approved):
LGTM!

.github/workflows/ci.yaml
```diff
@@ -40,8 +40,9 @@ jobs:
         - /var/run/docker.sock:/var/run/docker.sock
 
     steps:
+      - run: lscpu
```
anoadragon453, line 43: Do you intend to keep this in?

tests/csapi/main_test.go
```diff
@@ -0,0 +1,129 @@
...
+var complementBuilder *docker.Builder
+
+// TestMain is the main entry point for Complement.
+//
+// It will clean up any old containers/images/networks from the previous run, then run the tests, then clean up
+// again. No blueprints are made at this point as they are lazily made on demand.
+func TestMain(m *testing.M) {
+	cfg := config.NewConfigFromEnvVars()
+	cfg.PackageNamespace = "csapi"
+	log.Printf("config: %+v", cfg)
```
anoadragon453, line 33: Also perhaps left over from debugging?

anoadragon453 (approved):
lgtm too!

tests/csapi/main_test.go
kegsay: No, keeping the config logs there is useful, we've always had this in `main_test.go`

.github/workflows/ci.yaml
kegsay: Removed.

PR #172, opened 2021-07-22
Test failover logic for MSC3083

clokep:
This tests what happens if a server is unable to authorize a room join (due to not being able to invite users or not being able to check the proper allowed rooms).

Corresponds to matrix-org/synapse#10447

~~Based on #145~~

### DIFF
--- a/tests/msc3083_test.go
+++ b/tests/msc3083_test.go
@@ -393,7 +393,7 @@ func TestRestrictedRoomsRemoteJoinFailOver(t *testing.T) {
 				return false
 			}
 			must.EqualStr(t, ev.Get("content").Get("membership").Str, "join", "Charlie failed to join the room")
-			must.EqualStr(t, ev.Get("content").Get("membership").Str, "join_authorised_via_users_server", alice.UserID)
+			must.EqualStr(t, ev.Get("content").Get("join_authorised_via_users_server").Str, alice.UserID, "Join authorised via incorrect server")
 
 			return true
 		},
@@ -448,7 +448,7 @@ func TestRestrictedRoomsRemoteJoinFailOver(t *testing.T) {
 				return false
 			}
 			must.EqualStr(t, ev.Get("content").Get("membership").Str, "join", "Charlie failed to join the room")
-			must.EqualStr(t, ev.Get("content").Get("membership").Str, "join_authorised_via_users_server", alice.UserID)
+			must.EqualStr(t, ev.Get("content").Get("join_authorised_via_users_server").Str, alice.UserID, "Join authorised via incorrect server")
 
 			return true
 		},
--- a/tests/msc3083_test.go
+++ b/tests/msc3083_test.go
@@ -17,7 +17,7 @@ import (
 	"github.com/tidwall/gjson"
 )
 
-func failJoinRoom(t *testing.T, c *client.CSAPI, roomIDOrAlias string, serverName string) {
+func failJoinRoom(t *testing.T, c *client.CSAPI, roomIDOrAlias string, serverName string, expectedErrorCode int) {
 	t.Helper()
 
 	// This is copied from Client.JoinRoom to test a join failure.
@@ -30,7 +30,7 @@ func failJoinRoom(t *testing.T, c *client.CSAPI, roomIDOrAlias string, serverNam
 		client.WithQueries(query),
 	)
 	must.MatchResponse(t, res, match.HTTPResponse{
-		StatusCode: 403,
+		StatusCode: expectedErrorCode,
 	})
 }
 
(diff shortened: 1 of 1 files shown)


tests/msc3083_test.go
```diff
@@ -181,6 +184,275 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
...
+			}
+			must.EqualStr(t, ev.Get("content").Get("membership").Str, "leave", "Charlie failed to leave the room")
+
+			return true
+		},
+	)
+
+	// Have bob leave and rejoin. This should still work even though hs2 isn't in
+	// the room anymore!
+	bob.LeaveRoom(t, room)
```
kegsay, line 306: This should probably be synced with `alice.SyncUntilTimelineHas` as well right?

tests/msc3083_test.go
```diff
@@ -181,6 +184,275 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
...
+	bob.JoinRoom(t, room, []string{"hs1"})
+}
+
+// A server will request a failover if asked to /make_join and it does not have
+// the appropriate authorisation to complete the request.
+//
+// Setup 3 homeservers:
+// * hs1 creates the space/room.
+// * hs2 joins the room
+// * hs3 attempts to join via hs2 (should fail) and hs3 (should work)
```
kegsay, line 316: This makes no sense. hs3 can join via itself??

tests/msc3083_test.go
```diff
@@ -181,6 +184,275 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
...
+						DisplayName: "Charlie",
+					},
+				},
+			},
+		},
+	})
+	defer deployment.Destroy(t)
+
+	// Setup the user, space, and restricted room.
+	alice, space, room := setupRestrictedRoom(t, deployment)
```
kegsay, line 353: Why don't you use `setupRestrictedRoom` in the previous test?

tests/msc3083_test.go
```diff
@@ -181,6 +184,275 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
...
+			},
+		},
+	})
+
+	// Create a second user on a different homeserver.
+	bob := deployment.Client(t, "hs2", "@bob:hs2")
+
+	// Bob joins the room and space.
+	bob.JoinRoom(t, space, []string{"hs1"})
+	bob.JoinRoom(t, room, []string{"hs1"})
```
kegsay, line 373: Shouldn't we wait for these to come down `alice` sync stream?

tests/msc3083_test.go
```diff
@@ -181,6 +184,275 @@ func TestRestrictedRoomsRemoteJoin(t *testing.T) {
...
+
+	// Charlie should join the space (which gives access to the room).
+	charlie := deployment.Client(t, "hs3", "@charlie:hs3")
+	charlie.JoinRoom(t, space, []string{"hs1"})
+
+	// hs2 doesn't have anyone to invite from, so the join fails.
+	failJoinRoom(t, charlie, room, "hs2", 502)
```
kegsay, line 380: Fairly sure there won't be any guarantee that hs2 would've seen the join from hs3 because the join was between hs3 and hs1. Therefore, doing this will be racey.

tests/msc3083_test.go
clokep: No it cannot, it doesn't have the permissions to issue a join request.

tests/msc3083_test.go
clokep: It doesn't matter if `alice` sees them.

tests/msc3083_test.go
clokep: Is there any reason this is just always synced? Seems like it would be an improvement.

tests/msc3083_test.go
clokep: The joiners need to be on the same server while the creator needs to be on a separate server. Unfortunately `BlueprintFederationTwoLocalOneRemote` has alice and bob on one server and charlie on another.

clokep:
I rebased this, sorry about that. Wanted to remove the commits from the previous PR and forgot there had been a review on this already. 🤦

tests/msc3083_test.go
anoadragon453: @clokep perhaps the comment should have `hs1` at the end of the last line instead of `hs3`?

anoadragon453 (approved):
This looks pretty sensible to me now from a test coverage perspective. Just one minor comment on a previous discussion.

tests/msc3083_test.go
clokep: Ah, yeah that's what @kegsay was getting at.

PR #176, opened 2021-07-23
Adapt MSC2716 implementation to `org.matrix.msc2716` experimental room version

MadLittleMods:
Adapt MSC2716 implementation to `org.matrix.msc2716` experimental room version

As introduced in https://github.com/matrix-org/synapse/pull/10245


kegsay (approved):
The duplication hurts. Consider making this a global:
```go
var createRoomOpts = map[string]interface{}{
    "preset": "public_chat",
    "name": "the hangout spot",
    "room_version": "org.matrix.msc2716"
}
```

MadLittleMods:
> The duplication hurts. Consider making this a global:

Updated 👍

PR #177, opened 2021-07-26
Attempt to pull the corresponding Synapse branch when running tests.

clokep:
This matches matrix-org/pipelines#158, but for GHA.

The code is cribbed from matrix-org/synapse#10160, and made somewhat homeserver-agnostic, but it seems that Dendrite actually pulls the code it needs as part of building the Docker image, so it is quite different.

This fixes the tests in #145

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -55,10 +55,8 @@ jobs:
           #
           # 1. First check if there's a similarly named branch (GITHUB_HEAD_REF
           #    for pull requests, otherwise GITHUB_REF).
-          # 2. Attempt to use the base branch, e.g. when merging into release-vX.Y
-          #    (GITHUB_BASE_REF for pull requests).
           # 3. Use the default homeserver branch.
-          for BRANCH_NAME in "$GITHUB_HEAD_REF" "$GITHUB_BASE_REF" "${GITHUB_REF#refs/heads/}" "${{ matrix.default_branch }}"; do
+          for BRANCH_NAME in "$GITHUB_HEAD_REF" "${GITHUB_REF#refs/heads/}" "${{ matrix.default_branch }}"; do
             # Skip empty branch names and merge commits.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -41,6 +43,33 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+
+      - name: "Checkout corresponding ${{ matrix.homeserver }} branch"
+        if: ${{ matrix.homeserver == 'Synapse' }}
+        shell: bash
+        run: |
+          mkdir -p homeserver
+          # Attempt to use the version of of the homeserver which best matches the
```
anoadragon453, line 52:
```suggestion
          # Attempt to use the version of the homeserver which best matches the
```

.github/workflows/ci.yaml
```diff
@@ -41,6 +43,33 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+
+      - name: "Checkout corresponding ${{ matrix.homeserver }} branch"
+        if: ${{ matrix.homeserver == 'Synapse' }}
```
anoadragon453, line 48: We should perhaps put in a comment here to explain why Dendrite doesn't need this.

.github/workflows/ci.yaml
```diff
@@ -41,6 +43,33 @@ jobs:
...
+        shell: bash
+        run: |
+          mkdir -p homeserver
+          # Attempt to use the version of of the homeserver which best matches the
+          # current build. Depending on whether this is a PR or release, etc. we
+          # need to use different fallbacks.
+          #
+          # 1. First check if there's a similarly named branch (GITHUB_HEAD_REF
+          #    for pull requests, otherwise GITHUB_REF).
+          # 3. Use the default homeserver branch.
```
anoadragon453, line 58:
```suggestion
          # 2. Use the default homeserver branch.
```

anoadragon453 (approved):
We'll need to generalise this further once other homeserver's (that aren't on github.com/matrix-org/...) get added to the remit, but this works well for now :slightly_smiling_face:

PR #179, opened 2021-07-28
Add more thorough E2E replace key checks

kegsay:
There was a bug in the Dendrite impl which Sytest did not pick up.
These are regression tests for that bug.

cc @ara4n


neilalexander (approved):
LGTM

PR #180, opened 2021-07-29
feat: implement all room state tests

meenal06:


internal/match/json.go
```diff
@@ -163,7 +163,13 @@ func JSONCheckOff(wantKey string, wantItems []interface{}, mapper func(gjson.Res
 // item calling `fn`. If `fn` returns an error, iterating stops and an error is returned.
 func JSONArrayEach(wantKey string, fn func(gjson.Result) error) JSON {
 	return func(body []byte) error {
-		res := gjson.GetBytes(body, wantKey)
+		var res gjson.Result
+		if wantKey == "" {
```
meenal06, line 167: @kegsay I had to made this change since the response itself was an array of objects & I didn't find any function to use over there

meenal06:
@kegsay I am getting this error in the last test case
```sh
=== CONT  TestRoomState/Parallel/POST_/createRoom_with_creation_content
    apidoc_room_state_test.go:310: HTTP/1.1 200 OK
        Content-Length: 62
        Access-Control-Allow-Headers: Origin, X-Requested-With, Content-Type, Accept, Authorization
        Access-Control-Allow-Methods: GET, POST, PUT, DELETE, OPTIONS
        Access-Control-Allow-Origin: *
        Content-Type: application/json
        Date: Thu, 29 Jul 2021 19:52:25 GMT

        {"creator":"@alice:hs1","m.federate":false,"room_version":"6"}
    apidoc_room_state_test.go:312: MatchResponse key 'm.federate' missing - http://localhost:55519/_matrix/client/r0/rooms/%21AHDjbXwJJbPUMmYu:hs1/state/m.room.create => {"creator":"@alice:hs1","m.federate":false,"room_version":"6"}
```

I am unable to see what am I doing wrong over here 😓

tests/csapi/apidoc_room_state_test.go
```diff
@@ -195,5 +195,129 @@ func TestRoomState(t *testing.T) {
...
+					"m.federate": false,
+				},
+			})
+
+			res := authedClient.MustDoFunc(t, "GET", []string{"_matrix", "client", "r0", "rooms", roomID, "state", "m.room.create"})
+
+			must.MatchResponse(t, res, match.HTTPResponse{
+				JSON: []match.JSON{
+					match.JSONKeyPresent("m.federate"),
+					match.JSONKeyEqual("m.federate", false),
```
kegsay, line 318: This won't work because you're specifying the `federate` key inside an `m` object e.g
```json
{
    "m": {
        "federate": true
    }
}
```

You need to escape the `.` for it to be treated as:
```json
{
    "m.federate": true
}
```

So use `"m\\.federate"`. This rule is set in the [gjson library](https://pkg.go.dev/github.com/tidwall/gjson#Get):

> The dot and wildcard character can be escaped with '\'.

internal/match/json.go
kegsay: Okay, that is a great change, thank you!

tests/csapi/apidoc_room_state_test.go
```diff
@@ -195,5 +195,129 @@ func TestRoomState(t *testing.T) {
 				},
 			})
 		})
+		// sytest: POST /rooms/:room_id/state/m.room.name sets name
+		t.Run("POST /rooms/:room_id/state/m.room.name sets name", func(t *testing.T) {
```
kegsay, line 199: It's actually a PUT here, not POST. Please correct the test name (though the sytest comment remains unchanged).

tests/csapi/apidoc_room_state_test.go
```diff
@@ -195,5 +195,129 @@ func TestRoomState(t *testing.T) {
...
+
+			must.MatchResponse(t, res, match.HTTPResponse{
+				JSON: []match.JSON{
+					match.JSONKeyPresent("topic"),
+					match.JSONKeyEqual("topic", "room_topic_test"),
+				},
+			})
+		})
+		// sytest: POST /rooms/:room_id/state/m.room.topic sets topic
+		t.Run("POST /rooms/:room_id/state/m.room.topic sets topic", func(t *testing.T) {
```
kegsay, line 242: Again this is PUT.

tests/csapi/apidoc_room_state_test.go
```diff
@@ -195,5 +195,129 @@ func TestRoomState(t *testing.T) {
...
+				"name":       "room_test",
+				"topic":      "room_topic_test",
+			})
+
+			wantKeys := map[string]bool{
+				"m.room.create":       true,
+				"m.room.join_rules":   true,
+				"m.room.name":         true,
+				"m.room.power_levels": true,
+			}
```
kegsay, line 281: Should probably include `m.room.topic` as well.

tests/csapi/apidoc_room_state_test.go
meenal06: Thanks!

tests/csapi/apidoc_room_state_test.go
meenal06: Got it 👍

meenal06:
@kegsay please review, I have addressed all the comments.

kegsay (approved):

PR #182, opened 2021-08-01
feat: add room-aliases test

meenal06:


meenal06:
@kegsay dendrite doesn't have the endpoint `roomID/aliases` hence the tests were failing on dendrite & not on synapse

kegsay (approved):
Excellent, thank you!

PR #187, opened 2021-08-06
Test the updated MSC2946 room hierarchy endpoint.

clokep:
This adds tests for matrix-org/synapse#10549.

I opted to not remove the old tests for now since both endpoints are going to exist in conjunction right now. We should remoe those in the future when removing support for the old endpoint.


richvdh (approved):
lgtm.

Only thing I'd suggest is adding comments next to the old `/rooms/{room}/spaces` tests to say that they are due for removal.

PR #191, opened 2021-08-10
Fix the build for the master branch for synapse.

### LINKED ISSUE #184: Synapse tests run against `master` branch instead of `develop` by default
``` 
=== RUN   TestRestrictedRoomsRemoteJoinLocalUser
    msc3083_test.go:191: Deploy times: 6.570039ms blueprints, 3.985522787s containers
    client.go:370: POST hs2/_matrix/client/r0/createRoom => 200 OK (226.771354ms)
    client.go:370: POST hs2/_matrix/client/r0/createRoom => 400 Bad Request (5.686334ms)
    msc3083_test.go:204: CSAPI.MustDo POST http://172.17.0.1:49305/_matrix/client/r0/createRoom returned HTTP 400
```

https://github.com/matrix-org/complement/runs/3180791920#step:7:1029

clokep:
When the `master` branch of complement is built we want to run against the *default homeserver branch*, not the `master` branch (which will most likely always exist).

Fixes #184

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -55,10 +55,14 @@ jobs:
           # current build. Depending on whether this is a PR or release, etc. we
           # need to use different fallbacks.
           #
-          # 1. First check if there's a similarly named branch (GITHUB_HEAD_REF
-          #    for pull requests, otherwise GITHUB_REF).
-          # 2. Use the default homeserver branch.
-          for BRANCH_NAME in "$GITHUB_HEAD_REF" "${GITHUB_REF#refs/heads/}" "${{ matrix.default_branch }}"; do
+          # Check if there's a similarly named branch (GITHUB_HEAD_REF for pull
+          # requests, otherwise GITHUB_REF), but if this is the default complement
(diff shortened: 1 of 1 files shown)


clokep:
I don't know for sure this will work until it's on `master` though. 😢

kegsay:
Isn't the default branch `HEAD`? I use this in Dendrite upgrade tests.

clokep:
From the GitHub docs:

> `GITHUB_HEAD_REF`: Only set for pull request events. The name of the head branch.

So yes, this doesn't seem correct. The other option is:

> `GITHUB_REF`: The branch or tag ref that triggered the workflow. For example, `refs/heads/feature-branch-1`. If neither a branch or tag is available for the event type, the variable will not exist.

I suspect I need to use that one.

I'm not sure what you mean by the default branch being `HEAD` though. `HEAD` is just a pointer to the commit you have checked out.

clokep:
I think the updated version I have now will work correctly.

richvdh (approved):
seems plausible, if a bit fiddly.

PR #175, opened 2021-07-23
MSC2716 federation changes

MadLittleMods:
[MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) federation changes

 - Unskip the rest of the federation tests
 - Refactor `batchSendHistoricalMessages` to be able to compose state and message events
 - Add test to ensure it is possible to batch send in a private room
 - Add test to ensure we only pass back `state_events` that we passed in via `?state_events_at_start`, https://github.com/matrix-org/synapse/pull/10552

### Dev notes

```
COMPLEMENT_DIR=../complement ./scripts-dev/complement.sh TestBackfillingHistory
```

---

```go
messagesResAsdf := alice.MustDoFunc(t, "GET", []string{"_matrix", "client", "r0", "rooms", roomID, "messages"}, client.WithContentType("application/json"), client.WithQueries(url.Values{
	"dir":   []string{"b"},
	"limit": []string{"100"},
}))
messagesResAsdfBody := client.ParseJSON(t, messagesResAsdf)
eventDebugStringsFromResponseAsdf := getRelevantEventDebugStringsFromMessagesResponse(t, messagesResAsdfBody)
logrus.WithFields(logrus.Fields{
	"eventDebugStringsFromResponseAsdf": eventDebugStringsFromResponseAsdf,
}).Error("see messages")
```

### Todo

 - [x] Should work with private room where the `state_events_before` includes invitations and join events
 - [ ] ~~Redacting an insertion, chunk, marker so remove it from our new tracking tables~~
    - https://github.com/matrix-org/synapse/issues/10521
    - https://github.com/matrix-org/synapse/pull/10538
 - [ ] Two insertion events with the same `next_chunk_id` will reject. (`(room_id, next_chunk_id)` should be unique))
    - Marked as TODO tests for another iteration
 - [ ] What happens when two chunks try to point at the same insertion event?
    - Marked as TODO tests for another iteration

PR #192, opened 2021-08-11
Add tests for MSC2716 being backwards compatible with existing room versions

MadLittleMods:
Add tests for [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) being backwards compatible with existing room versions

Synapse changes added in https://github.com/matrix-org/synapse/pull/10566


kegsay (approved):
On the Complement side, dunno about the synapse side.

MadLittleMods:
Thanks for the review @kegsay! This PR depends on https://github.com/matrix-org/complement/pull/175 so I'll have to wait to merge it until that one lands ⏩

PR #196, opened 2021-08-21
feat: add content tests for download/upload

meenal06:

### DIFF
--- a/tests/csapi/apidoc_content_test.go
+++ b/tests/csapi/apidoc_content_test.go
@@ -1,11 +1,12 @@
 package csapi_tests
 
 import (
-	"github.com/tidwall/gjson"
 	"io/ioutil"
 	"net/url"
 	"strings"
 	"testing"
+
+	"github.com/tidwall/gjson"
 
 	"github.com/matrix-org/complement/internal/b"
 	"github.com/matrix-org/complement/internal/client"
--- a/tests/csapi/apidoc_content_test.go
+++ b/tests/csapi/apidoc_content_test.go
@@ -1,0 +1,62 @@
+package csapi_tests
+
+import (
+	"github.com/tidwall/gjson"
+	"io/ioutil"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/client"
+)
+
+var contentID *url.URL
+var contentText = "Here is the content I am uploading"
+
+func TestContent(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	authedClient := deployment.Client(t, "hs1", "@alice:hs1")
+	unauthedClient := deployment.Client(t, "hs1", "")
+	// sytest: POST /media/r0/upload can create an upload
+	t.Run("POST /media/r0/upload can create an upload", func(t *testing.T) {
+
+		gotUrl := ""
+		queryParams := url.Values{}
+		queryParams.Set("access_token", authedClient.AccessToken)
+
+		contentType := client.WithContentType("text/plain")
+		reqBody := client.WithRawBody([]byte(contentText))
+		res := unauthedClient.MustDoFunc(t, "POST", []string{"_matrix", "media", "r0", "upload"}, reqBody, contentType, client.WithQueries(queryParams))
+
+		body, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			t.Fatalf("unable to read response body: %v", err)
+		}
+
+		gotUrl = gjson.GetBytes(body, "content_uri").Str
+		if gotUrl == "" {
+			t.Fatal("expected URL to hae propogated the value")
+		}
+
+		contentID, err = url.Parse(gotUrl)
+		if err != nil {
+			t.Fatal("invalid URL returned in response")
+		}
+	})
+	// sytest: GET /media/r0/download can fetch the value again
+	t.Run("GET /media/r0/download can fetch the value again", func(t *testing.T) {
+
+		res := unauthedClient.MustDoFunc(t, "GET", []string{"_matrix", "media", "r0", "download", contentID.Host, strings.Trim(contentID.Path, "/")})
+
+		responseData, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			t.Fatalf("unable to read response body: %v", err)
+		}
+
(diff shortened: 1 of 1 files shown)


meenal06:
@kegsay please review the following PR

tests/csapi/apidoc_content_test.go
```diff
@@ -0,0 +1,63 @@
...
+var contentID *url.URL
+var contentText = "Here is the content I am uploading"
+
+func TestContent(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	authedClient := deployment.Client(t, "hs1", "@alice:hs1")
+	unauthedClient := deployment.Client(t, "hs1", "")
+	// sytest: POST /media/r0/upload can create an upload
+	t.Run("POST /media/r0/upload can create an upload", func(t *testing.T) {
```
kegsay, line 24: Please add these tests to https://github.com/matrix-org/complement/blob/master/tests/media_nofilename_test.go#L40 instead

tests/csapi/apidoc_content_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+		gotUrl := ""
+		queryParams := url.Values{}
+		queryParams.Set("access_token", authedClient.AccessToken)
```
kegsay, line 28: This shouldn't be necessary. Just use an authed client as always. The perl test didn't because the framework wouldn't let you if you weren't sending JSON.

tests/csapi/apidoc_content_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+		contentType := client.WithContentType("text/plain")
+		reqBody := client.WithRawBody([]byte(contentText))
+		res := unauthedClient.MustDoFunc(t, "POST", []string{"_matrix", "media", "r0", "upload"}, reqBody, contentType, client.WithQueries(queryParams))
```
kegsay, line 32: In-line `contentType` and `reqBody` as it makes it a bit easier to read.

tests/csapi/apidoc_content_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+		res := unauthedClient.MustDoFunc(t, "GET", []string{"_matrix", "media", "r0", "download", contentID.Host, strings.Trim(contentID.Path, "/")})
+
+		responseData, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			t.Fatalf("unable to read response body: %v", err)
+		}
+
+		if string(responseData) != contentText {
+			t.Fatal("failed to fetch valid content text")
```
kegsay, line 60: Please log the actual values as is typical for Go tests: `t.Fatalf("got content '%s' want '%s'", string(responseData), contentText)`

tests/csapi/apidoc_content_test.go
```diff
@@ -0,0 +1,63 @@
...
+		if err != nil {
+			t.Fatal("invalid URL returned in response")
+		}
+	})
+	// sytest: GET /media/r0/download can fetch the value again
+	t.Run("GET /media/r0/download can fetch the value again", func(t *testing.T) {
+
+		res := unauthedClient.MustDoFunc(t, "GET", []string{"_matrix", "media", "r0", "download", contentID.Host, strings.Trim(contentID.Path, "/")})
+
+		responseData, err := ioutil.ReadAll(res.Body)
```
kegsay, line 54: You need to check for the headers as well.
```go
if res.Header.Get("Content-Type") != "text/plain" {
    t.Errorf(...)
}
```

tests/csapi/apidoc_content_test.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		gotUrl = gjson.GetBytes(body, "content_uri").Str
+		if gotUrl == "" {
+			t.Fatal("expected URL to hae propogated the value")
+		}
+
+		contentID, err = url.Parse(gotUrl)
+		if err != nil {
+			t.Fatal("invalid URL returned in response")
```
kegsay, line 46: Please log the error and the `gotUrl`.

It is never sufficient to just log a simple error message in tests like this. Think what the experience is like for someone whose code is failing because of this test. What information would they find useful? They certainly will want to know _why_ the test is failing (the `err` in this case). Giving them some more context to try to help them debug this (the `gotUrl` in this case) is also helpful.

Think about it like when you write tests in Complement. Sometimes the server encounters an error, and Complement gives detailed error messages why. It doesn't just say "request failed" because it just isn't helpful.

tests/csapi/apidoc_content_test.go
```diff
@@ -0,0 +1,63 @@
...
+		contentType := client.WithContentType("text/plain")
+		reqBody := client.WithRawBody([]byte(contentText))
+		res := unauthedClient.MustDoFunc(t, "POST", []string{"_matrix", "media", "r0", "upload"}, reqBody, contentType, client.WithQueries(queryParams))
+
+		body, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			t.Fatalf("unable to read response body: %v", err)
+		}
+
+		gotUrl = gjson.GetBytes(body, "content_uri").Str
```
kegsay, line 39: Please stick to Go naming conventions: `gotURL` not `gotUrl`.

tests/csapi/apidoc_content_test.go
```diff
@@ -0,0 +1,63 @@
...
+		if gotUrl == "" {
+			t.Fatal("expected URL to hae propogated the value")
```
kegsay, line 41: This isn't a helpful error message because it doesn't explain where this URL came from (the error) and give context to other fields that are in the JSON. Imagine if `body` had the shape:
```json
{
    "content_url": "mxc://foo/bar"
}
```
This looks correct but it isn't. Running this test would fail with:
```
expected URL to hae propogated the value
```
What does this mean from the perspective of the person testing their code? Not a lot. An improvement would be:
```
expected 'content_uri' in response body
```
This lets the developer know that a key is missing. We can be even more helpful though by providing context: what keys were in the response? If we don't, the developer may spend 20-30 minutes thinking that there is no URI at all in the response which isn't quite true. The best message would be:
```
expected 'content_uri' in response body, got: { "content_url": "mxc://foo/bar" }
```
This _immediately_ tells the developer that they made a typo: `url` instead of `uri`. Good error messages _saves developer time_, please treat them with just as much importance as the actual mechanics of the test.

kegsay (changes requested):

PR #198, opened 2021-08-23
Unblacklist user_query_keys_test for synapse

DMRobertson:
After my changes in matrix-org/synapse#10593

### DIFF
--- a/tests/csapi/user_query_keys_test.go
+++ b/tests/csapi/user_query_keys_test.go
@@ -1,7 +1,3 @@
-// +build !synapse_blacklist
-
-// Rationale for being included in Synapse's blacklist: https://github.com/matrix-org/synapse/issues/10354
-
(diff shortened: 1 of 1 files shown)


clokep (approved):
Seems reasonable since the tests pass. 😄

DMRobertson:
Hmmm. I can't see TestKeysQueryWithDeviceIDAsObjectFails in the CI run anywhere---does that mean this test isn't getting run today (even for dendrite?)

DMRobertson:
> Hmmm. I can't see TestKeysQueryWithDeviceIDAsObjectFails in the CI run anywhere---does that mean this test isn't getting run today (even for dendrite?)

Can see it in the raw logs---just couldn't search for it with Ctrl-F while viewing the output from github actions. Assuming it's a PICNIC on my part.

PR #202, opened 2021-08-30
/make_join responses should not include signatures on events.

clokep:
This corresponds to matrix-org/synapse#10404 which shipped with Synapse 1.39.0.

I confirmed that Synapse 1.38.0 fails this test.

I'm not 100% sure we want to assert this via complement, but it seems like a bit of a tricky thing that's "wrong" to do.

### DIFF
--- a/tests/federation_room_join_test.go
+++ b/tests/federation_room_join_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/matrix-org/gomatrix"
 	"github.com/matrix-org/gomatrixserverlib"
 
+	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
 
 	"github.com/matrix-org/complement/internal/b"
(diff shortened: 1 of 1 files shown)


clokep:
> I kinda wonder if we should make it the joining server's problem, in that if it wants to modify the event, it needs to remove any existing signatures so that there are no invalid sigs.

That could work too. I guess we can't error and try another server since that wouldn't be backwards compatible with older Synapses.

clokep:
Note that I originally wanted to write this since it was non-obvious that Synapse was even sending these signatures so my hope was that we could stop other homeservers from running into the same implementation issue. 🤷 I do agree that I don't like asserting things *aren't* there though. I can close this then.

Do you have thoughts on a next step? Is it clarifying the spec? An MSC for a joining server rejecting (or stripping) a response that includes signatures?

clokep:
Based on conversations with Synapse, Dendrite, and Conduit devs it doesn't seem we want to be asserting this.

PR #203, opened 2021-09-07
Add tests to verify ?chunk_id actually corresponds to an insertion event that exists (MSC2716)

MadLittleMods:
Add tests to verify ?chunk_id actually corresponds to an insertion event that exists

Part of [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716)

Add tests for:

 - https://github.com/matrix-org/synapse/pull/10776

PR #204, opened 2021-09-07
Update tests to accommodate `/batch_send` returning separate meta event fields (MSC2716)

MadLittleMods:
Update tests to accommodate `/batch_send` returning separate meta event fields.

Corresponding Synapse changes in https://github.com/matrix-org/synapse/pull/10777

See https://github.com/matrix-org/matrix-doc/pull/2716#discussion_r684765034

Part of [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716)


tests/msc2716_test.go
```diff
@@ -112,43 +112,53 @@ func TestBackfillingHistory(t *testing.T) {
 			eventIDsAfter := createMessagesInRoom(t, alice, roomID, 2)
 
 			// Insert the most recent chunk of backfilled history
-			insertTime1 := timeAfterEventBefore.Add(timeBetweenMessages * 3)
+			insertTime0 := timeAfterEventBefore.Add(timeBetweenMessages * 3)
```
MadLittleMods, line 115: Updating from `insertTime1` -> `insertTime0` to match the chunk explanation below

tests/msc2716_test.go
```diff
@@ -1147,14 +1154,38 @@ func batchSendHistoricalMessages(
...
 }
+
+func getInsertionEventIDFromBatchSendResponseBody(t *testing.T, body []byte) (insertionEventID string) {
+	insertionEventID = client.GetJSONFieldStr(t, body, "insertion_event_id")
+
+	return insertionEventID
+}
+
+func getChunkEventIdFromBatchSendResponseBody(t *testing.T, body []byte) (chunkEventId string) {
+	chunkEventId = client.GetJSONFieldStr(t, body, "chunk_event_id")
```
kegsay, line 1182: I don't understand the rationale for splitting this out to a different function. `getChunkEventIdFromBatchSendResponseBody` is a convolutedly named proxy function for something which is clearer (and in fact shorter!): `client.GetJSONFieldStr(t, body, "chunk_event_id")`

This is just extra code which makes it less clear to see what operations are actually happening in the tests. Am I missing something here?

tests/msc2716_test.go
```diff
@@ -610,12 +620,11 @@ func TestBackfillingHistory(t *testing.T) {
...
-			historicalEventIDs := getEventsFromBatchSendResponseBody(t, batchSendResBody)
-			baseInsertionEventID := historicalEventIDs[len(historicalEventIDs)-1]
+			historicalEventIDs := getEventIDsFromBatchSendResponseBody(t, batchSendResBody)
+			baseInsertionEventID := getBaseInsertionEventIDFromBatchSendResponseBody(t, batchSendResBody)
 
-			// [1 insertion event + 2 historical events + 1 chunk event + 1 insertion event]
-			if len(historicalEventIDs) != 5 {
-				t.Fatalf("Expected eventID list should be length 5 but saw %d: %s", len(historicalEventIDs), historicalEventIDs)
+			if len(historicalEventIDs) != 2 {
+				t.Fatalf("Expected eventID list should be length 2 but saw %d: %s", len(historicalEventIDs), historicalEventIDs)
```
kegsay, line 627: Please please please document why you expect this to be 2. I don't see you changing the number of injected events, so why shouldn't this be 5? I'm guessing the event IDs returned in the batch endpoint are the earliest or something? But this is really really not clear. It's hard enough to follow all the event IDs and markers and timestamps without randomly changing a 5 into a 2 without any reason why.

MadLittleMods: What's changed here is we have gone from including the 3 insertion and chunk events in the `event_ids` response, to separating them out to their own fields.

So now `event_ids` in the response only includes the 2 historical events that you sent in your request which is much more obvious for a downstream user (spawned from https://github.com/matrix-org/matrix-doc/pull/2716#discussion_r684765034)

I've updated the assertion message to make it clear `Expected 2 event_ids in the response that correspond to the 2 events we sent in the request but saw XX: ...`

tests/msc2716_test.go
MadLittleMods: I'm not sure anymore. Probably just morphed from something more involved and now useless. Removed ⏩

MadLittleMods:
@kegsay Thanks for the initial round of review 🐷 Good shout on the useless getters!

Merging as the chain of PRs is getting a little too unwieldy. I'm willing to change anything you comment about in a follow-up though ⏩

 - https://github.com/matrix-org/complement/pull/203
 - https://github.com/matrix-org/complement/pull/204
 - https://github.com/matrix-org/complement/pull/205
 - https://github.com/matrix-org/complement/pull/206

PR #206, opened 2021-09-21
Add tests to verify that we can resolve member avatar/displayname state for historical messages (MSC2716)

MadLittleMods:
Add tests to verify that we can resolve member avatar/displayname state for historical messages ([MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716))

Part of testing the batch importing script for Gitter https://gitlab.com/gitterHQ/webapp/-/merge_requests/2229#note_654645018

## Dev notes

```
.map((event) => ({
	type: event.type,
	sender: event.sender,
	depth: event.depth,
	stream_ordering: event.stream_ordering,
	body: event.content.body,
	event_id: event.event_id,
}))
```


tests/msc2716_test.go
```diff
@@ -457,6 +458,82 @@ func TestBackfillingHistory(t *testing.T) {
...
+			stateJsonResult1 := gjson.GetBytes(contextResBody1, "state")
+			stateFromContext1 := stateJsonResult1.Array()
+
+			logrus.WithFields(logrus.Fields{
+				"stateFromContext": stateFromContext1,
+				"len":              len(stateFromContext1),
+			}).Error("gaweegegegwegawagewg 1")
+
+			if len(stateFromContext1) == 0 {
+				t.Fatalf("Expected some state events in the context response for historical event in second batch but saw %d: %s", len(stateFromContext1), stateFromContext1)
```
MadLittleMods, line 533: This currently doesn't work because the historical messages don't have proper `state_groups`.

> Element is having trouble pulling in the avatar/displayname because the floating outlier member event state (from `state_events_at_start`) is not resolved into the current room state as expected but also the historical events don't have proper `state_groups`.
> 
> Instead of worrying about `state_groups`, we could just pull from profile as a fallback which is also noted in https://matrix.org/docs/spec/client_server/latest#lazy-loading-room-members ("Clients which do not support tab-completion may instead pull in profiles for arbitrary users (e.g. read receipts, typing notifications) on demand by querying the room state or /profile.")
> 
> *-- https://gitlab.com/gitterHQ/webapp/-/merge_requests/2229#note_683532091*

tests/msc2716_test.go
```diff
@@ -533,18 +532,7 @@ func TestImportHistoricalMessages(t *testing.T) {
...
-			}))
-
-			must.MatchResponse(t, messagesRes, match.HTTPResponse{
-				JSON: []match.JSON{
-					match.JSONCheckOffAllowUnwanted("chunk", makeInterfaceSlice(historicalEventIDs), func(r gjson.Result) interface{} {
-						return r.Get("event_id").Str
-					}, nil),
-				},
-			})
+			validateBatchSendRes(t, remoteCharlie, roomID, batchSendRes, false)
```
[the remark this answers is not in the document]
MadLittleMods: Probably best to revert for the federated tests and just add a comment about potentially using `validateBatchSendRes` in the future so we can merge this now.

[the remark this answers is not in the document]
MadLittleMods: Reverted for federated tests and added fixme comments

tests/msc2716_test.go
```diff
@@ -1211,3 +1254,159 @@ func batchSendHistoricalMessages(
 
 	return res
 }
+
+// Verify that the batch of historical messages looks correct and in order in
+// the message scrollback. We can also optionally check that the historical
```
kegsay, line 1259: "in order in the message scrollback" - please specify if it's old -> new or new -> old.

tests/msc2716_test.go
```diff
@@ -1211,3 +1254,159 @@ func batchSendHistoricalMessages(
...
+				// Now that we found it, pop the message off the expected list
+				workingExpectedEventIDOrder = workingExpectedEventIDOrder[1:]
+			}
+
+			// Found all of the expected events, stop iterating
+			if len(workingExpectedEventIDOrder) == 0 {
+				return false
+			}
+
+			return err == nil
```
kegsay, line 1401: Control flow feels weird here, why not just assign to `err` and `return false` up at `:1389`?

tests/msc2716_test.go
```diff
@@ -1211,3 +1254,159 @@ func batchSendHistoricalMessages(
...
+				historicalEventFilter,
+			),
+		},
+	})
+}
+
+// Looks through a list of events to find the sliding window of expected event
+// ID's somewhere in the list in order. The expected list can start anywhere in
+// the overall list.
+func matcherJSONEventIDArrayInOrder(wantKey string, expectedEventIDOrder []string, eventFilter func(gjson.Result) bool) match.JSON {
```
kegsay, line 1345: Please document your inputs more. `wantKey` feels like it should be mandatory (and the term "want" is usually paired with "got" for testing but that isn't the case here) but it's in fact an optional key which is where to start looking for events from. I would rename this to `jsonArrayKey`.

tests/msc2716_test.go
kegsay: Also, I would actually write `/messages` rather than "message scrollback". The former is much clearer that this function is going to be doing an HTTP hit to verify the events.

tests/msc2716_test.go
```diff
@@ -1211,3 +1254,159 @@ func batchSendHistoricalMessages(
...
+	var expectedEventIDOrder []string
+	if baseInsertionEventID != "" {
+		expectedEventIDOrder = append(expectedEventIDOrder, baseInsertionEventID)
+	}
+	expectedEventIDOrder = append(expectedEventIDOrder, batchEventID)
+	expectedEventIDOrder = append(expectedEventIDOrder, reversed(historicalEventIDs)...)
+	expectedEventIDOrder = append(expectedEventIDOrder, insertionEventID)
+
+	if validateState {
+		// Get the pagination token for the end of the historical batch itself
```
kegsay, line 1287: I don't like the term "end". It's so hard to know if that means old or new, especially since this entire MSC revolves around making old messages at physically later points in time. It's a tricky one to get terminology right but it would help readability on these tests:
 - start / end : ambiguous time, but clear on index positions (though it's interesting that we then access `[0]` in `:1288` :(
 - earliest / latest : ambiguous time (earliest at send time or event time?), ambiguous on index positions (depends on sort order)

You know this problem space better than I do, perhaps you can come up with better names.

kegsay (approved):
Broadly speaking LGTM. Most of my complaints are to do with naming as I constantly get confused as to the timings / orderings of events.

tests/msc2716_test.go
```diff
@@ -779,6 +812,12 @@ func TestImportHistoricalMessages(t *testing.T) {
 			// Send the marker event
 			sendMarkerAndEnsureBackfilled(t, as, remoteCharlie, roomID, baseInsertionEventID)
 
+			// FIXME: In the future, we should probably replace the following logic
+			// with `validateBatchSendRes` to re-use and have some more robust
+			// assertion logic here. We're currently not using it because the message
+			// order isn't quite perfect when a remote federated homserver gets
```
MadLittleMods, line 818: `homserver` typo

tests/msc2716_test.go
```diff
@@ -527,11 +547,6 @@ func TestImportHistoricalMessages(t *testing.T) {
 			})
```
MadLittleMods, line 547: TODO: Use `validateBatchSendRes` here (https://github.com/matrix-org/complement/pull/219 was merged in)

tests/msc2716_test.go
MadLittleMods: `newest-in-time` / `oldest-in-time` is probably the best for describing which end of the list we're at.

For describing the order, `chronological` / `reverse chronological` also works but it's a bit wordy and can kinda get confusing to reason about.

tests/msc2716_test.go
MadLittleMods: That works, updated 👍

For reference, it was that way because it was mimicking [`JSONArrayEach`](https://github.com/matrix-org/complement/blob/5535399363421bc45b762dce3d4d8432607ce583/internal/match/json.go#L182) and I previously had this middle logic refactored into an internal callback at some point.

MadLittleMods:
Thanks for the review @kegsay! Added some more comments to make this more clear. I'm striving to make this digestible for future onlookers so your pokes there are very welcome!

PR #208, opened 2021-09-28
Do a clearer join -> join transition for restricted rooms tests.

clokep:
Corresponds to matrix-org/synapse#10933.


richvdh (approved):
lgtm

PR #209, opened 2021-10-01
More test coverage to ensure we process MSC2716 events in existing room versions

MadLittleMods:
More test coverage to ensure we process MSC2716 events in existing room versions

Synapse changes in https://github.com/matrix-org/synapse/pull/10962

### DIFF
--- a/tests/msc2716_test.go
+++ b/tests/msc2716_test.go
@@ -780,6 +780,9 @@ func TestBackfillingHistory(t *testing.T) {
 				eventIdBefore := eventIDsBefore[0]
 				timeAfterEventBefore := time.Now()
 
+				// Create eventIDsAfter to avoid the "No forward extremities left!" 500 error from Synapse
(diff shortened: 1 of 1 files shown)


tests/msc2716_test.go
```diff
@@ -782,6 +782,9 @@ func TestImportHistoricalMessages(t *testing.T) {
 				eventIdBefore := eventIDsBefore[0]
 				timeAfterEventBefore := time.Now()
 
+				// Create eventIDsAfter to avoid the "No forward extremities left!" 500 error from Synapse
+				createMessagesInRoom(t, alice, roomID, 2)
```
kegsay, line 786: Poor comment wording. You reference `eventIDsAfter` like it's a variable but it isn't used here?

Do you really mean:

> Create *events after* to avoid the "No forward extremities left!" 500 error from Synapse

?

What does this have to do with room versions anyway? This just seems to be a logic error?

PR #210, opened 2021-10-01
Test auth of events with rejected auth_events

richvdh:
We should make sure that events that refer to rejected events in their
auth_events are themselves rejected.

See https://github.com/matrix-org/synapse/issues/9595, and the fixes at https://github.com/matrix-org/synapse/pull/10956.

### DIFF
--- a/tests/federation_room_event_auth_test.go
+++ b/tests/federation_room_event_auth_test.go
@@ -1,3 +1,6 @@
+// These tests currently fail on Dendrite, due to Dendrite bugs.
+// +build !dendrite_blacklist
+
 package tests
 
 import (
@@ -19,6 +22,8 @@ import (
 func TestInboundFederationRejectsEventsWithRejectedAuthEvents(t *testing.T) {
 	/* These tests check that events which refer to rejected events in auth_events
 	 * are themselves rejected.
+	 *
+	 * They are regression tests for https://github.com/matrix-org/synapse/issues/9595.
 	 *
 	 * In order to inject an outlier, we include it as an extra auth_event in a
 	 * regular event. Doing so means that the regular event should itself be
--- a/tests/federation_room_event_auth_test.go
+++ b/tests/federation_room_event_auth_test.go
@@ -41,6 +41,7 @@ func TestInboundFederationRejectsEventsWithRejectedAuthEvents(t *testing.T) {
 	 *   S
 	 *
 	 * Where:
+	 *    |  represents a "prev_event" link (older events are at the top)
 	 *    .... represents an "auth_event" link
 	 *    C is the room creation series
 	 *    R is a rejected event
--- a/tests/federation_room_event_auth_test.go
+++ b/tests/federation_room_event_auth_test.go
@@ -1,0 +1,200 @@
+package tests
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/gorilla/mux"
+	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/tidwall/gjson"
+
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/federation"
+	"github.com/matrix-org/complement/internal/must"
+)
+
+func TestInboundFederationRejectsEventsWithRejectedAuthEvents(t *testing.T) {
+	/* These tests check that events which refer to rejected events in auth_events
+	 * are themselves rejected.
+	 *
+	 * In order to inject an outlier, we include it as an extra auth_event in a
+	 * regular event. Doing so means that the regular event should itself be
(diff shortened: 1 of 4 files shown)


richvdh:
looks like dendrite has this bug too. I'm not sure how best to handle that: should I blacklist the test for dendrite somehow for now?

tests/federation_room_event_auth_test.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/federation"
+	"github.com/matrix-org/complement/internal/must"
+)
+
+func TestInboundFederationRejectsEventsWithRejectedAuthEvents(t *testing.T) {
+	/* These tests check that events which refer to rejected events in auth_events
+	 * are themselves rejected.
+	 *
+	 * In order to inject an outlier, we include it as an extra auth_event in a
```
DMRobertson, line 23: I've just tried to remind myself what an outlier is by looking in the spec. No matches that I could see. I'm guessing this isn't Synapse-specific but just something that arises when synchronising DAGs. Is it worth sticking a definition in the spec somewhere? A glossary even?

DMRobertson: There is https://matrix-org.github.io/synapse/latest/development/room-dag-concepts.html. Maybe these are synapse-specific terms/implementation choices after all?

tests/federation_room_event_auth_test.go
```diff
@@ -0,0 +1,206 @@
...
+	 * everywhere. This acts as a sentinel so that we can be sure that the
+	 * events have all been correctly propagated.
+	 *
+	 * The DAG ends up looking like this:
+	 *
+	 *       C
+	 *     / | \
+	 *    /  R  \
+	 *   |   ^   \
+	 *   |    .. O
```
DMRobertson, line 37:
```suggestion
	 *   |   ... O
```

tests/federation_room_event_auth_test.go
```diff
@@ -0,0 +1,206 @@
...
+	 *   |       ^
+	 *   X .......
+	 *   |
+	 *   S
+	 *
+	 * Where:
+	 *    .... represents an "auth_event" link
```
DMRobertson, line 44: What do the solid lines (`|` `/` `\`) denote? Guessing `prev_events`.

Not sure what the distinction between `prev_events` and `auth_events` is. "This is the last thing I saw" versus "This is what enables me to send this event"?

tests/federation_room_event_auth_test.go
```diff
@@ -0,0 +1,206 @@
...
+	 *    C is the room creation series
+	 *    R is a rejected event
+	 *    O is an outlier, which should be rejected
+	 *    X is an event with O among its auth_events, which should be rejected
+	 *           as a side-effect of O being rejected
+	 *    S is the final regular event, which acts as a sentinel
+	 *
+	 * To check if the outlier is rejected, we simply request the event via
+	 * /rooms/{roomID}/event. If it is rejected, we should get a 404.
```
DMRobertson, line 53: Sanity check: spec says that 404 means `The event was not found or you do not have permission to read this event.` Is it possible that 404 could be returned even if the event was accepted? Don't think so---we'have permission to see (at least) all events after we joined.

DMRobertson (approved):

tests/federation_room_event_auth_test.go
richvdh: hrm. It's a grey area. It's not a concept that the spec ever refers to explicitly, since it tries not to define the exact algorithms that server implementations should use for these endpoints.

That said, I know that Dendrite uses a similar concept, and, though I've not looked at Conduit's implementation, a search for `outlier` brings up some likely-looking hits, so it probably does too.

So... it's not a thing that the spec defines, but it is a common concept to Matrix server implementations.

Every so often the idea of a "server implementation guide" gets raised (see https://github.com/matrix-org/matrix.org/tree/master/implementation-guides), and then nobody has time to work on them. That might be a natural home for defining some of these concepts.

tests/federation_room_event_auth_test.go
richvdh: yes, exactly, on both fronts.

auth events are defined by the spec here: https://spec.matrix.org/unstable/server-server-api/#auth-events-selection.

tests/federation_room_event_auth_test.go
richvdh: yes, I don't think permissions should be a factor here. We're using a `public_chat` preset which makes history visible to everyone anyway.

tests/federation_room_event_auth_test.go
richvdh: added a comment to define `|`.

DMRobertson (approved):

PR #212, opened 2021-10-13
Ensure many batches are able to be imported (MSC2716)

MadLittleMods:
Ensure many batches are able to be imported and `/messages` still gives us a 200 OK.

Used to reproduce https://gitlab.com/gitterHQ/webapp/-/merge_requests/2229#note_665906390 which led to the fix in Synapse, https://github.com/matrix-org/synapse/pull/11027.

### DIFF
--- a/tests/msc2716_test.go
+++ b/tests/msc2716_test.go
@@ -226,6 +226,9 @@ func TestBackfillingHistory(t *testing.T) {
 			// eventIDsAfter
 			createMessagesInRoom(t, alice, roomID, 2)
 
+			// Import a long chain of batches connected to each other.
+			// We want to make sure Synapse doesn't blow up after we import
+			// many messages.
 			nextBatchID := ""
 			for i := 0; i < numBatches; i++ {
 				insertTime := timeAfterEventBefore.Add(timeBetweenMessages * time.Duration(numBatches-numHistoricalMessagesPerBatch*i))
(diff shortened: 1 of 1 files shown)


tests/msc2716_test.go
```diff
@@ -207,6 +207,54 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+					createMessageEventsForBatchSendRequest([]string{virtualUserID}, insertTime, numHistoricalMessagesPerBatch),
+					// Status
+					200,
+				)
+				batchSendResBody := client.ParseJSON(t, batchSendRes)
+				nextBatchID = client.GetJSONFieldStr(t, batchSendResBody, "next_batch_id")
+			}
+
+			// Ensure the /message response gives a 200
+			alice.MustDoFunc(t, "GET", []string{"_matrix", "client", "r0", "rooms", roomID, "messages"}, client.WithContentType("application/json"), client.WithQueries(url.Values{
```
kegsay, line 252: It would be nice to assert the number of events returned too, and that they have different event IDs.

MadLittleMods: 👍 Going to close this one in favor of https://github.com/matrix-org/complement/pull/214 which actually checks for all of the messages in the batches

MadLittleMods:
Thanks for the review @kegsay 🦎

PR #214, opened 2021-10-19
Draft: Test that `/messages` works on remote homeserver and can be backfilled properly after many batches (MSC2716)

MadLittleMods:
Test that `/messages` works on remote homeserver and can be backfilled properly after many batches

Related to https://github.com/matrix-org/complement/pull/212

Used while developing https://github.com/matrix-org/synapse/pull/11114

### Dev notes

```go
	logrus.WithFields(logrus.Fields{
		"stack": string(debug.Stack()),
	}).Error("destroy stack")
```

```go
var firstEvent gjson.Result
var lastEvent gjson.Result
if len(eventsJson) > 0 {
	firstEvent = eventsJson[0]
	lastEvent = eventsJson[len(eventsJson)-1]
}
logrus.WithFields(logrus.Fields{
	"events":     len(eventsJson),
	"firstEvent": fmt.Sprintf("%s (%s)", firstEvent.Get("event_id"), firstEvent.Get("type")),
	"lastEvent":  fmt.Sprintf("%s (%s)", lastEvent.Get("event_id"), lastEvent.Get("type")),
	"start":      gjson.GetBytes(messsageResBody, "start").Str,
	"end":        gjson.GetBytes(messsageResBody, "end").Str,
}).Error("Number of events from `/messages`")
```

 - https://dave.cheney.net/2020/03/10/go-test-v-streaming-output
 - https://github.com/golang/go/issues/46959#issuecomment-947881974
 - https://github.com/golang/go/issues/27826


tests/msc2716_test.go
```diff
@@ -255,19 +256,74 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+					createJoinStateEventsForBatchSendRequest([]string{virtualUserID}, insertTime),
+					createMessageEventsForBatchSendRequest([]string{virtualUserID}, insertTime, numHistoricalMessagesPerBatch),
+					// Status
+					200,
+				)
+				batchSendResBody := client.ParseJSON(t, batchSendRes)
+				// Make sure we see all of the historical messages
+				expectedEventIDs = append(expectedEventIDs, client.GetJSONFieldStringArray(t, batchSendResBody, "event_ids")...)
+				nextBatchID = client.GetJSONFieldStr(t, batchSendResBody, "next_batch_id")
+			}
```
MadLittleMods, line 302: Need to add marker event for a federated server to guarantee discovering the history

internal/client/client.go
```diff
@@ -310,7 +310,7 @@ func (c *CSAPI) DoFunc(t *testing.T, method string, paths []string, opts ...Requ
 	}
 	// debug log the request
 	if c.Debug {
-		t.Logf("Making %s request to %s", method, reqURL)
+		t.Logf("Making %s request to %s", method, req.URL)
```
MadLittleMods, line 313: This adds the query parameters to the logged URL

tests/msc2716_test.go
```diff
@@ -1003,14 +1165,14 @@ func sendMarkerAndEnsureBackfilled(t *testing.T, as *client.CSAPI, c *client.CSA
 	return markerEventID
 }
 
-func createMessagesInRoom(t *testing.T, c *client.CSAPI, roomID string, count int) (eventIDs []string) {
+func createMessagesInRoom(t *testing.T, c *client.CSAPI, roomID string, count int, messageSuffix string) (eventIDs []string) {
```
MadLittleMods, line 1168: Just making it easier to spot which messages were created at what point in the test and have some context when looking at an event ID.

tests/msc2716_test.go
```diff
@@ -913,6 +984,92 @@ func fetchUntilMessagesResponseHas(t *testing.T, c *client.CSAPI, roomID string,
...
+			len(expectedEventIDs),
+			len(leftoverEventIDs),
+			leftoverEventIDs,
+			len(actualEventIDList),
+			actualEventIDList,
+		)
+	}
+
+	for {
+		if time.Since(start) > 200*c.SyncUntilTimeout {
```
MadLittleMods, line 1019: This timeout is currently very large to accommodate the long ~20s `/messages` durations. We also have to make that request 11 times during the test which adds up very fast.

Synapse really has to chug for those requests 👹 and ideally wouldn't have to modify this at all.

I would need to look into optimizing Synapse to make this fast which we should probably do anyway as this is painfully slow.

internal/client/client.go
kegsay: Thanks! :D

tests/msc2716_test.go
```diff
@@ -255,19 +255,87 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+			eventIDsBefore := createMessagesInRoom(t, alice, roomID, 2, "eventIDsBefore")
+			eventIdBefore := eventIDsBefore[len(eventIDsBefore)-1]
+			timeAfterEventBefore := time.Now()
+
+			// We chose the magic number 11 because Synapse currently limits the
+			// backfill extremities to 5. 10 also seemed like a round number someone
+			// could pick for other homeserver implementations so I just did 10+1 to
+			// make sure it also worked in that case.
+			//numBatches := 11
+			numBatches := 2
```
kegsay, line 274: Lovely comment but then you seem to use 2. Why?

tests/msc2716_test.go
```diff
@@ -255,19 +255,87 @@ func TestImportHistoricalMessages(t *testing.T) {
 			})
 		})
 
+		t.Run("Backfill still works after many batches are imported", func(t *testing.T) {
```
kegsay, line 258: I don't think 2 batches is "many". The batch itself has many events though, so perhaps rename this test?

tests/msc2716_test.go
```diff
@@ -255,19 +255,87 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+					createMessageEventsForBatchSendRequest([]string{virtualUserID}, insertTime, numHistoricalMessagesPerBatch),
+					// Status
+					200,
+				)
+				batchSendResBody := client.ParseJSON(t, batchSendRes)
+				// Make sure we see all of the historical messages
+				expectedEventIDs = append(expectedEventIDs, client.GetJSONFieldStringArray(t, batchSendResBody, "event_ids")...)
+				nextBatchID = client.GetJSONFieldStr(t, batchSendResBody, "next_batch_id")
+
+				// Grab the base insertione event ID to reference later in the marker event
```
kegsay, line 307:
```suggestion
				// Grab the base insertion event ID to reference later in the marker event
```

tests/msc2716_test.go
```diff
@@ -255,19 +255,87 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+				if i == 0 {
+					baseInsertionEventID = client.GetJSONFieldStr(t, batchSendResBody, "base_insertion_event_id")
+				}
+			}
+
+			// Make sure we see the events at the very start of the message history
+			expectedEventIDs = append(expectedEventIDs, eventIDsBefore...)
```
kegsay, line 314: So this is new -> old?

tests/msc2716_test.go
kegsay: This is currently set to a whopping 1000s by default https://github.com/matrix-org/complement/blob/136fd601570c43fbfb1cb10647884dff2338f4b4/internal/docker/deployment.go#L57 which seems excessive even given this.

tests/msc2716_test.go
```diff
@@ -913,6 +988,91 @@ func fetchUntilMessagesResponseHas(t *testing.T, c *client.CSAPI, roomID string,
...
+			}
+
+			if len(workingExpectedEventIDMap) == 0 {
+				return
+			}
+		}
+
+		checkCounter++
+		// Add a slight delay so we don't hammmer the messages endpoint
+		time.Sleep(500 * time.Millisecond)
```
kegsay, line 1072: Uhh, why? It's a test server with only us using it? 500ms waits seems excessive. 10ms seems okay if you are concerned logic errors would consume all CPU.

tests/msc2716_test.go
```diff
@@ -913,6 +988,91 @@ func fetchUntilMessagesResponseHas(t *testing.T, c *client.CSAPI, roomID string,
...
+				generateErrorMesssageInfo(),
+			)
+		}
+
+		for _, ev := range events {
+			eventID := ev.Get("event_id").Str
+			actualEventIDList = append(actualEventIDList, eventID)
+
+			if _, keyExists := workingExpectedEventIDMap[eventID]; keyExists {
+				delete(workingExpectedEventIDMap, eventID)
```
kegsay, line 1062: So this function doesn't actually check the events are in the correct order, just that they exist?

tests/msc2716_test.go
```diff
@@ -913,6 +988,91 @@ func fetchUntilMessagesResponseHas(t *testing.T, c *client.CSAPI, roomID string,
...
+		messsageResBody := client.ParseJSON(t, messagesRes)
+
+		messageResEnd = client.GetJSONFieldStr(t, messsageResBody, "end")
+
+		wantKey := "chunk"
+		keyRes := gjson.GetBytes(messsageResBody, wantKey)
+		if !keyRes.Exists() {
+			t.Fatalf("missing key '%s'", wantKey)
+		}
+		if !keyRes.IsArray() {
```
kegsay, line 1044: Why not use `JSONArrayEach`? https://github.com/matrix-org/complement/blob/136fd601570c43fbfb1cb10647884dff2338f4b4/internal/match/json.go#L164 It does these checks for you.

kegsay (approved):

tests/msc2716_test.go
MadLittleMods: @kegsay Sorry this wasn't clear as undrafting indicates "marked this pull request as ready for review" but I didn't assign you this one yet specifically because of this problem.

The test is good to go and `numBatches` could be switched back to `11` and shipped but want to make it actually acceptable time-wise to run against Synapse before merging.

I used `numBatches=2` during testing because it's much faster to see results while developing.

tests/msc2716_test.go
MadLittleMods: See:

> I used `numBatches=2` during testing because it's much faster to see results while developing.
>
> https://github.com/matrix-org/complement/pull/214#discussion_r746829737

Running the tests against Synapse is extremely slow atm, https://github.com/matrix-org/complement/pull/214#discussion_r741678240

tests/msc2716_test.go
MadLittleMods: It will be 11 when we ship, see https://github.com/matrix-org/complement/pull/214#discussion_r747128460

tests/msc2716_test.go
MadLittleMods: Removed from both places this was happening ⏩

---

This first spawned from the original MSC2716 Complement PR(https://github.com/matrix-org/complement/pull/68) for [`fetchUntilMessagesResponseHas`](https://github.com/matrix-org/complement/pull/68/files#diff-9868c2480f7570d7028a300cdf808ea451dee2ba46adc5a0bfd5fb0b5951a2eaR697). I think the thinking was that if the message response didn't have it, just give it some time to process everything and try again. But whatever, servers are fast 🤸‍♀️

Potentially lots of `/messages` requests is hard to debug if it takes just a bit of time to process all of the events. Just ran the tests without the delay and I don't see a big difference so should be fine 👍

tests/msc2716_test.go
MadLittleMods: I think it was because we can't break early with `JSONArrayEach` (can only break early if you throw an error which fails the test).

tests/msc2716_test.go
MadLittleMods: Correct, just a checkoff function

tests/msc2716_test.go
MadLittleMods: The order doesn't matter because it's just a list of expected ID's, `expectedEventIDs`

We also just use it in check off function (`paginateUntilMessageCheckOff`) which doesn't care about the order.

---

There is a similar variable in another test where order does matter but we call it `expectedEventIDOrder`

tests/msc2716_test.go
kegsay: Is the early return really a deal breaker? I don't think it'll affect runtime performance that much?

internal/client/client.go
MadLittleMods: Split out to https://github.com/matrix-org/complement/pull/250

tests/msc2716_test.go
```diff
@@ -220,19 +220,87 @@ func TestImportHistoricalMessages(t *testing.T) {
 			validateBatchSendRes(t, as, roomID, batchSendRes, false)
 		})
 
+		t.Run("Backfill still works after many batches are imported", func(t *testing.T) {
+			t.Parallel()
```
MadLittleMods, line 224: TODO: Move this test under the `Federation` section once https://github.com/matrix-org/complement/pull/249 merges

tests/msc2716_test.go
MadLittleMods: Updated to re-use `JSONArrayEach`. It's not so bad ⏩

tests/msc2716_test.go
```diff
@@ -1022,9 +1100,104 @@ func fetchUntilMessagesResponseHas(t *testing.T, c *client.CSAPI, roomID string,
...
+		must.MatchResponse(t, messagesRes, match.HTTPResponse{
+			JSON: []match.JSON{
+				match.JSONArrayEach("chunk", func(ev gjson.Result) error {
+					foundEventInMessageResponse = true
+					eventID := ev.Get("event_id").Str
+					actualEventIDList = append(actualEventIDList, eventID)
+
+					if _, keyExists := denyEventIDMap[eventID]; keyExists {
+						return fmt.Errorf(
+							"paginateUntilMessageCheckOff found unexpected message=%s in deny list while paginating. %s",
```
MadLittleMods, line 1175: 👆 Calling out a change since the last review pass. I added a `denyListEventIDs` so we can ensure that the historical state doesn't appear between batches of historical messages while we're paginating.

tests/msc2716_test.go
```diff
@@ -99,7 +99,7 @@ func TestImportHistoricalMessages(t *testing.T) {
 			//
```
MadLittleMods, line 99: Reviewers note: A lot of the diff has been split out into separate PRs:

 - https://github.com/matrix-org/complement/pull/250
 - https://github.com/matrix-org/complement/pull/248
 - https://github.com/matrix-org/complement/pull/251
 - https://github.com/matrix-org/complement/pull/249

MadLittleMods: All of these are merged in now 🚀

kegsay:
@MadLittleMods what is the status of this PR? It's still marked as a Draft.

tests/msc2716_test.go
MadLittleMods: > @MadLittleMods what is the status of this PR? It's still marked as a Draft.
>
> https://github.com/matrix-org/complement/pull/214#issuecomment-1123426080

Status is still the same as the last update in this thread. It's too slow on Synapse for me to be comfortable merging it yet so runs for everyone.

tests/msc2716_test.go
```diff
@@ -876,6 +876,77 @@ func TestImportHistoricalMessages(t *testing.T) {
 					},
 				})
 			})
+
+			t.Run("Backfill still works after many batches are imported", func(t *testing.T) {
```
MadLittleMods, line 880: Make it obvious that these batches are connected together

```suggestion
			t.Run("Backfill still works after many contiguous batches are imported", func(t *testing.T) {
```

tests/msc2716_test.go
MadLittleMods: In terms of optimizing Synapse to make this test viable to run time-wise, I'm a bit blocked on a race condition in some recent code, https://github.com/matrix-org/synapse/pull/12394#discussion_r872012615

richvdh: > In terms of optimizing Synapse to make this test viable to run time-wise, I'm a bit blocked on a race condition in some recent code, [matrix-org/synapse#12394 (comment)](https://github.com/matrix-org/synapse/pull/12394#discussion_r872012615) -> [matrix-org/synapse#12646](https://github.com/matrix-org/synapse/issues/12646)

Now https://github.com/matrix-org/synapse/issues/12988 I think

kegsay:
Marked as stale; this PR will be closed in 7 days if there isn't any further progress. Please re-open when you get time. Long-lived branches are fine, but long-lived PRs like this always being in `Draft` is not fine.

tests/msc2716_test.go
MadLittleMods: Made progress on optimizing Synapse:

 - Saves 10s per `/messages` call: https://github.com/matrix-org/synapse/pull/13863 
 - Still need to figure out https://github.com/matrix-org/synapse/pull/13864

tests/msc2716_test.go
```diff
@@ -1294,6 +1366,12 @@ func paginateUntilMessageCheckOff(t *testing.T, c *client.CSAPI, roomID string,
 		if len(workingExpectedEventIDMap) == 0 {
 			return
 		}
+
+		// Since this will throw an error if they key does not exist,
+		// do this at the end of the loop. It's a valid scenario to be at the end
+		// of the room and have no more to paginate so we want the `return` when
+		// we've found all of the expected events to have a chance to run first.
+		messageResEnd = client.GetJSONFieldStr(t, messsageResBody, "end")
```
MadLittleMods, line 1374: Split out to https://github.com/matrix-org/complement/pull/489

PR #221, opened 2021-11-05
Draft: Make sure historical state events doesn't come down `/sync` or `/transactions` (MSC2716)

MadLittleMods:
Make sure historical state events doesn't come down [`/sync`](https://matrix.org/docs/spec/client_server/latest#get-matrix-client-r0-sync) (client-server API)  or [`/transactions`](https://matrix.org/docs/spec/application_service/r0.1.2#put-matrix-app-v1-transactions-txnid)(Application service API)

Adding regression tests for https://github.com/matrix-org/synapse/issues/11241

### DIFF
--- a/tests/msc2716_test.go
+++ b/tests/msc2716_test.go
@@ -15,6 +15,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/sirupsen/logrus"
 	"github.com/tidwall/gjson"
 
 	"github.com/matrix-org/complement/internal/b"
@@ -255,7 +256,7 @@ func TestImportHistoricalMessages(t *testing.T) {
 			})
 		})
 
-		t.Run("Historical events from /batch_send do not come down in an incremental sync", func(t *testing.T) {
+		t.Run("Historical events from batch_send do not come down in an incremental sync", func(t *testing.T) {
 			t.Parallel()
 
 			roomID := as.CreateRoom(t, createPublicRoomOpts)
@@ -268,6 +269,10 @@ func TestImportHistoricalMessages(t *testing.T) {
 
 			// Create some "live" events to saturate and fill up the /sync response
 			createMessagesInRoom(t, alice, roomID, 5)
+
+			// Get a /sync since token we can try paginating from after
+			// we batch send
+			since := doInitialSync(t, alice)
 
 			// Import a historical event
 			batchSendRes := batchSendHistoricalMessages(
@@ -283,19 +288,32 @@ func TestImportHistoricalMessages(t *testing.T) {
 			)
 			batchSendResBody := client.ParseJSON(t, batchSendRes)
 			historicalEventIDs := client.GetJSONFieldStringArray(t, batchSendResBody, "event_ids")
-			historicalEventId := historicalEventIDs[0]
+			historicalStateEventIDs := client.GetJSONFieldStringArray(t, batchSendResBody, "state_event_ids")
 
-			// This is just a dummy event we search for after the historicalEventId
+			// This is just a dummy event we search for after the historicalEventIDs/historicalStateEventIDs
 			eventIDsAfterHistoricalImport := createMessagesInRoom(t, alice, roomID, 1)
 			eventIDAfterHistoricalImport := eventIDsAfterHistoricalImport[0]
 
-			// Sync until we find the eventIDAfterHistoricalImport.
-			// If we're able to see the eventIDAfterHistoricalImport that occurs after
-			// the historicalEventId without seeing eventIDAfterHistoricalImport in
-			// between, we're probably safe to assume it won't sync
-			alice.SyncUntil(t, "", `{ "room": { "timeline": { "limit": 3 } } }`, "rooms.join."+client.GjsonEscape(roomID)+".timeline.events", func(r gjson.Result) bool {
-				if r.Get("event_id").Str == historicalEventId {
-					t.Fatalf("We should not see the %s historical event in /sync response but it was present", historicalEventId)
+			logrus.WithFields(logrus.Fields{
+				"eventIDAfterHistoricalImport": eventIDAfterHistoricalImport,
+				"historicalEventIDs":           historicalEventIDs,
+				"historicalStateEventIDs":      historicalStateEventIDs,
+			}).Error("afewfeew")
+
+			// Sync from before we did any batch sending until we find the
+			// eventIDAfterHistoricalImport. If we're able to see
+			// eventIDAfterHistoricalImport without any the
+			// historicalEventIDs/historicalStateEventIDs in between, we're probably
+			// safe to assume it won't sync.
+			alice.SyncUntil(t, since, "", "rooms.join."+client.GjsonEscape(roomID)+".timeline.events", func(r gjson.Result) bool {
+				logrus.WithFields(logrus.Fields{
+					"event_id":  r.Get("event_id").Str,
+					"type":      r.Get("type").Str,
+					"state_key": r.Get("state_key").Str,
+				}).Error("event from sync")
+
+				if includes(r.Get("event_id").Str, historicalEventIDs) || includes(r.Get("event_id").Str, historicalStateEventIDs) {
+					t.Fatalf("We should not see the %s historical event in /sync response but it was present", r.Get("event_id").Str)
 				}
 
 				return r.Get("event_id").Str == eventIDAfterHistoricalImport
@@ -924,6 +942,16 @@ func reversed(in []string) []string {
 		out[i] = in[len(in)-i-1]
 	}
 	return out
+}
+
+func includes(needle string, haystack []string) bool {
+	for _, item := range haystack {
+		if needle == item {
+			return true
+		}
+	}
+
+	return false
 }
 
 func fetchUntilMessagesResponseHas(t *testing.T, c *client.CSAPI, roomID string, check func(gjson.Result) bool) {


tests/msc2716_test.go
```diff
@@ -255,7 +256,7 @@ func TestImportHistoricalMessages(t *testing.T) {
 			})
 		})
 
-		t.Run("Historical events from /batch_send do not come down in an incremental sync", func(t *testing.T) {
+		t.Run("Historical events from batch_send do not come down in an incremental sync", func(t *testing.T) {
```
MadLittleMods, line 259: Removing the `/` in the test name so I can actually run it individually, `COMPLEMENT_ALWAYS_PRINT_SERVER_LOGS=1 COMPLEMENT_DIR=../complement ./scripts-dev/complement.sh TestImportHistoricalMessages/parallel/Historical_events_from_batch_send_do_not_come_down_in_an_incremental_sync`

See https://gist.github.com/MadLittleMods/4ab08f51609fab759247f299a4e33406 for why there is a problem using a `/` to match a single test.

tests/msc2716_test.go
```diff
@@ -283,19 +287,20 @@ func TestImportHistoricalMessages(t *testing.T) {
...
-			// between, we're probably safe to assume it won't sync
-			alice.SyncUntil(t, "", `{ "room": { "timeline": { "limit": 3 } } }`, "rooms.join."+client.GjsonEscape(roomID)+".timeline.events", func(r gjson.Result) bool {
-				if r.Get("event_id").Str == historicalEventId {
-					t.Fatalf("We should not see the %s historical event in /sync response but it was present", historicalEventId)
+			// Sync from before we did any batch sending until we find the
+			// eventIDAfterHistoricalImport. If we're able to see
+			// eventIDAfterHistoricalImport without any the
+			// historicalEventIDs/historicalStateEventIDs in between, we're probably
+			// safe to assume it won't sync.
+			alice.SyncUntil(t, since, "", "rooms.join."+client.GjsonEscape(roomID)+".timeline.events", func(r gjson.Result) bool {
```
MadLittleMods, line 301: Made this test more clear on what's happening. We now paginate sync from before we `/batch_send` until an event that occurred after `/batch_send` and just make sure we didnt' see any historical state or events in between.

tests/msc2716_test.go
```diff
@@ -283,19 +287,20 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+				if includes(r.Get("event_id").Str, historicalEventIDs) || includes(r.Get("event_id").Str, historicalStateEventIDs) {
```
MadLittleMods, line 302: Here is where we additionally check for the `historicalStateEventIDs` now

tests/msc2716_test.go
```diff
@@ -283,25 +289,122 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+					},
+				})
+
+				// Acknowledge that we've seen the transaction
+				w.WriteHeader(200)
+				w.Write([]byte("{}"))
+			}).Methods("PUT")
+
+			srv := &http.Server{
+				Addr:    ":9111",
```
MadLittleMods, line 346: I'm assuming we probably want to generate and make this random. It's a bit tricky since it also has to match the port in the registration.

Instead of specifying a `URL`, we could probably generate a `URL` in `blueprints.go` (like we do with `HSToken` and `ASToken`) when we process the blueprint so the registration is generated correctly and the info is available in our test again as `deployment.HS["hs1"].ApplicationServices`

MadLittleMods: Updated to use random port ⏩

internal/b/blueprints.go
```diff
@@ -18,10 +18,16 @@ import (
...
 	"encoding/hex"
 	"fmt"
+	"net"
 	"strconv"
 	"strings"
 )
 
+var (
+	// HostnameRunningComplement is the hostname of Complement from the perspective of a Homeserver.
+	HostnameRunningComplement = "host.docker.internal"
```
MadLittleMods, line 28: Can't import this from `docker.HostnameRunningComplement` because then we will have a circular dependency

tests/msc2716_test.go
```diff
@@ -283,25 +293,139 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+		t.Run("Historical events from batch_send do not get pushed out as application service transactions", func(t *testing.T) {
+			t.Parallel()
+
+			// Find the application service port defined in the registration file
+			asRegistration := deployment.HS["hs1"].ApplicationServices["my_as_id"]
+			asURLMatches := asURLRegexp.FindStringSubmatch(asRegistration)
+			if asURLMatches == nil {
+				t.Fatalf("Unable to find application service `url` in registration=%s", asRegistration)
+			}
+			asPort := asURLMatches[2]
```
MadLittleMods, line 325: Parsing yaml with regex probably isn't very good, "Not a regular language, blah" but didn't want to pull in a yaml parser here. Feel free to poke in a different direction.

internal/b/blueprints.go
```diff
@@ -206,6 +228,18 @@ func normalizeApplicationService(as ApplicationService) (ApplicationService, err
...
+	if as.URL == "" {
+		// Since, we're just checking and not reserving the port, we could
+		// potentially run into an issue where the port is no longer available when
+		// we actually try to bind to it later on
+		port, err := getFreePort()
+		if err != nil {
+			return as, err
+		}
+
+		as.URL = fmt.Sprintf("http://%s:%d", HostnameRunningComplement, port)
```
MadLittleMods, line 240: Just checking, do we even care about having a random port for the application service?

Previously I just had it run on port `9111` but changed it to a random port in https://github.com/matrix-org/complement/pull/221/commits/06935a72aa3a5fa00fd458949c63ed993294e85d

tests/msc2716_test.go
```diff
@@ -283,25 +293,139 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+
+			// Create a listener and handler to stub an application service listening
+			// for transactions from a homeserver.
```
MadLittleMods, line 328: Perhaps we want to create `NewApplicationService` to mirror `federation.NewServer`?

The Application service here just needs to listen to a single endpoint though (just an http listener), nothing special.

tests/msc2716_test.go
```diff
@@ -283,25 +293,139 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+					// Tests will likely fail if the server is not listening anyways
+					t.Logf("Failed to listen and serve our fake application service: %s", err)
+				}
+			}()
+			defer func() {
+				err := srv.Shutdown(context.Background())
+				if err != nil {
+					t.Fatalf("Failed to shutdown our fake application service: %s", err)
+				}
+			}()
```
MadLittleMods, line 373: This is inspired by the [`Server.Listen`](https://github.com/matrix-org/complement/blob/11adc36c58c29c0603ded8a6276563276d2d5649/internal/federation/server.go#L252-L280) code

internal/b/blueprints.go
```diff
@@ -206,6 +228,19 @@ func normalizeApplicationService(as ApplicationService) (ApplicationService, err
 	as.HSToken = hex.EncodeToString(hsToken)
 	as.ASToken = hex.EncodeToString(asToken)
 
+	if as.URL == "" {
+		// Since, we're just checking and not reserving the port, we could
+		// potentially run into an issue where the port is no longer available when
+		// we actually try to bind to it later on
```
kegsay, line 234: Tests are run concurrently, so we really do care about this.

internal/b/blueprints.go
```diff
@@ -190,6 +196,22 @@ func normaliseUser(u string, hsName string) (string, error) {
 	return u, nil
 }
 
+// Asks the kernel for a free open port that is ready to use.
+// via https://github.com/phayes/freeport/blob/95f893ade6f232a5f1511d61735d89b1ae2df543/freeport.go#L7-L20
+func getFreePort() (int, error) {
```
kegsay, line 201: This isn't the right place for this.

internal/b/blueprints.go
kegsay: Yes, tests are run concurrently so we can't bind to the same port all the time.

internal/b/blueprints.go
kegsay: That doesn't mean you should make the constant here, that means you should move the logic to where we actually care about the host/port aka `/docker`.

tests/msc2716_test.go
kegsay: Just pull in a YAML parser.

kegsay (changes requested):
Complement lacks full AS support. This PR adds enough to make it work but in a way I'm unhappy with. This PR also adds in regression tests for `/sync` which I'm happy with.

Please split out the `/sync` stuff into a different PR (which can be accepted) and leave this PR with me. I do not want to merge this AS work in its current shape because:
 - It has concurrency issues.
 - It asks the test writer to do way too much to set up an AS listener (parsing the AS YAML, manually setting up routers/servers and listening and serving, etc)
 - The locations of some of this code feel extremely out of place with the rest of the surrounding code.

I'll go through and actually add AS support to Complement, using this test as a guide for how tests will make use of the functionality, and then we can merge in the `/transactions` bits.

tests/msc2716_test.go
```diff
@@ -248,25 +258,139 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+			alice.SyncUntil(t, since, "", "rooms.join."+client.GjsonEscape(roomID)+".timeline.events", func(r gjson.Result) bool {
+				if includes(r.Get("event_id").Str, historicalEventIDs) || includes(r.Get("event_id").Str, historicalStateEventIDs) {
+					t.Fatalf("We should not see the %s historical event in /sync response but it was present", r.Get("event_id").Str)
 				}
 
 				return r.Get("event_id").Str == eventIDAfterHistoricalImport
 			})
 		})
 
+		t.Run("Historical events from batch_send do not get pushed out as application service transactions", func(t *testing.T) {
```
MadLittleMods, line 281: > Please split out the `/sync` stuff into a different PR (which can be accepted) and leave this PR with me.

Thanks for the review @kegsay 🐢, I've split out the `/sync` stuff to https://github.com/matrix-org/complement/pull/235

kegsay:
I hope to get around to adding AS support to Complement soon.

PR #222, opened 2021-11-12
Use sub-tests for restricted join tests

neilalexander:
The sub-tests make it clearer which parts of the test are failing and which are succeeding.

This also adds `MatchSuccess` and `MatchFailure` for wider checks of whether something had failed or not (it's either in the HTTP 200 range or it isn't), rather than checking specific error codes.

### DIFF
--- a/tests/restricted_rooms_test.go
+++ b/tests/restricted_rooms_test.go
@@ -71,11 +71,11 @@ func setupRestrictedRoom(t *testing.T, deployment *docker.Deployment) (*client.C
 func checkRestrictedRoom(t *testing.T, alice *client.CSAPI, bob *client.CSAPI, allowed_room string, room string) {
 	t.Helper()
 
-	t.Run("JoinShouldFailInitially", func(t *testing.T) {
+	t.Run("Join should fail initially", func(t *testing.T) {
 		failJoinRoom(t, bob, room, "hs1")
 	})
 
-	t.Run("JoinShouldSucceedWhenJoinedToAllowedRoom", func(t *testing.T) {
+	t.Run("Join should succeed when joined to allowed room", func(t *testing.T) {
 		// Join the allowed room, attempt to join the room again, which now should succeed.
 		bob.JoinRoom(t, allowed_room, []string{"hs1"})
 		bob.JoinRoom(t, room, []string{"hs1"})
@@ -98,7 +98,7 @@ func checkRestrictedRoom(t *testing.T, alice *client.CSAPI, bob *client.CSAPI, a
 		)
 	})
 
-	t.Run("JoinShouldFailWhenLeftAllowedRoom", func(t *testing.T) {
+	t.Run("Join should fail when left allowed room", func(t *testing.T) {
 		// Leaving the room works and the user is unable to re-join.
 		bob.LeaveRoom(t, room)
 		bob.LeaveRoom(t, allowed_room)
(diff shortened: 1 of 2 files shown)


clokep:
Maybe this is a misunderstanding on my part, but don't sub-tests run in parallel? That's generally why I haven't used them in complement tests.

neilalexander:
> Maybe this is a misunderstanding on my part, but don't sub-tests run in parallel?

They only run in parallel if you call `t.Parallel()` at the beginning of the subtest.

tests/restricted_rooms_test.go
```diff
@@ -73,88 +71,98 @@ func setupRestrictedRoom(t *testing.T, deployment *docker.Deployment) (*client.C
...
-		b.Event{
-			Type:     "m.room.member",
-			Sender:   bob.UserID,
-			StateKey: &bob.UserID,
-			Content: map[string]interface{}{
-				"membership":  "join",
-				"displayname": "Bobby",
-				// This should be ignored since this is a join -> join transition.
-				"join_authorised_via_users_server": "unused",
+	t.Run("JoinShouldSucceedWhenJoinedToAllowedRoom", func(t *testing.T) {
```
kegsay, line 78:
```suggestion
	t.Run("Join should succeed when joined to allowed room", func(t *testing.T) {
```

We don't need to camel case here.

tests/restricted_rooms_test.go
```diff
@@ -73,88 +71,98 @@ func setupRestrictedRoom(t *testing.T, deployment *docker.Deployment) (*client.C
 func checkRestrictedRoom(t *testing.T, alice *client.CSAPI, bob *client.CSAPI, allowed_room string, room string) {
 	t.Helper()
 
-	failJoinRoom(t, bob, room, "hs1", 403)
-
-	// Join the allowed room, attempt to join the room again, which now should succeed.
-	bob.JoinRoom(t, allowed_room, []string{"hs1"})
-	bob.JoinRoom(t, room, []string{"hs1"})
+	t.Run("JoinShouldFailInitially", func(t *testing.T) {
```
kegsay, line 74:
```suggestion
	t.Run("Join should fail initially", func(t *testing.T) {
```

tests/restricted_rooms_test.go
```diff
@@ -73,88 +71,98 @@ func setupRestrictedRoom(t *testing.T, deployment *docker.Deployment) (*client.C
...
-		t,
-		room,
-		b.Event{
-			Type:     "m.room.join_rules",
-			Sender:   alice.UserID,
-			StateKey: &emptyStateKey,
-			Content: map[string]interface{}{
-				"join_rule": "restricted",
-				"allow":     []string{"invalid"},
+	t.Run("JoinShouldFailWithMangledJoinRules", func(t *testing.T) {
```
kegsay, line 130:
```suggestion
	t.Run("Join should fail with mangled join rules", func(t *testing.T) {
```

tests/restricted_rooms_test.go
```diff
@@ -73,88 +71,98 @@ func setupRestrictedRoom(t *testing.T, deployment *docker.Deployment) (*client.C
...
+
+			return ev.Get("content").Get("membership").Str == "leave"
+		})
 
-		return ev.Get("content").Get("membership").Str == "leave"
+		failJoinRoom(t, bob, room, "hs1")
 	})
 
-	failJoinRoom(t, bob, room, "hs1", 403)
+	t.Run("JoinShouldSucceedWhenInvited", func(t *testing.T) {
```
kegsay, line 120:
```suggestion
	t.Run("Join should succeed when invited", func(t *testing.T) {
```

tests/restricted_rooms_test.go
```diff
@@ -73,88 +71,98 @@ func setupRestrictedRoom(t *testing.T, deployment *docker.Deployment) (*client.C
...
 			},
-		},
-	)
+		)
+	})
 
-	// Leaving the room works and the user is unable to re-join.
-	bob.LeaveRoom(t, room)
-	bob.LeaveRoom(t, allowed_room)
+	t.Run("JoinShouldFailWhenLeftAllowedRoom", func(t *testing.T) {
```
kegsay, line 101:
```suggestion
	t.Run("Join should fail when left allowed room", func(t *testing.T) {
```

kegsay (approved):

kegsay:
`--- FAIL: TestRestrictedRoomsRemoteJoinFailOver (15.34s)` still though.

PR #229, opened 2021-11-19
Add ServerRoom.MustHaveMembershipForUser

### LINKED ISSUE #137: federation.ServerRoom should have CheckMembershipForUser
See https://github.com/matrix-org/complement/pull/131/files

kegsay:
As per https://github.com/matrix-org/complement/pull/131/files#r658674594

Fixes #137

### DIFF
--- a/internal/federation/server_room.go
+++ b/internal/federation/server_room.go
@@ -113,6 +113,7 @@ func (r *ServerRoom) AuthChain() (chain []*gomatrixserverlib.Event) {
 	return
 }
 
+// Check that the user currently has the membership provided in this room. Fails the test if not.
 func (r *ServerRoom) MustHaveMembershipForUser(t *testing.T, userID, wantMembership string) {
 	t.Helper()
 	state := r.CurrentState("m.room.member", userID)
--- a/internal/federation/server_room.go
+++ b/internal/federation/server_room.go
@@ -3,6 +3,7 @@ package federation
 import (
 	"encoding/json"
 	"fmt"
+	"testing"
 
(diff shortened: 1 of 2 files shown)


internal/federation/server_room.go
```diff
@@ -112,6 +113,21 @@ func (r *ServerRoom) AuthChain() (chain []*gomatrixserverlib.Event) {
 	return
 }
 
+func (r *ServerRoom) MustHaveMembershipForUser(t *testing.T, userID, wantMembership string) {
```
richvdh, line 116: a brief doc comment might not go amiss, even if it's a bit obvious.

richvdh (approved):
it feels like a bit of a shame to lose the context about when the test is happening - "after invite" / "after reject", but I'll not fight for it.

Thanks!

kegsay:
You don't lose context because this is using `t.Helper()` so the file/line number will be the stack frame above. In other words you'll see for example:
```
federation_room_invite_test.go:54: no membership state for @alice:localhost
federation_room_invite_test.go:60: no membership state for @alice:localhost
```
and the surrounding code at those line numbers give you the context of after invite / after reject.

richvdh:
> You don't lose context because this is using `t.Helper()` so the file/line number will be the stack frame above. In other words you'll see for example:
> 
> ```
> federation_room_invite_test.go:54: no membership state for @alice:localhost
> federation_room_invite_test.go:60: no membership state for @alice:localhost
> ```

it'd be nice to know what the context is without having to correlate line numbers, but I agree, this helps.

Issue #227, opened 2021-11-19
Support per-test blacklists


kegsay:
Rather than just per-file like we have currently.

PR #230, opened 2021-11-19
Add ability to skip individual tests

kegsay:
- Added `runtime.SkipIf`
- Re-combined federation test as it can now be individually skipped on Dendrite.

From a CI perspective, no changes are required as this uses the existing build
tags setup to detect HS impl. Fixes https://github.com/matrix-org/complement/issues/227


neilalexander (approved):
Overall PR looks fine. Not sure if it is idiomatic to overload a standard library package name (`runtime`) but I don't suppose in this case it really matters given that the `runtime` packgae is really quite specialised.

kegsay:
That was my thinking: it's rare to use `runtime` so overloading it seems fine to me. I don't think it's idiomatic or not idiomatic to overload a stdlib package name, it depends on your application imo.

PR #235, opened 2021-11-23
Make sure historical state events don't come down incremental `/sync` (MSC2716)

MadLittleMods:

Related to https://github.com/matrix-org/synapse/issues/11241

Split off from https://github.com/matrix-org/complement/pull/221

### DIFF
--- a/tests/msc2716_test.go
+++ b/tests/msc2716_test.go
@@ -220,7 +220,7 @@ func TestImportHistoricalMessages(t *testing.T) {
 			validateBatchSendRes(t, as, roomID, batchSendRes, false)
 		})
 
-		t.Run("Historical events from /batch_send do not come down in an incremental sync", func(t *testing.T) {
+		t.Run("Historical events from batch_send do not come down in an incremental sync", func(t *testing.T) {
 			t.Parallel()
 
 			roomID := as.CreateRoom(t, createPublicRoomOpts)
@@ -233,6 +233,10 @@ func TestImportHistoricalMessages(t *testing.T) {
 
 			// Create some "live" events to saturate and fill up the /sync response
 			createMessagesInRoom(t, alice, roomID, 5)
+
+			// Get a /sync `since` pagination token we can try paginating from later
+			// on
+			since := doInitialSync(t, alice)
 
 			// Import a historical event
 			batchSendRes := batchSendHistoricalMessages(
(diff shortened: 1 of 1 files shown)


tests/msc2716_test.go
```diff
@@ -248,19 +252,20 @@ func TestImportHistoricalMessages(t *testing.T) {
...
-			// between, we're probably safe to assume it won't sync
-			alice.SyncUntil(t, "", `{ "room": { "timeline": { "limit": 3 } } }`, "rooms.join."+client.GjsonEscape(roomID)+".timeline.events", func(r gjson.Result) bool {
-				if r.Get("event_id").Str == historicalEventId {
-					t.Fatalf("We should not see the %s historical event in /sync response but it was present", historicalEventId)
+			// Sync from before we did any batch sending until we find the
+			// eventIDAfterHistoricalImport. If we're able to see
+			// eventIDAfterHistoricalImport without any the
+			// historicalEventIDs/historicalStateEventIDs in between, we're probably
+			// safe to assume it won't sync.
+			alice.SyncUntil(t, since, "", "rooms.join."+client.GjsonEscape(roomID)+".timeline.events", func(r gjson.Result) bool {
```
MadLittleMods, line 266: Made this test more clear on what's happening. We now paginate sync from before we `/batch_send` until an event that occurred after `/batch_send` and just make sure we didnt' see any historical state or events in between.

tests/msc2716_test.go
```diff
@@ -248,19 +252,20 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+				if includes(r.Get("event_id").Str, historicalEventIDs) || includes(r.Get("event_id").Str, historicalStateEventIDs) {
```
MadLittleMods, line 267: Here is where we additionally check for the `historicalStateEventIDs` now

tests/msc2716_test.go
```diff
@@ -220,7 +220,7 @@ func TestImportHistoricalMessages(t *testing.T) {
 			validateBatchSendRes(t, as, roomID, batchSendRes, false)
 		})
 
-		t.Run("Historical events from /batch_send do not come down in an incremental sync", func(t *testing.T) {
+		t.Run("Historical events from batch_send do not come down in an incremental sync", func(t *testing.T) {
```
MadLittleMods, line 223: Removing the `/` in the test name so I can actually run it individually, `COMPLEMENT_ALWAYS_PRINT_SERVER_LOGS=1 COMPLEMENT_DIR=../complement ./scripts-dev/complement.sh TestImportHistoricalMessages/parallel/Historical_events_from_batch_send_do_not_come_down_in_an_incremental_sync`

See https://gist.github.com/MadLittleMods/4ab08f51609fab759247f299a4e33406 for why there is a problem using a `/` to match a single test.

kegsay (approved):

MadLittleMods:
Thanks for the review @kegsay 🐿

PR #236, opened 2021-11-25
Alter container startup to wait for container healthcheck to OK.

michaelkaye:
Wait for healthcheck (if it exists) to return OK before checking version endpoint; In all cases, ensure the version endpoint is available before starting.

Users with an image containing a healthcheck they do not want to use can make a tweak to their container by adding
```
HEALTHCHECK NONE
```
to their docker file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ If you're looking to run against a custom Dockerfile, it must meet the following
 
 - The Dockerfile must `EXPOSE 8008` and `EXPOSE 8448` for client and federation traffic respectively.
 - The homeserver should run and listen on these ports.
+- The homeserver should become healthy within `COMPLEMENT_SPAWN_HS_TIMEOUT_SECS` if a `HEALTHCHECK` is specified in the Dockerfile.
 - The homeserver needs to `200 OK` requests to `GET /_matrix/client/versions`.
 - The homeserver needs to manage its own storage within the image.
 - The homeserver needs to accept the server name given by the environment variable `SERVER_NAME` at runtime.
--- a/dockerfiles/Synapse.Dockerfile
+++ b/dockerfiles/Synapse.Dockerfile
@@ -31,6 +31,3 @@ WORKDIR /data
 EXPOSE 8008 8448
 
 ENTRYPOINT ["/conf/start.sh"]
-
-HEALTHCHECK --start-period=5s --interval=1s --timeout=1s \
-    CMD curl -fSs http://localhost:8008/health || exit 1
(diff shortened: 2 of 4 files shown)


kegsay (approved):
I assume this is for more complex containers that want to say, wait for a postgres to be up and listening first, which may race with 200 OKing the /versions endpoint which can just return static data.

It's questionable whether this should be Complement's job to wait for _both_ healthchecks and `/versions` but given it's an optional extra and isn't too much more complexity, this looks like a good addition to me and would definitely aid HS devs when writing valid Complement images. LGTM

kegsay:
Please update https://github.com/matrix-org/complement#image-requirements mentioning the HEALTHCHECK.

michaelkaye:
> It's questionable whether this should be Complement's job to wait for _both_ healthchecks and `/versions` but given it's an optional extra and isn't too much more complexity, this looks like a good addition to me and would definitely aid HS devs when writing valid Complement images. LGTM

In my head, the healthcheck is waiting for (whatever the container is) to be ready; the /versions check is for "is this actually now being a matrix homeserver"; I was wondering if we needed any overrides or options for this.

> Please update https://github.com/matrix-org/complement#image-requirements mentioning the HEALTHCHECK.

Will do.

Is it a problem that the Tests / complement isn't passing; i've been trying to kick it various ways to work?

kegsay:
Yes it's a problem. We don't normally see synapse tests timing out after 10min. It should run on 3mins.

kegsay:
There's definitely something wrong here, CI seems unhappy even on repeat runs. My guess is that the health check never passes or something?

ShadowJonathan:
This may be fixed by https://github.com/matrix-org/synapse/pull/11533, btw

kegsay:
This cannot merge until https://github.com/matrix-org/synapse/issues/8780 is fixed unfortunately.

michaelkaye:
Apologies for the commit noise - am reverting my tweak to see if that breaks the build and the fix is a real fix.

michaelkaye:
I should mention that after edeb928  was committed, the synapse tests ran in about 3min

Having reverted in 0dfa0bf , the tests have failed again.

I think the tweak is probably valid; I don't want to apply it in the upstream synapse healthcheck configuration, as for production-type purposes, checking 15s after a restart is a good policy; but for the rapid-fire complement usecase, it feels like a reduced delay is the right thing. I wonder if this is causing some issue like a resource exhaustion of some kind - maybe too many socket opens or something; which will happen because each test takes 15s to start, so there are 30 attempts to check healthcheck (500ms between checks)

Re-applying my fix now, if that's green I suggest we should merge.

michaelkaye:
The conflict is quite big; remade this PR on another branch, see #293

PR #237, opened 2021-11-29
Reproduce matrix-org/synapse#9768

DMRobertson:


tests/csapi/direct_message_create_test.go
```diff
@@ -0,0 +1,56 @@
...
+	bob := deployment.Client(t, "hs1", "@bob:hs1")
+
+	// Alice invites Bob to a room for direct messaging.
+	roomID := alice.CreateRoom(t, map[string]interface{}{
+		"invite":    []string{bob.UserID},
+		"is_direct": true,
+	})
+
+	// Bob receives the invite and joins.
+	next_batch := bob.SyncUntilInvitedTo(t, roomID)
```
kegsay, line 25: Please use `camelCase` instead for variable names.

tests/csapi/direct_message_create_test.go
```diff
@@ -0,0 +1,56 @@
...
+	bob.JoinRoom(t, roomID, []string{"hs1"})
+
+	// Wait until Bob sees that he's joined.
+	isBobJoinEvent := func(ev gjson.Result) bool {
+		return ev.Get("type").Str == "m.room.member" && ev.Get("state_key").Str == bob.UserID && ev.Get("content").Get("membership").Str == "join"
+	}
+	bob.SyncUntilTimelineHas(t, roomID, isBobJoinEvent)
+
+	// Repeat the sync, but in a form where we can inspect the output.
```
kegsay, line 34: This doesn't make sense because `SyncUntilTimelineHas` returns each event in the callback `func(ev gjson.Result)`. You can do any inspection or assertion you like on the event, and indeed you already do on `:30` to check the type/state_key/membership.

tests/csapi/direct_message_create_test.go
```diff
@@ -0,0 +1,56 @@
...
+	res := bob.MustDoFunc(t, "GET", []string{"_matrix", "client", "r0", "sync"}, client.WithQueries(query))
+	body := client.ParseJSON(t, res)
+	eventRes := gjson.GetBytes(body, "rooms.join").Get(roomID).Get("timeline.events")
+	if !eventRes.IsArray() {
+		t.Fatal("events in timeline was not an array: ", eventRes.Raw)
+	}
+	events := eventRes.Array()
+	bobJoinEvents := 0
+	for _, ev := range events {
+		if isBobJoinEvent(ev) {
```
kegsay, line 49: This doesn't make sense, why are you asserting the same thing twice?

tests/csapi/direct_message_create_test.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"net/url"
+	"testing"
+
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/client"
+	"github.com/tidwall/gjson"
+)
+
+func TestDirectMessageCreate(t *testing.T) {
```
kegsay, line 12: Not a good test name. I had to read https://github.com/matrix-org/synapse/issues/9768 to figure out what the regression test here is for. Please at the very least mention duplicate membership event _somewhere_.

tests/csapi/direct_message_create_test.go
kegsay: This makes more sense now that I know you're testing for duplicate events.

kegsay (changes requested):

kegsay:
The test file needs to be formatted with `go fmt` and it isn't. Please set up your IDE to auto-format on save.

In addition, it's not clear if this test case is race-free. We assume in the test currently that the duplicate event will appear immediately after the `/sync` returns the first event. It would be better if we instead sent a sentinel event (i.e some random other event _in the same room_) and then only treat this test as a success when that sentinel event is received down `/sync`, as that guarantees that the member event has been fully processed.

DMRobertson:
> The test file needs to be formatted with `go fmt` and it isn't.

The required incantation was

```shell
goimports -local github.com/matrix-org/complement -w internal tests
```

Which I suggest we write down in ONBOARDING.md.

kegsay:
@DMRobertson please update this PR to use the new functional options for syncing as shown in https://github.com/matrix-org/complement/pull/272 - you shouldn't need to modify `client.go` anymore now! :D

kegsay:
@DMRobertson what is the status of this PR?

DMRobertson:
> @DMRobertson what is the status of this PR?

It's on my wishlist of things to get to if I find time. Not planning to actively work on the linked Synapse issue though; feel free to close if it helps manage the repo. (I can always reopen.)

kegsay:
Closing on the basis that you can re-open it later. This PR is just adding noise.

PR #238, opened 2021-11-30
Improve federation_keys_test.go

ShadowJonathan:
I was taking a cursory glance at `federation_keys_test.go`, and comparing it with its sytest equivalent, when i noticed that the bare test would slip through a lot of potential problems, type mismatches, and other invariants.

I've cleaned those up a little, but some decisions are a bit opinionated.

### DIFF
--- a/tests/federation_keys_test.go
+++ b/tests/federation_keys_test.go
@@ -29,60 +29,215 @@ import (
 func TestInboundFederationKeys(t *testing.T) {
 	deployment := Deploy(t, b.BlueprintAlice)
 	defer deployment.Destroy(t)
+
 	fedClient := &http.Client{
 		Timeout:   10 * time.Second,
 		Transport: &docker.RoundTripper{Deployment: deployment},
 	}
+
 	res, err := fedClient.Get("https://hs1/_matrix/key/v2/server")
 	must.NotError(t, "failed to GET /keys", err)
-	var key ed25519.PublicKey
-	var keyID string
+
+	var keys map[string]ed25519.PublicKey
+	var oldKeys map[string]ed25519.PublicKey
+
 	body := must.MatchResponse(t, res, match.HTTPResponse{
 		StatusCode: 200,
 		JSON: []match.JSON{
-			match.JSONKeyPresent("old_verify_keys"),
+			match.JSONKeyTypeEqual("valid_until_ts", gjson.Number),
 			match.JSONKeyEqual("server_name", "hs1"),
+
+			// Check validity of verify_keys+old_verify_keys and store values
 			match.JSONMapEach("verify_keys", func(k, v gjson.Result) error {
 				// Currently we always expect ed25519 keys
 				if !strings.HasPrefix(k.Str, "ed25519:") {
-					return fmt.Errorf("complement: unknown key ID type, only ed25519 is supported, got '%s'", k.Str)
+					return fmt.Errorf("complement: Key '%s' has no 'ed25519:' prefix", k.Str)
 				}
+
+				key := v.Get("key")
+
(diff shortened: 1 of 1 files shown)


tests/federation_keys_test.go
```diff
@@ -29,60 +29,226 @@ import (
...
 	wantTime := time.Now()
 	if gotTime.Before(wantTime) {
-		t.Errorf("valid_until_ts : timestamp is in the past: %s < %s", gotTime, wantTime)
+		t.Errorf("valid_until_ts: timestamp is in the past: %s < %s", gotTime, wantTime)
+	}
+
+	checkKeysAndSignatures(t, body, jsonObj, keys, oldKeys)
+}
+
+func checkKeysAndSignatures(t *testing.T, body []byte, jsonObj gjson.Result, keys, oldKeys map[string]ed25519.PublicKey) {
```
kegsay, line 150: It seems needless to make an additional function here given it's only used in this one test, and makes assumptions like it being `hs1`. I would just inline this function into the test personally.

tests/federation_keys_test.go
```diff
@@ -29,60 +29,226 @@ import (
...
+
+	// Test signatures for all verify_keys, these *have* to exist.
+	for keyName, keyBytes := range keys {
+		sigBase64 := jsonObj.Get(fmt.Sprintf("signatures.hs1.%s", keyName))
+		if !sigBase64.Exists() {
+			t.Fatalf("signatures: missing signature for Key '%s'", keyName)
+		}
+
+		if sigBase64.Type != gjson.String {
+			t.Fatalf("signatures: Signature for Key '%s' has unexpected type, expected String, got '%s'", keyName, sigBase64.Type.String())
```
kegsay, line 202: You can use `must.GetJSONFieldStr(t, bodyBytes, fmt.Sprintf("signatures.hs1.%s", keyName))` to get string values from keys, which does checks for type/existence.

kegsay (approved):
Overall good, this is much more thorough than before, thanks!

ShadowJonathan:
I still think that the test does not much to invoke edgecase behaviour that's being tested for here, I imagine that more than half of this test should be a generic helper function that vets a server's response the same exhaustive way under different conditions.

tests/federation_keys_test.go
ShadowJonathan: I'm not sure if that's possible, the CI Linter complained about complexity :(

tests/federation_keys_test.go
```diff
@@ -29,60 +29,226 @@ import (
...
+			// Check signatures
+			match.JSONMapEach("signatures", func(k, v gjson.Result) error {
+				if !v.IsObject() {
+					return fmt.Errorf("signatures: Value for server '%s' is not an object", k.Str)
+				}
+
+				for key, v := range v.Map() {
+					// Check for ed25519 prefix
+					if !strings.HasPrefix(key, "ed25519:") {
+						return fmt.Errorf("signatures: Key '%s' for server '%s' has no 'ed25519:' prefix", key, k.Str)
```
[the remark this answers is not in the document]
ShadowJonathan: [It is](https://spec.matrix.org/v1.1/server-server-api/#get_matrixkeyv2serverkeyid);

```json5
{
  "old_verify_keys": {
    "ed25519:0ldk3y": {
      // ...
    }
  },
  "server_name": "example.org",
  "signatures": {
    "example.org": {  // <-
      // ...
    }
  },
  "valid_until_ts": 1652262000000,
  "verify_keys": {
    "ed25519:abc123": {
      // ...
    }
  }
}
```

kegsay (approved):

PR #244, opened 2021-12-01
Add test for SYN-442

ShadowJonathan:
This;
- adds the sytest test for SYN-442: "Test that we can be reinvited to a room we created"
- adds `SyncUntilJoined` helper function


ShadowJonathan:
Apparently some of these endpoints are racy on synapse, https://github.com/matrix-org/matrix-doc/issues/3533 is related.

Adding `SyncUntil` checks for now.

kegsay (approved):
Excellent.

Issue #245, opened 2021-12-01
[nit] `SyncUntil` does not mention in its name it only matches `key` on arrays


ShadowJonathan:
Initially i thought that this function was generic enough to just match `key` and throw that variable into the check function, but turns out that's not the case.

I suggest renaming it to `SyncUntilArray`, and adding a generic function `SyncUntil` which does above expected behaviour.

kegsay:
https://github.com/matrix-org/complement/pull/272

PR #246, opened 2021-12-01
Change `SyncUntil` to `SyncUntilArray`

ShadowJonathan:
Fixes https://github.com/matrix-org/complement/issues/245

Turns out i actually do need a `SyncUtil` as described above (for SYN-627, I need to receive events from different rooms and log them individually as they come in, I need to wait for and process an arbitrary value), so just I'll rename this alongside it.

I'm refactoring this in a seperate PR instead of the same PR for SYN-627, because this PR would be easier to review comparatively when bundled with the other helper function changes.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -216,11 +216,11 @@ func (c *CSAPI) SyncUntil(t *testing.T, since, filter, key string, check func(gj
 func (c *CSAPI) syncUntilInternal(t *testing.T, since, filter, key string, wasFailed, timedOut *bool, checkCounter *int, check func(gjson.Result) bool) {
 	t.Helper()
 	start := time.Now()
-	*checkCounter = 0
 
 	// Initialize values
 	*wasFailed = t.Failed()
 	*timedOut = false
+	*checkCounter = 0
 
 	for {
 		if time.Since(start) > c.SyncUntilTimeout {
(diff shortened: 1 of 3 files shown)


internal/client/client.go
```diff
@@ -148,38 +148,84 @@ func (c *CSAPI) SyncUntilInvitedTo(t *testing.T, roomID string) {
...
+func (c *CSAPI) SyncUntil(t *testing.T, since, filter, key string, check func(gjson.Result) bool) {
+	var wasFailed *bool
+	var timedOut *bool
+	var checkCounter *int
+
+	c.syncUntilInternal(t, since, filter, key, wasFailed, timedOut, checkCounter, check)
+}
+
+// Internal function helping both SyncUntil and SyncUntilArray with some logging coordination upon failure
+func (c *CSAPI) syncUntilInternal(t *testing.T, since, filter, key string, wasFailed, timedOut *bool, checkCounter *int, check func(gjson.Result) bool) {
```
ShadowJonathan, line 216: This function signature is a bit bonkers, but essentially, most of these are used by `SyncUntilArray` to "feed back" the values in its `defer`'d function (as `t.Fatalf` just stops the goroutine like a panic and runs all defers up the stack, and i wanted to separate `SyncUntil` and `SyncUntilArray` functionality.

kegsay:
https://github.com/matrix-org/complement/pull/272

PR #248, opened 2021-12-02
Send marker events in all federation tests (MSC2716)

MadLittleMods:
Send marker events in all federation tests 🎯

The marker event is important to let remote homeservers know there are historical messages at the given insertion event it points to.

It's just a lucky happen-stance that the tests passed before without the need for the marker event. Marker events are required/expected in order for the history to be properly visible on remote homeservers.

Split out from https://github.com/matrix-org/complement/pull/214


kegsay (approved):

MadLittleMods:
Thanks for the review @kegsay 🐵

PR #251, opened 2021-12-02
Add `messageSuffix` to more easily spot which message is which when debugging (MSC2716)

MadLittleMods:
Add `messageSuffix` to more easily spot which message is which when debugging

Split off from https://github.com/matrix-org/complement/pull/214


kegsay (approved):

MadLittleMods:
Thanks for taking a look @kegsay 🦌

Issue #25, opened 2020-09-09
Add sync tests


kegsay:
See https://github.com/matrix-org/dendrite/pull/1394#issuecomment-687056673

PR #257, opened 2021-12-02
Address matrix-org/dendrite#1394 bugs

ShadowJonathan:
Fixes https://github.com/matrix-org/complement/issues/25

Adds `SyncLeftFrom`

cc @neilalexander

### DIFF
--- a/tests/csapi/sync_test.go
+++ b/tests/csapi/sync_test.go
@@ -3,7 +3,6 @@ package csapi_tests
 import (
 	"net/url"
 	"testing"
-	"time"
 
 	"github.com/tidwall/gjson"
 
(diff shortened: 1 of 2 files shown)


tests/csapi/sync_test.go
```diff
@@ -0,0 +1,116 @@
...
+	body := client.ParseJSON(t, res)
+	since = client.GetJSONFieldStr(t, body, "next_batch")
+
+	alice.JoinRoom(t, roomID, nil)
+
+	// We can't SyncUntil or else "pollute" the sync internal state, we must do as-if we're a slow client.
+	// Thus, two-second delays.
+	// WARNING: If you're wondering why this test is failing in a flaky fashion, and are looking at this test,
+	// this may the reason why, sorry :(
+	time.Sleep(2 * time.Second)
```
kegsay, line 41: The bug report makes no mention of syncing between the join/leaves, so we shouldn't need to sleep here. It should be fine to join->leave->join -> sleep(1 second) -> do the sync. Try not to sleep a lot as this will slow down complement runs for everyone all the time.

tests/csapi/sync_test.go
```diff
@@ -0,0 +1,116 @@
...
+	body = client.ParseJSON(t, res)
+	jsonRes := gjson.GetBytes(body, "rooms.leave."+client.GjsonEscape(roomID))
+	if !jsonRes.Exists() {
+		t.Errorf("Bob just rejected an invite, it should show up under 'leave' in a full sync")
+	}
+	since = client.GetJSONFieldStr(t, body, "next_batch")
+
+	bob.JoinRoom(t, roomID, nil)
+
+	res = bob.MustDoFunc(t, "GET", []string{"_matrix", "client", "r0", "sync"}, client.WithQueries(url.Values{
```
kegsay, line 106: Need to wait until the join comes down before doing the full sync else this races.

kegsay (approved):
Otherwise this is great, thanks!

tests/csapi/sync_test.go
ShadowJonathan: Just because the bug report makes no mention of syncing, doesnt mean that *doing* the sync is allowed.

I'm taking a conservative approach here by sleeping here, but i'll give this a look again later.

kegsay (approved):

PR #260, opened 2021-12-02
Improve versions test a bit

ShadowJonathan:
This has an opinionated section, even though the spec doesnt really strictly mandate version strings in an explicit sense, i thought it is reasonable to add a regex matcher in here.

### DIFF
--- a/tests/csapi/apidoc_version_test.go
+++ b/tests/csapi/apidoc_version_test.go
@@ -2,6 +2,7 @@ package csapi_tests
 
 import (
 	"fmt"
+	"regexp"
 	"testing"
 
 	"github.com/tidwall/gjson"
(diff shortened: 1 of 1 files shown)


kegsay (changes requested):
I don't mind checking `unstable_features` but I do mind regexing the versions given the spec is silent on it.

ShadowJonathan:
Can I emit a warning instead? I still think that there's some degree of expectation that complement could look at.

If not, i'll remove it here and just add a spec issue, though i should probably do that anyways.

kegsay:
Warnings won't be looked at as they won't appear on success. Please just remove it for now and file a spec issue, xreffing it as a comment would be fantastic so we know why we don't regexp them.

ShadowJonathan:
@kegsay, given https://github.com/matrix-org/matrix-doc/issues/3594#issuecomment-1003130527, it turns out the spec isn't silent about this after all, I'd have to adjust the regex a little bit, but do you think that is good enough to re-add this to this test?

kegsay:
Yup, sounds good to me.

tests/csapi/apidoc_version_test.go
```diff
@@ -11,13 +12,32 @@ import (
...
+		res := client.MustDoFunc(t, "GET", []string{"_matrix", "client", "versions"})
+
+		// Matches;
+		// - r0.?.?
+		//  where ? is any single digit
+		// - v1^.*(-#)
+		//  where 1^ is 1 through 9 for the first digit, then any digit thereafter,
+		//  and * is any single or multiple of digits
+		//  optionally with dash-separated metadata: (-#)
+		versionRegex, _ := regexp.Compile("^(" + r0Regex + "|" + GlobalVersionRegex + ")$")
```
kegsay, line 40: It is common to inline this in the global variable as `var regex = regexp.MustCompile(....)` we should probably just do that and move these lovely comments to there as well.

PR #261, opened 2021-12-02
Add 7 history visibility tests (3 sytests)

ShadowJonathan:
This adds;

```
./tests/30rooms/51event.pl:test "/event/ does not allow access to events before the user joined",
./tests/30rooms/51event.pl:test "/event/ on joined room works",
./tests/30rooms/51event.pl:test "/event/ on non world readable room does not work",
```

But also 4 other tests.

All 7;
> **TestFetchEvent:**
> Fetches an event after join, and succeeds.
> sytest: /event/ on joined room works

> **TestFetchHistoricalJoinedEventDenied:**
> Tries to fetch an event before join, and fails.
> history_visibility: joined
> sytest: /event/ does not allow access to events before the user joined

> **TestFetchHistoricalSharedEvent:**
> Tries to fetch an event before join, and succeeds.
> history_visibility: shared
Skipped in dendrite: https://github.com/matrix-org/dendrite/issues/617

> **TestFetchHistoricalInvitedEventFromBetweenInvite:**
> Tries to fetch an event between being invited and joined, and succeeds.
> history_visibility: invited
Skipped in dendrite

> **TestFetchHistoricalInvitedEventFromBeforeInvite:**
> Tries to fetch an event before being invited, and fails.
> history_visibility: invited

> **TestFetchEventNonWorldReadable:**
> Tries to fetch an event without having joined, and fails.
> history_visibility: shared
> sytest: /event/ on non world readable room does not work

> **TestFetchEventWorldReadable:**
> Tries to fetch an event without having joined, and succeeds.
> history_visibility: world_readable
Skipped in dendrite

### DIFF
--- a/tests/csapi/apidoc_room_history_visibility_test.go
+++ b/tests/csapi/apidoc_room_history_visibility_test.go
@@ -1,7 +1,6 @@
 package csapi_tests
 
 import (
-	"github.com/matrix-org/complement/runtime"
 	"net/http"
 	"testing"
 
@@ -11,6 +10,7 @@ import (
 	"github.com/matrix-org/complement/internal/client"
 	"github.com/matrix-org/complement/internal/match"
 	"github.com/matrix-org/complement/internal/must"
+	"github.com/matrix-org/complement/runtime"
 )
 
 // TODO most of this can be refactored into data-driven tests
--- a/tests/csapi/apidoc_room_history_visibility_test.go
+++ b/tests/csapi/apidoc_room_history_visibility_test.go
@@ -27,6 +27,7 @@ func createRoomWithVisibility(t *testing.T, c *client.CSAPI, visibility string) 
 					"history_visibility": visibility,
 				},
 				"type": "m.room.history_visibility",
+				"state_key": "",
 			},
 		},
 		"preset": "public_chat",
--- a/tests/csapi/apidoc_room_history_visibility_test.go
+++ b/tests/csapi/apidoc_room_history_visibility_test.go
@@ -169,8 +169,7 @@ func TestFetchHistoricalInvitedEventFromBetweenInvite(t *testing.T) {
 	roomID := createRoomWithVisibility(t, alice, "invited")
 
 	alice.InviteRoom(t, roomID, bob.UserID)
-
-	bob.SyncUntilInvitedTo(t, roomID)
+	bob.MustSyncUntil(t, client.SyncReq{}, client.SyncInvitedTo(bob.UserID, roomID))
 
 	eventID := alice.SendEventSynced(t, roomID, b.Event{
 		Type: "m.room.message",
@@ -225,8 +224,7 @@ func TestFetchHistoricalInvitedEventFromBeforeInvite(t *testing.T) {
 	})
 
 	alice.InviteRoom(t, roomID, bob.UserID)
-
-	bob.SyncUntilInvitedTo(t, roomID)
+	bob.MustSyncUntil(t, client.SyncReq{}, client.SyncInvitedTo(bob.UserID, roomID))
 
 	bob.JoinRoom(t, roomID, nil)
 	bob.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(bob.UserID, roomID))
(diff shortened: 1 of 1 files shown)


tests/csapi/apidoc_room_history_visibility_test.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func createRoomWithVisibility(t *testing.T, c *client.CSAPI, visibility string) string {
+	return c.CreateRoom(t, map[string]interface{}{
+		"initial_state": []map[string]interface{}{
+			{
+				"content": map[string]interface{}{
+					"history_visibility": visibility,
+				},
+				"type": "m.room.history_visibility",
```
kegsay, line 29: Suprised this works without a `state_key` explicitly set. Can we do this please?

tests/csapi/apidoc_room_history_visibility_test.go
```diff
@@ -0,0 +1,337 @@
...
+	"github.com/tidwall/gjson"
+
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
+	"github.com/matrix-org/complement/runtime"
+)
+
+// TODO most of this can be refactored into data-driven tests
```
kegsay, line 16: Up to you if you want to try this out in this PR or not. Idiomatic Go would have table-based tests like:
```go
testCases := []struct{
    visibility string  // shared, joined, etc 
    requesterMembershipState string // typically join, sometimes invite
    wantResponse match.HTTPResponse
    skipOnRuntimes []string // dendrite
}{
    { // example for TestFetchEvent
        visibility: "shared",
        requesterMembershipState: "join",
        wantResponse: match.HTTPResponse{
		StatusCode: 200,
		JSON: []match.JSON{
			// No harm in checking if the event data is also as expected
			match.JSONKeyEqual("content", map[string]interface{}{
				"msgtype": "m.text",
				"body":    "Hello world",
			}),
			match.JSONKeyEqual("type", "m.room.message"),

			// the spec technically doesn't list these following keys, but we're still checking them because sytest did.
			// see: https://github.com/matrix-org/matrix-doc/issues/3540
			match.JSONKeyEqual("room_id", roomID),
			match.JSONKeyEqual("sender", alice.UserID),
			match.JSONKeyEqual("event_id", eventID),
			match.JSONKeyTypeEqual("origin_server_ts", gjson.Number),
		},
	},
    }
}
deployment := Deploy(t, b.BlueprintOneToOneRoom)
defer deployment.Destroy(t)
bob := deployment.Client(t, "hs1", "@bob:hs1")

for _, tc := range testCases {
	roomID := createRoomWithVisibility(t, alice, tc.visibility)
	switch (tc.requesterMembershipState) {
		case "join":
			bob.JoinRoom(t, roomID, nil)
			bob.SyncUntilTimelineHas(t, roomID, func(event gjson.Result) bool {
				return event.Get("type").Str == "m.room.member" &&
					event.Get("content.membership").Str == "join" &&
					event.Get("state_key").Str == bob.UserID
			})
		// etc...
	}

	eventID := alice.SendEventSynced(t, roomID, b.Event{
		Type: "m.room.message",
			"msgtype": "m.text",
			"body":    "Hello world",
		},
	})

	res := fetchEvent(t, bob, roomID, eventID)

	must.MatchResponse(t, res, tc.wantResponse)
}
```   

Probably should be wrapped in sub-tests but you get the idea. It would be nice to see this done in this PR given the sheer number of visibility tests this is adding.

kegsay (approved):

tests/csapi/apidoc_room_history_visibility_test.go
ShadowJonathan: Hmmm, thanks for the example, I'll try this then, and wrap them in sub-tests

tests/csapi/apidoc_room_history_visibility_test.go
kegsay, line 29:
```suggestion
				"type": "m.room.history_visibility",
				"state_key": "",
```

PR #263, opened 2021-12-02
Add 3 sytests

ShadowJonathan:
This adds 3 miscellaneous sytests;
- Rooms can be created with an initial invite list (SYN-205)
- Getting push rules doesn't corrupt the cache SYN-390
- Non-present room members cannot ban others

### DIFF
--- a/tests/csapi/room_ban_test.go
+++ b/tests/csapi/room_ban_test.go
@@ -1,13 +1,14 @@
 package csapi_tests
 
 import (
-	"io/ioutil"
 	"testing"
 
 	"github.com/tidwall/gjson"
 
 	"github.com/matrix-org/complement/internal/b"
 	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
 )
 
 // This is technically a tad different from the sytest, in that it doesnt try to ban a @random_dude:test,
@@ -70,9 +71,7 @@ func TestNotPresentUserCannotBanOthers(t *testing.T) {
 		"reason":  "testing",
 	}))
 
-	if res.StatusCode != 403 {
-		defer res.Body.Close()
-		body, _ := ioutil.ReadAll(res.Body)
-		t.Fatalf("Banning a user succeeded where it shouldn't: (%d) %s", res.StatusCode, string(body))
-	}
+	must.MatchResponse(t, res, match.HTTPResponse{
+		StatusCode: 403,
+	})
 }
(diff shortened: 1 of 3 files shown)


tests/csapi/room_ban_test.go
```diff
@@ -0,0 +1,78 @@
...
+	})
+
+	bob.JoinRoom(t, roomID, nil)
+
+	alice.SendEventSynced(t, roomID, b.Event{
+		Type:     "m.room.power_levels",
+		StateKey: b.Ptr(""),
+		Content: map[string]interface{}{
+			"users": map[string]interface{}{
+				charlie.UserID: 100,
```
kegsay, line 56: I'm surprised this doesn't throw a 4xx as iirc Alice cannot give herself 0 power?

tests/csapi/room_ban_test.go
```diff
@@ -0,0 +1,78 @@
...
+			event.Get("content.membership").Str == "join" &&
+			event.Get("state_key").Str == bob.UserID
+	})
+
+	res := charlie.DoFunc(t, "POST", []string{"_matrix", "client", "r0", "rooms", roomID, "ban"}, client.WithJSONBody(t, map[string]interface{}{
+		"user_id": bob.UserID,
+		"reason":  "testing",
+	}))
+
+	if res.StatusCode != 403 {
```
kegsay, line 73: You can use:
```go
    StatusCode: 403,
}); 
```
This will consume the body and print it with the error message just as you do further down.

kegsay (approved):
Excellent PR, thank you!

tests/csapi/room_ban_test.go
ShadowJonathan: Right, i made this PR a bit earlier before i started using that more consistently, i'll change to match :+1:

tests/csapi/room_ban_test.go
ShadowJonathan: This wasn't intentional, but now that you're pointing it out, it *is* odd that this is somehow allowed.

There's either two things happening here;
- `m.room.power_levels` arent set, and so anyone can send them (according to the spec)
- `m.room.power_levels` is set, and alice (with PL 100) can "demote" herself while also promoting someone else, as the auth check to post a PL is checked before the outcome of that post is considered

I'm curious what the spec says about this though, and this is definitely something that should be clarified.

ShadowJonathan: Made an issue; https://github.com/matrix-org/matrix-doc/issues/3545

PR #267, opened 2021-12-03
Test that invites from ignored users are omitted from incremental syncs

DMRobertson:
Reproduces matrix-org/synapse#11506.

Leaving this as a draft until I fix the issue in Synapse. Sticking up a PR for now to see what Dendrite does.

Also note:  fd3a555  is cherry-picked from #237

### DIFF
--- a/tests/csapi/ignored_users_test.go
+++ b/tests/csapi/ignored_users_test.go
@@ -20,9 +20,9 @@ import (
 // that
 // > Servers must not send room invites from ignored users to clients.
 //
-// Synapse does not have this property, as detailed in
-// https://github.com/matrix-org/synapse/issues/11506.
-// This reproduces that bug.
+// This is a regression test for
+// https://github.com/matrix-org/synapse/issues/11506
+// to ensure that Synapse complies with this part of the spec.
 func TestInviteFromIgnoredUsersDoesNotAppearInSync(t *testing.T) {
 	deployment := Deploy(t, b.BlueprintCleanHS)
 	defer deployment.Destroy(t)
--- a/tests/csapi/ignored_users_test.go
+++ b/tests/csapi/ignored_users_test.go
@@ -48,7 +48,7 @@ func TestInviteFromIgnoredUsersDoesNotAppearInSync(t *testing.T) {
 	alice.MustDoFunc(
 		t,
 		"PUT",
-		[]string{"_matrix", "client", "v3", "user", alice.UserID, "account_data", "m.ignored_user_list"},
+		[]string{"_matrix", "client", "r0", "user", alice.UserID, "account_data", "m.ignored_user_list"},
 		client.WithJSONBody(t, map[string]interface{}{
 			"ignored_users": map[string]interface{}{
 				bob.UserID: map[string]interface{}{},
@@ -91,7 +91,7 @@ func TestInviteFromIgnoredUsersDoesNotAppearInSync(t *testing.T) {
 	response := alice.MustDoFunc(
 		t,
 		"GET",
-		[]string{"_matrix", "client", "v3", "sync"},
+		[]string{"_matrix", "client", "r0", "sync"},
 		client.WithQueries(queryParams),
 	)
 	bobRoomPath := "rooms.invite." + client.GjsonEscape(bobRoom)
--- a/tests/csapi/ignored_users_test.go
+++ b/tests/csapi/ignored_users_test.go
@@ -47,10 +47,8 @@ func TestInviteFromIgnoredUsersDoesNotAppearInSync(t *testing.T) {
 		"PUT",
 		[]string{"_matrix", "client", "v3", "user", alice.UserID, "account_data", "m.ignored_user_list"},
 		client.WithJSONBody(t, map[string]interface{}{
-			"content": map[string]interface{}{
-				"ignored_users": map[string]interface{}{
-					bob.UserID: map[string]interface{}{},
-				},
+			"ignored_users": map[string]interface{}{
+				bob.UserID: map[string]interface{}{},
 			},
 		}),
 	)
--- a/tests/csapi/ignored_users_test.go
+++ b/tests/csapi/ignored_users_test.go
@@ -1,0 +1,92 @@
+package csapi_tests
+
+import (
+	"net/url"
+	"testing"
+
+	"github.com/matrix-org/complement/internal/b"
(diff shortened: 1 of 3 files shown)


DMRobertson:
Fails on Dendrite:
```
 PUT hs1/_matrix/client/v3/user/@alice:hs1/account_data/m.ignored_user_list => 404 Not Found (722.71µs)
```

Guessing this isn't implemented in Dendrite yet: if so, suggest blacklisting this test.

ShadowJonathan:
You can do that yourself by adding `runtime.SkipIf(t, runtime.Dendrite)` to the start of the test, fwiw

DMRobertson:
> Guessing this isn't implemented in Dendrite yet: if so, suggest blacklisting this test.

matrix-org/dendrite#600

DMRobertson:
> You can do that yourself by adding `runtime.SkipIf(t, runtime.Dendrite)` to the start of the test, fwiw

My understanding is that [`build constraints`](https://pkg.go.dev/cmd/go#hdr-Build_constraints) are the done thing here, as in https://github.com/matrix-org/complement/pull/267/commits/0b453d965a052c8ed146c825ddfe6bdd54f8da91

ShadowJonathan:
Maybe, but more recently, skipping specific tests were introduced, these are triggered on `*_blacklist` in the same way.

tests/csapi/ignored_users_test.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// The Spec says here
+//     https://spec.matrix.org/v1.1/client-server-api/#server-behaviour-13
+// that
+// > Servers must not send room invites from ignored users to clients.
+//
+// Synapse does not have this property, as detailed in
+// https://github.com/matrix-org/synapse/issues/11506.
+// This reproduces that bug.
```
clokep, line 25: Maybe?

```suggestion
// Synapse does not handle this properly, as detailed in
// https://github.com/matrix-org/synapse/issues/11506.
// This reproduces that bug.
```

I don't think this is true though since this is now fixed, it might be better to say:

> This is a regression test for `link to issue`.

tests/csapi/ignored_users_test.go
```diff
@@ -0,0 +1,105 @@
...
+func TestInviteFromIgnoredUsersDoesNotAppearInSync(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintCleanHS)
```
clokep, line 27: This test could be much simpler, I think, by starting with `BlueprintOneToOneRoom`? I'm not sure the third user is very helpful here.

clokep:
I requested review from @kegsay since I believe he has thoughts about the generic infrastructure of complement.

tests/csapi/ignored_users_test.go
DMRobertson: The third user was an attempt to stop this test from being racey. We want to be convinced that the server has processed Bob's invite and chosen to ignore it. I could assert that there's no invite from Bob in Alice's next sync, but... what if that's only true because Alice synced really quickly?

In https://github.com/matrix-org/complement/pull/237#issuecomment-982475273,Kegan suggested sending and awaiting a dummy event. The third user was my attempt at doing something similar here.

clokep: I think you could do that by sending a dummy user into the room, yes.

DMRobertson: Oh, would Alice expect to see state updates for a room she's invited to, but hasn't joined yet? If so that would be much clearer.

clokep: > Oh, would Alice expect to see state updates for a room she's invited to, but hasn't joined yet? If so that would be much clearer.

No, they would not see those. I think I'm confused though -- isn't this just checking that some sort of events got processed?

tests/csapi/ignored_users_test.go
```diff
@@ -0,0 +1,105 @@
...
+	defer deployment.Destroy(t)
+	alice := deployment.RegisterUser(t, "hs1", "alice", "sufficiently_long_password_alice")
+	bob := deployment.RegisterUser(t, "hs1", "bob", "sufficiently_long_password_bob")
+	chris := deployment.RegisterUser(t, "hs1", "chris", "sufficiently_long_password_chris")
+
+	// Alice creates a room for herself.
+	public_room := alice.CreateRoom(t, map[string]interface{}{
```
kegsay, line 34: Please do not use underscores for variable names.

tests/csapi/ignored_users_test.go
```diff
@@ -0,0 +1,105 @@
...
+				ev.Get("state_key").Str == alice.UserID &&
+				ev.Get("content.membership").Str == "join"
+		},
+	)
+
+	// Alice ignores Bob.
+	alice.MustDoFunc(
+		t,
+		"PUT",
+		[]string{"_matrix", "client", "v3", "user", alice.UserID, "account_data", "m.ignored_user_list"},
```
kegsay, line 51: Dendrite doesn't implement any `/v3/` endpoint. Is there some spec documentation on the changes between `r0` and `v3`?

kegsay (approved):
For now, this is fine and I'm happy for it to be merged.

For later, this is not fine. I've filed https://github.com/matrix-org/complement/issues/271 and will retrospectively fix up this PR to use the new API when it lands.

tests/csapi/ignored_users_test.go
clokep: > Dendrite doesn't implement any `/v3/` endpoint. Is there some spec documentation on the changes between `r0` and `v3`?

All r0 endpoints became v3 with Matrix 1.1 -- @DMRobertson Maybe we should just use r0 for now here?

tests/csapi/ignored_users_test.go
DMRobertson: Sorry, old habits. Is there some way to configure `gofmt`/`goimports` or VSCode to flag this up?

tests/csapi/ignored_users_test.go
DMRobertson: Happy to use r0; I was just blindly going by what the latest spec said.

DMRobertson:
Using r0 means that we PUT the account data, but the ignored invite isn't still ignored. I'd forgotten that  matrix-org/dendrite#600 was a thing.

clokep (approved):

PR #269, opened 2021-12-08
Use server's default room version to create new rooms

richvdh:
When we are creating a test room on the Complement server, use the same version
as the server defaults to.


kegsay (approved):

kegsay:
Dendrite is unhappy because `tests/msc2836_test.go:203:30: undefined: roomVer`

PR #275, opened 2021-12-25
Add SYN-627 test

ShadowJonathan:
Adds;
- `./tests/90jira/SYN-627.pl:test "Events come down the correct room",`
- `WaiterGroup`

I needed functionality like `sync.WaitGroup`, which'd fail on a particular timeout passed.

Previously, this function was just in the same `apidoc_room_event_test.go` file, but I moved to a generic file after i realised that didn't make much sense.


tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+			<-sem
+		}()
+	}
+
+	var rooms []string
+
+	for i := 0; i < roomAmount; i++ {
+		// This implicitly waits until all rooms are done creating
+		rooms = append(rooms, <-chanRooms)
```
kegsay, line 60: You can/should be `range`ing over `chanRooms` and close it once the last room has been sent, rather than do this.

kegsay (changes requested):

tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	const roomAmount = 30
+
+	// Create all rooms
+
+	var chanRooms = make(chan string, roomAmount)
+
+	// semaphore, to reduce load on the server when creating rooms
+	// on slow PCs, increasing this will time out the test :(
+	var sem = make(chan bool, 1)
```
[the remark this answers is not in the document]
ShadowJonathan: I'm not entirely sure how to replicate those on golang, as golang doesn't have a concept of specific thread pools, this'll emulate that "pool", only with a semaphore to block/allow goroutines.

If there's a better suggestion how to handle this, I'll implement it, but for now, this is what I know (what works).

tests/csapi/apidoc_room_event_test.go
ShadowJonathan: That's hard to do with a large amount of concurring goroutines, so I'm not sure how to do this.

tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// sytest: Events come down the correct room
+func TestEventsInCorrectRoom(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+
+	const roomAmount = 30
```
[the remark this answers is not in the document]
ShadowJonathan: Iirc this was the original amount in the sytest.

Alternatively, we can just not add this sytest, as I'm questioning its usefulness.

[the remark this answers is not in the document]
kegsay: I'd keep it at 10 or something more manageable unless we are explicitly trying to test a limit or something.

tests/csapi/main_test.go
```diff
@@ -127,3 +127,45 @@ func (w *Waiter) Finish() {
...
+type WaiterGroup struct {
+	*sync.Mutex
+	finishCh chan bool
+	count    uint
+}
+
+// NewWaiterGroup returns a struct similar to sync.WaitGroup, and is a "sibling" to Waiter
+// A WaiterGroup can start with a specific count, which then can only go down with every
+// WaiterGroup.Done invocation.
+func NewWaiterGroup(count uint) *WaiterGroup {
```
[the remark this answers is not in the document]
ShadowJonathan: I've removed this entirely in a new commit.

ShadowJonathan:
I removed `WaiterGroup` as I don't really know what added value it'd bring to the repository. It can be re-added if you want, but to more easily ship this PR, i'm removing it.

tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+	// Create all rooms
+
+	var doneRooms = make(chan string, roomAmount)
+	var todoRooms = make(chan bool, goConcurrency)
+
+	for i := 0; i < goConcurrency; i++ {
+		go func() {
+			for {
+				_, ok := <-todoRooms
```
kegsay, line 46: There's a tidier way to do this:
```go
for _ = range todoRooms {
    // ...
}
```
This is saying "consume from `todoRooms` repeatedly until it is closed, then exit the for loop", which is what you are effectively doing here.

tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,206 @@
...
+	// get current next_batch
+	_, since := alice.MustSync(t, client.SyncReq{TimeoutMillis: "0"})
+
+	type txnAndId struct {
+		txn    int
+		roomId string
+	}
+
+	var doneEvents = make(chan string, roomAmount)
+	var todoEvents = make(chan txnAndId, goConcurrency)
```
kegsay, line 89: It doesn't make sense to me to set the buffer of this channel to `goConcurrency` and not `roomAmount`. `todoEvents` is injected with `roomAmount` items, so this will cause things to block on sending (because `goConcurrency < roomAmount`) for no justified reason.

tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,206 @@
...
+		}()
+	}
+
+	for _, id := range rooms {
+		todoEvents <- txnAndId{
+			roomId: id,
+			txn:    txnCount,
+		}
+
+		txnCount++
```
kegsay, line 121: You might as well just have `for i, id := range rooms` and then inject `i` rather than using a new variable `txnCount` and have to increment it.

tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,206 @@
...
+			}
+
+			if !eventsJson.IsArray() {
+				t.Fatalf("Expected events array in sync response to be an array: %s", eventsJson)
+			}
+
+			events := eventsJson.Array()
+
+			if len(events) > 1 {
+				t.Fatalf("Expected 0 or 1 events, got %d events: %s", len(events), events)
```
kegsay, line 169: Is this fair to say? We don't ever set a timeline_limit filter. These rooms are freshly made, so will include the initial set of state events, so it could entirely reasonably be >1 events.

EDIT: Oh but we are using a since token that is after the initial room creations, right. Needs a comment!

tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,206 @@
...
+			rooms = append(rooms, newRoom)
+		case <-time.After(channelTimeout):
+			t.Fatalf("Room creation timed out, got to %d rooms (from %d)", i, roomAmount)
+		}
+	}
+
+	// Send events to all rooms with corresponding roomID
+
+	// get current next_batch
+	_, since := alice.MustSync(t, client.SyncReq{TimeoutMillis: "0"})
```
kegsay, line 81: There are no guarantees that this token will be _after_ all the rooms have been created. You likely need a sentinel event here to know for sure that the previous requests have been processed.

kegsay (changes requested):

kegsay:
Marked as stale; this PR will be closed in 7 days if there isn't any further progress. Please re-open when you get time.

ShadowJonathan:
Sorry for ignoring this, I'll work on this later this week, thanks for reminding me 👍

ShadowJonathan:
Whoops, didn't intend for that to request review, sorry!

tests/csapi/apidoc_room_event_test.go
ShadowJonathan: I'm fairly sure it has little effect, yes, the only effect i can think of is that it will block until `roomAmount - goConcurrency` items have been processed (in `doneEvents`), through concurrency, so it makes the below bit where it awaits timeout a bit of a dud unless the last `goConcurrency` amount of items take that long.

In other words, changing this to `roomAmount` will make it detect timeouts earlier, but I'll apply the suggestion regardless. 👍

tests/csapi/apidoc_room_event_test.go
ShadowJonathan: This bit is also the exact invariant the original sytest test was testing; make sure events go down the correct room.

I'm fairly sure that this test will not fail on modern servers anyways, but I also have no compelling reason to remove it, as programmer mistakes can still exist, and this is essentially a canary for other problems.

tests/csapi/apidoc_room_event_test.go
ShadowJonathan: I'm assuming that a sync request like this, after a response from the room creation endpoint, will include those created rooms, is this a bad assumption?

https://github.com/matrix-org/matrix-spec/issues/938 is related to this issue, btw.

tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+	for i := 0; i < roomAmount; i++ {
+		todoRooms <- true
+	}
+
+	close(todoRooms)
+
+	var rooms []string
+
+	for i := 0; i < roomAmount; i++ {
```
kegsay, line 63: I don't understand why we need to do this all concurrently. We're making 10 rooms, not 1000. Why not just:
```go
var rooms []string
for i := 0; i < roomAmount; i++ {
    rooms = append(rooms, alice.CreateRoom(t, map[string]interface{}{
        "preset": "public_chat",
    })
}
```
That is 6 lines and clearer that what is here.

tests/csapi/apidoc_room_event_test.go
kegsay: In this current form it is probably fine to sync now, as you've waited for all /createRoom calls to 200 OK. To be sure, you could just hit /sync until you see all `roomAmount` rooms.

tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,198 @@
...
+	for i := 0; i < goConcurrency; i++ {
+		go func() {
+			for {
+				tid, ok := <-todoEvents
+
+				if !ok {
+					return
+				}
+
+				paths := []string{"_matrix", "client", "r0", "rooms", tid.roomId, "send", "m.room.message", strconv.Itoa(tid.txn)}
```
kegsay, line 94: Replace all of this with `alice.SendEventUnsyncedWithTxnID`.

tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,198 @@
...
+			txn:    i,
+		}
+	}
+
+	close(todoEvents)
+
+	for i := 0; i < roomAmount; i++ {
+		select {
+		case <-doneEvents:
+			break
```
kegsay, line 119: Why do you discard the event ID? Needs comments what you're trying to do here.

tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+				paths := []string{"_matrix", "client", "r0", "rooms", tid.roomId, "send", "m.room.message", strconv.Itoa(tid.txn)}
+				res := alice.MustDoFunc(t, "PUT", paths, client.WithJSONBody(t, map[string]interface{}{
+					"msgtype": "m.text",
+					"body":    tid.roomId,
+				}))
+				body := client.ParseJSON(t, res)
+				eventID := client.GetJSONFieldStr(t, body, "event_id")
+
+				doneEvents <- eventID
```
kegsay, line 102: This channel is never closed.

tests/csapi/apidoc_room_event_test.go
```diff
@@ -0,0 +1,198 @@
...
+	type txnAndId struct {
+		txn    int
+		roomId string
+	}
+
+	var doneEvents = make(chan string, roomAmount)
+	var todoEvents = make(chan txnAndId, roomAmount)
+
+	for i := 0; i < goConcurrency; i++ {
+		go func() {
```
kegsay, line 86: Again, I don't understand the concurrency usage here, it's more complex, more error prone for debatable gains.

kegsay (changes requested):
Remove the concurrency and use newer client methods, then LGTM.

Issue #258, opened 2021-12-02
Support sytests with deprecated endpoints?


ShadowJonathan:
While converting tests, I came upon the following ones;

```
./tests/80torture/03events.pl:test "GET /events with negative 'limit'",
./tests/80torture/03events.pl:test "GET /events with non-numeric 'limit'",
./tests/80torture/03events.pl:test "GET /events with non-numeric 'timeout'",
./tests/80torture/03events.pl:test "GET /initialSync with non-numeric 'limit'",
```

`/events` and `/initialSync` have long-been deprecated according to the spec, should I still be converting these? Or can i add some sort of "excluded" marker next to them? (and build that into the `sytest-coverage` tool)

kegsay:
No, don't test deprecated endpoints. If a sytest is _using_ a deprecated endpoint then that test should be updated. If a sytest is _testing_ a deprecated endpoint then it can be safely skipped.

ShadowJonathan:
What about sytest coverage? Should i edit it to add the exclusion bit?

kegsay:
Yeah, we shouldn't be taking deprecated tests into account when working out the % parity.

ShadowJonathan:
Alright, while i think it's easy to remove it from the list and PR that, i'll make a change to sytest-coverage to instead see if a line is suffixed with an `X`, and then exclude that line.

This way we can also track which tests were ignored explicitly, instead of having to dig through git history

PR #276, opened 2021-12-25
Cleanup `sytest.list`, add ignore list

ShadowJonathan:
Review note; completely ignore `sytest.list` except the first two paragraphs, which should be able to explain what's happening.

In this PR, i'm setting out a new way of tracking the sytest list. Whereas previously it was defined by a one-time dump, this'll keep track of the sytest commit it was generated from.

I'm also adding a way to start ignoring tests in a seperate list, sytest-coverage has been updated to reflect this.

---

The idea going forward is that, one in a while, we'd want to update the `sytest.list` file deterministically, i've embedded the linux command that makes this possible to be replicated with a sytest repo.

This PR, while updating the sytest list to the commit, also sees a "diff" of tests removed/added from sytest:

<details>
<summary>"Diff"</summary>

```py
# All tests that were added
{'A prev_batch token from incremental sync can be used in the v1 messages API',
 'Federation publicRoom Name/topic keys are correct',
 'Inbound /make_join rejects attempts to join rooms where all users have left',
 'Local users can peek by room alias',
 'Local users can peek into world_readable rooms by room ID',
 'Membership event with an invalid displayname in the send_join response should not cause room join to fail',
 "POST /register allows registration of usernames with '$chr'",
 'Peeked rooms only turn up in the sync for the device who peeked them',
 'Rejects invalid device keys',
 'Rooms with names are correctly named in pushes',
 'Tags appear in the v1 /initialSync',
 "We can't peek into rooms with $visibility history_visibility"}

# All tests that were removed
{'Device lists caches should be flushed when you re-encounter a user',
 'Device lists get refreshed when you encounter an unrecognised device',
 "If you send >20 device lists updates in a row, they don't get lost?",
 'Left room members do not cause problems for presence',
 'Outbound federation rejects /invite responses which are not correctly signed',
 'Rooms with names are correctly named in pushed',
 'Tags appear in the v1 /initalSync',
 "User signups are forbidden from starting with '_'",
 'When a room is upgraded to E2E, device lists caches should be flushed'}
```

From the removed tests, there's just one test mentioned in complement;

`User signups are forbidden from starting with '_'` is mentioned in `tests/csapi/apidoc_register_test.go` as a TODO, this can probably be ignored. (iirc, the test in question is also not mandated by the spec)

---

Some tests mentioned in https://github.com/matrix-org/complement/issues/258 were added to the ignore list, and one where the typo of the name got fixed between list updates


ShadowJonathan:
This PR adds a whole bunch of tests to ignore, but mainly I grepped and added the following ones;
- The ones mentioned in matrix-org/complement#258
- The ones mentioning /events or /initialSync directly
- The ones relating to groups/communities

cmd/sytest-coverage/main.go
```diff
@@ -100,6 +87,60 @@ func main() {
...
+	for _, line := range testLines {
+		name, filename := extract(line)
+		if name == "" || filename == "" {
+			continue
+		}
+		if _, ok := ignoredTests[name]; ok {
+			continue
+		}
+		for _, path := range ignoredPaths {
+			if strings.Contains(filename, path) {
```
kegsay, line 132: `Contains` probably isn't good enough in the general case as an ignored path of "foo" matches "food" which isn't what we want. Probably fine for now though.

kegsay (approved):
Good improvement, thank you!

Issue #278, opened 2021-12-26
`JSONKeyEqual` will give confusing error message when comparing JSON numbers


ShadowJonathan:
> MatchResponse key 'matrix:image:size' got '2239' want '2239'

This is when using `JSONKeyEqual` on a `match.JSON` array.

Changing the error message section to the following;

```go
  if !reflect.DeepEqual(gotValue, wantValue) {
    return fmt.Errorf("key '%s' got '%v' (%T) want '%v' (%T)", wantKey, gotValue, gotValue, wantValue, wantValue)
  }
```

Yields this more useful error message;

> MatchResponse key 'matrix:image:size' got '2239' (float64) want '2239' (int)

This behaviour should probably be accounted for when extracting values with `JSONKeyEqual`, so either `JSONKeyEqual` will have to coerce all numerical types into `float64`, or give this above more-clear error message.

kegsay:
Bleh, thanks for pointing this out. Agreed, test writers should not need to know or care about this.

ShadowJonathan:
I got tripped up by this again today:

```golang
		PLContent := map[string]interface{}{
			"invite": 100,
			"users": map[string]interface{}{
				alice.UserID: 100,
			},
		}

		eventId := alice.SendEventSynced(t, roomID, b.Event{
			Type:     "m.room.power_levels",
			StateKey: b.Ptr(""),
			Content:  PLContent,
		})

		res := alice.MustDoFunc(
			t,
			"GET",
			[]string{"_matrix", "client", "v3", "rooms", roomID, "event", eventId},
		)

		must.MatchResponse(t, res, match.HTTPResponse{
			JSON: []match.JSON{
				match.JSONKeyEqual("content", PLContent),
			},
		})
```

> `MatchResponse key 'content' got 'map[invite:100 users:map[@alice:hs1:100]]' want 'map[invite:100 users:map[@alice:hs1:100]]'`

This fails because the integer inside the powerlevel event's content is converted to float64, so doing a roundtrip is not possible.

ShadowJonathan:
One thing we can do is do a [canonicalisation roundtrip](https://blog.gojek.io/relooking-at-golangs-reflect-deepequal/) by `Marshal`ling and `Unmarshal` the compared-to `interface{}`, to make sure it matches the type quirks that reflect.DeepEquals needs.

PR #279, opened 2021-12-26
Add 12 remaining 51media sytests

ShadowJonathan:
This will add ~~all~~some remaining `51media` sytests, excluding the ones ignored in https://github.com/matrix-org/complement/pull/276;

```
./tests/51media/01unicode.pl
X Alternative server names do not cause a routing loop
  Can download specifying a different Unicode file name
  Can download with Unicode file name locally
  Can download with Unicode file name over federation
  Can upload with Unicode file name

./tests/51media/03ascii.pl
  Can download file '$filename'
  Can download specifying a different ASCII file name
  Can fetch images in room
  Can send image in room message
  Can upload with ASCII file name

./tests/51media/10thumbnail.pl
  POSTed media can be thumbnailed
  Remote media can be thumbnailed

./tests/51media/20urlpreview.pl
X Test URL preview

./tests/51media/30config.pl
  Can read configuration endpoint
```

Edit: Tests prefixed with `X` are removed after review.

This also adds a `SplitMxc` small helper function to `client`, for code dedup.


internal/data/data.go
```diff
@@ -0,0 +1,47 @@
+package data
+
+import "encoding/base64"
+
+const TestPngImageB64 = `
```
ShadowJonathan, line 5: `go embed` isn't available until 1.16, and the go.mod is at 1.13, I didn't want to bump (unless it is desirable to embed this blob from a seperate file)

tests/url_preview_test.go
```diff
@@ -0,0 +1,98 @@
...
+	var e = client.GjsonEscape
+
+	must.MatchResponse(t, res, match.HTTPResponse{
+		JSON: []match.JSON{
+			match.JSONKeyEqual(e("og:title"), oGraphTitle),
+			match.JSONKeyEqual(e("og:type"), oGraphType),
+			match.JSONKeyEqual(e("og:url"), oGraphUrl),
+			match.JSONKeyEqual(e("matrix:image:size"), 2239.0),
+			match.JSONKeyEqual(e("og:image:height"), 129.0),
+			match.JSONKeyEqual(e("og:image:width"), 279.0),
```
ShadowJonathan, line 84: See https://github.com/matrix-org/complement/issues/278, all JSON keys are returned floats, apparantly.

internal/validate/img.go
```diff
@@ -0,0 +1,24 @@
...
+	"fmt"
+)
+
+// TODO write more full PNG validator, this one is taken from sytest
+
+var pngMagic = []byte{0x89, 'P', 'N', 'G', 0x0D, 0x0A, 0x1A, 0x0A}
+var chunkOfIHDR = []byte{0, 0, 0, 0x0D, 'I', 'H', 'D', 'R'}
+
+// ValidatePng validates the bytes of a PNG blob
+func ValidatePng(png []byte) error {
```
kegsay, line 14: Just use https://pkg.go.dev/image/png#Decode ?

internal/data/data.go
kegsay: `go embed` would be preferable here, as it's much easier to check that the PNG is valid as you can just open it rather than just trust that this base-64 is a PNG. I'm happy for this to be bumped to 1.16

tests/media_filename_test.go
```diff
@@ -0,0 +1,152 @@
...
+				t.Parallel()
+
+				alice.UploadContent(t, data.TestPngImage, asciiFileName, "image/png")
+			})
+
+			// sytest: Can download file '$filename'
+			for _, filename := range []string{"ascii", "name with spaces", "name;with;semicolons"} {
+
+				// else linter will scream
+				var filename = filename
```
kegsay, line 39: This isn't a linter thing. Without this, the test function will be called 3 times with the last string in the array `"name;with;semicolons"` as you're just holding a reference to the variable which is overwritten on each loop.

tests/media_thumbnail_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+		if err != nil {
+			t.Fatalf("validating thumbnail png failed: %s", err)
+		}
+	} else {
+		// TODO: more mimetypes
+		t.Fatalf("Encountered unknown mimetype %s", mimeType)
+	}
+
+	// todo test for thumbnail size conformity?
```
kegsay, line 83: If you use https://pkg.go.dev/image/png you can check the dimensions of the image.

tests/url_preview_test.go
```diff
@@ -0,0 +1,108 @@
+package tests
```
kegsay, line 1: Please remove this file and the test within. Complement does not listen on `:8448` anymore so this won't work, and I'd want to have a standardised way of spinning up mini web-servers.

tests/media_routing_loop_test.go
```diff
@@ -0,0 +1,133 @@
+package tests
```
kegsay, line 1: Please remove this file and the test within. Complement does not listen on `:8448` anymore so this won't work, and I'd want to have a standardised way of spinning up mini web-servers.

kegsay (changes requested):
This needs a bit of work, but overall the code is excellent. The main points:
 - Remove the tests which spin up web servers.
 - Use `image/png` instead of a homebrew validator.
 - Upgrade complement to 1.16 and use the `embed` package.

tests/media_filename_test.go
ShadowJonathan: I know, the comment was partially in humor, but i'll edit it to explain why this line is here.

tests/url_preview_test.go
ShadowJonathan: What do you mean with "test within"?

And i'll try to figure out what a nice standardized way could be, but iirc this was the best way to do it when i wrote this, but that could've changed by now.

tests/media_routing_loop_test.go
ShadowJonathan: Ditto https://github.com/matrix-org/complement/pull/279#discussion_r833165265

tests/url_preview_test.go
kegsay: Basically remove the test and file and don't replace it with something else.

ShadowJonathan: Why exactly? Should "Test URL preview" not be tested? Or URL previews at all?

kegsay: They should be tested, but only once we have a standardised way to test with webservers like this.

ShadowJonathan: :+1: Alright, then i'll remove the tests related to URL previews, and make a note to add in that standardized way first, before i'll re-add them.

kegsay (approved):
Looks amazing, thank you!

PR #280, opened 2022-01-05
Add /event_auth test

kegsay:
Requires https://github.com/matrix-org/gomatrixserverlib/pull/283 to land first before it'll pass on Synapse. Broken on Dendrite.


tests/federation_event_auth_test.go
```diff
@@ -0,0 +1,87 @@
...
+	// now update the auth chain a bit: dendrite had a bug where it returned the auth chain for all
+	// the current state in addition to the event asked for
+	alice.SendEventSynced(t, roomID, b.Event{
+		Type:     "m.room.join_rules",
+		StateKey: b.Ptr(""),
+		Content: map[string]interface{}{
+			"join_rule": "invite",
+		},
+	})
+	waiter.Wait(t, 1*time.Second) // wait for the join rule to make it to the complement server
```
neilalexander, line 61: Does this have the potential to race?

tests/federation_event_auth_test.go
```diff
@@ -0,0 +1,87 @@
...
+	if len(resp.AuthEvents) != len(joinRuleEvent.AuthEvents()) {
+		msg := "got:\n"
+		for _, e := range resp.AuthEvents {
+			msg += string(e.JSON()) + "\n\n"
+		}
+		t.Fatalf("got %d auth events, wanted %d.\n%s\nwant: %s", len(resp.AuthEvents), len(joinRuleEvent.AuthEvents()), msg, joinRuleEvent.AuthEventIDs())
+	}
+	// make sure all the events match
+	wantIDs := map[string]bool{}
+	for _, id := range joinRuleEvent.AuthEventIDs() {
```
neilalexander, line 78: Seeing as the spec calls for the endpoint to return the entire auth chain recursively, it would be nice if this test exercised that. A simple way might be for `charlie` to leave and re-join the room and check that the auth events of both the leave and the second join also appear.

neilalexander (approved):
As a simple test this is fine, wondering if we might want to put a bit more into checking the auth chain after state transitions though. (Or if we should do that in a separate test?)

tests/federation_event_auth_test.go
kegsay: No, this is the point of using the Waiter. It will wait up to 1s for the event to arrive. The waiter stops waiting when the event arrives (see further up)

neilalexander: Ah, OK. I misunderstood what was happening there.

neilalexander (approved):

Issue #60, opened 2021-01-18
Rename 'CI' environment variable to 'COMPLEMENT_IN_DOCKER' or similar


anoadragon453:
This grew out of a discussion [here](https://github.com/matrix-org/pipelines/pull/114#discussion_r551842653), where it was found that the `CI` environment variable is used by Complement to know whether it is running inside of a docker container already (and thus needs to talk to the host's docker socket) or is running on the host directly. It's assumed that if Complement is running inside of docker, `CI=true`.

@richvdh proposed to rename this variable to `COMPLEMENT_IN_DOCKER` (or something similar) instead, as it's more descriptive of what the variable actually specifies. Plus people won't be tempted to add CI-specific things outside of what's necessary (and thus making CI code split further from production code) :innocent:.

Kegsay:
I chose `CI=true` because of https://docs.gitlab.com/ee/ci/variables/predefined_variables.html - basically Gitlab sets this flag automatically for their gitlab runners. Similarly, Travis CI does the same - https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

richvdh:
right, but the feature you're enabling here isn't "running under CI", it's "running under docker". 

 * there's no reason why I shouldn't do this stuff locally as well as under CI.
 * there's nothing to say that if I'm running under CI, I'm inherently running inside a docker container. Sure that's the approach we take on our BK setup but other CI envs might solve the problem completely differently.
 * calling it `CI` tells me, as a complement user absolutely nothing about what difference it makes, making it a scary black box.
 * calling it `CI` means that I might be tempted to make complement do other things differently under CI than in a normal run, which is generally a bad smell.

All in all: I think you chose ... poorly :)

Kegsay:
I agree with your concerns and I'm not opposed to updating the name, but this was my thinking at the time.

PR #289, opened 2022-01-21
Don't listen on :8448 for the Complement federation server

kegsay:
Instead, listen on a random OS-allocated high numbered port then
do a switcheroo on the `ServerName` so it reads correctly e.g
`host.docker.internal:56185`. This means the server name will be
invalid if it is read before `Server.Listen()` is called so we now
guard common access points in `Server` which rely on the server
name and fail tests if the server is not yet listening when those
functions are called. To further guard against misuse of server
name whilst it isn't valid, turn it into a private field.

As of this PR, it will now (in theory) be possible to scale out tests indefinitely as we no longer have any central point/resource which all tests rely on. This should make it possible to split `/tests` into more packages which make sense from a Matrix PoV.

Indirectly fixes https://github.com/matrix-org/complement/issues/60

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -32,9 +32,9 @@ jobs:
             default_branch: master
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2 # Checkout complement
 
-      # Env vars are set file a file given by $GITHUB_PATH
+      # Env vars are set file a file given by $GITHUB_PATH. We need both Go 1.17 and GOPATH on env.
       # See https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-system-path
       - name: "Set Go Version"
         run: |
@@ -49,9 +49,8 @@ jobs:
         # We don't need to install Go because it is included on the Ubuntu 20.04 image:
         # See https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md specifically GOROOT_1_17_X64
         run: |
-          sudo apt-get update && sudo apt-get install -y libolm3
+          sudo apt-get update && sudo apt-get install -y libolm3 libolm-dev
           go get -v github.com/haveyoudebuggedit/gotestfmt/v2/cmd/gotestfmt@latest
-
 
       - name: "Checkout corresponding ${{ matrix.homeserver }} branch"
         # This is only done for Synapse since Dendrite's docker file pulls in
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,7 +41,12 @@ jobs:
       # servers which listen on random high numbered ports.
       - name: "Install Complement Dependencies"
         run: |
-          sudo apt-get update && sudo apt-get install -y libolm3 golang-go
+          sudo apt-get update && sudo apt-get install -y libolm3
+          sudo apt-get remove golang-go
+          wget https://dl.google.com/go/go1.16.4.linux-amd64.tar.gz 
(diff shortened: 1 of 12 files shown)


neilalexander (approved):

.github/workflows/ci.yaml
```diff
@@ -82,7 +88,7 @@ jobs:
         working-directory: homeserver
 
       - run: docker build -t homeserver -f dockerfiles/${{ matrix.homeserver }}.Dockerfile dockerfiles/
-      - run: set -o pipefail && go test -p 2 -v -json -tags "${{ matrix.tags }}" ./tests/... 2>&1 | gotestfmt
+      - run: export PATH=$PATH:/usr/local/go/bin && set -o pipefail && cd complement-master && go test -p 2 -v -json -tags "${{ matrix.tags }}" ./tests/... 2>&1 | gotestfmt
```
richvdh, line 91: consider splitting this onto separate lines for (much) easier readability

```suggestion
      - run: |
          export PATH=$PATH:/usr/local/go/bin
          set -o pipefail
          cd complement-master
          go test -p 2 -v -json -tags "${{ matrix.tags }}" ./tests/... 2>&1 | gotestfmt
```

internal/federation/server.go
```diff
@@ -63,7 +64,7 @@ func NewServer(t *testing.T, deployment *docker.Deployment, opts ...func(*Server
 		Priv:                        priv,
 		KeyID:                       "ed25519:complement",
 		mux:                         mux.NewRouter(),
-		ServerName:                  docker.HostnameRunningComplement,
+		serverName:                  docker.HostnameRunningComplement,
```
richvdh, line 67: I'd suggest adding a comment here to note how/when it gets updated.

internal/federation/handle.go
```diff
@@ -18,7 +18,7 @@ import (
 func MakeJoinRequestsHandler(s *Server, w http.ResponseWriter, req *http.Request) {
 	// Check federation signature
 	fedReq, errResp := gomatrixserverlib.VerifyHTTPRequest(
-		req, time.Now(), gomatrixserverlib.ServerName(s.ServerName), s.keyRing,
+		req, time.Now(), gomatrixserverlib.ServerName(s.serverName), s.keyRing,
```
richvdh, line 21: I haven't gone through these in detail, but it seems problematic that they are changing to `serverName` rather than `ServerName()`.

AIUI, the difference is that `serverName` works before we start listening, but *may return an incomplete server name*. If we pass the wrong server name in here, aren't we checking that the request is signed for the wrong server?

internal/federation/server.go
```diff
@@ -114,16 +115,29 @@ func NewServer(t *testing.T, deployment *docker.Deployment, opts ...func(*Server
 	return srv
 }
 
+func (s *Server) ServerName() string {
```
richvdh, line 118: a doc-comment never goes amiss.

internal/federation/handle.go
kegsay: No because this is called when the server is already running, hence `s.serverName` is already set to the right value.

richvdh: In which case, why not call `ServerName()`, to include the safety check?

kegsay: Because there is no need for the safety check. The server _must_ be listening at this point in order for this code to be hit.

richvdh: for now, sure. But this is about robustness, and failing in a clear way when an assumption is violated.

Anyway, I don't feel that strongly. If you don't think the check is worthwhile, feel free to skip it. I'm not going to go through each of these changes to check that .listen() must have happened first though, so I'll have to take your word for it.

PR #301, opened 2022-02-08
Add `TestUnrejectRejectedEvents`

neilalexander:
This test is designed to ensure that we unreject a previously rejected event if we now have all of the prerequisites satisfied for it to pass on a second processing attempt. It is designed to ensure we don't regress, as we've been making changes around making sure Dendrite does the right thing in these cases (like matrix-org/dendrite#2159).

Since federation `/event` etc can probably return rejected events, the best way I can see to test this is to ensure that it only appears in sync for the first time *after* the prerequisite event.


kegsay (approved):
Excellent and super clear, thank you!

PR #308, opened 2022-02-11
Update GHA to allow for non-`master` default branch

richvdh:


richvdh:
dunno why CI failed, but I think it's unrelated

kegsay (approved):

PR #316, opened 2022-02-15
Fix flakey room forget tests

S7evinK:
Somehow chaining `.Get` seems to be the problem, not the escape part (I added it anyway).

### DIFF
--- a/tests/csapi/apidoc_room_forget_test.go
+++ b/tests/csapi/apidoc_room_forget_test.go
@@ -4,7 +4,6 @@ import (
 	"net/http"
 	"net/url"
 	"testing"
-	"time"
 
 	"github.com/tidwall/gjson"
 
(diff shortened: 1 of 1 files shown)


tests/csapi/apidoc_room_forget_test.go
```diff
@@ -73,19 +72,18 @@ func TestRoomForget(t *testing.T) {
 			alice.MustDoFunc(t, "POST", []string{"_matrix", "client", "r0", "rooms", roomID, "forget"})
 			time.Sleep(50 * time.Millisecond)
```
kegsay, line 73: Please remove this. You can probably use a sentinel event to catch this. Do the following:
- Alice forgets the room
- Bob sends a message in roomID
- Sync until you see the message on Bob's sync stream
- Then sync Alice and check that the room isn't there.

This is called the sentinel method because you're using a separate event to act as a sentinel to detect when something previous has happened. This only reliably works when there is a single queue handling sync data, but for the case of Dendrite this should work.

kegsay (changes requested):
LGTM otherwise.

kegsay (approved):

PR #319, opened 2022-02-16
Have synapse log to stdout

richvdh:
... which means that `docker logs` also logs to stdout, which is much easier to
tesl with.


DMRobertson (approved):
Where does this logging go without this change---stderr? The howling void?

richvdh:
stderr. See https://docs.python.org/3/library/logging.handlers.html#logging.StreamHandler (and also https://docs.python.org/3/library/logging.config.html#access-to-external-objects for the `ext://` syntax)

PR #321, opened 2022-02-17
Support registering as an admin via shared secrets 

kegsay:
Requires homeservers to use 'complement' as the shared secret. This is very similar to sytest which also uses this endpoint to register admins so it can test admin endpoints: https://github.com/matrix-org/sytest/blob/develop/tests/10apidoc/01register.pl#L304

Manually tested on dendrite/synapse and it works, assuming dendrite is running https://github.com/matrix-org/dendrite/pull/2194

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ If you're looking to run against a custom Dockerfile, it must meet the following
 - The homeserver needs to manage its own storage within the image.
 - The homeserver needs to accept the server name given by the environment variable `SERVER_NAME` at runtime.
 - The homeserver needs to assume dockerfile `CMD` or `ENTRYPOINT` instructions will be run multiple times.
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -76,6 +76,7 @@ If you're looking to run against a custom Dockerfile, it must meet the following
 - The homeserver needs to manage its own storage within the image.
 - The homeserver needs to accept the server name given by the environment variable `SERVER_NAME` at runtime.
 - The homeserver needs to assume dockerfile `CMD` or `ENTRYPOINT` instructions will be run multiple times.
+- The homeserver nees to use `complement` as the registration shared secret for `/_synapse/admin/v1/register`, if supported. If this endpoint 404s then these tests are skipped.
```
neilalexander, line 79: needs

S7evinK (approved):

neilalexander (approved):
I'm guessing we need an accompanying change to Dendrite's `generate-config` for the shared secret?

PR #322, opened 2022-02-18
Add tests for server notices

S7evinK:
Tests for https://github.com/matrix-org/dendrite/pull/2180

### DIFF
--- a/tests/csapi/admin_test.go
+++ b/tests/csapi/admin_test.go
@@ -84,10 +84,10 @@ func TestServerNotices(t *testing.T) {
 	t.Run("Alice can leave the alert room, after joining it", func(t *testing.T) {
 		alice.LeaveRoom(t, roomID)
 	})
-	t.Run("After leaving the alert room, a new room is created", func(t *testing.T) {
+	t.Run("After leaving the alert room and on re-invitation, no new room is created", func(t *testing.T) {
 		sendServerNotice(t, admin, reqBody, nil)
 		newRoomID := syncUntilInvite(t, alice)
-		if roomID == newRoomID {
+		if roomID != newRoomID {
 			t.Errorf("expected a new room, but they are the same")
 		}
 	})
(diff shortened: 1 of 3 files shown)


tests/csapi/admin_test.go
```diff
@@ -13,3 +21,111 @@ func TestCanRegisterAdmin(t *testing.T) {
...
+	body := must.MatchResponse(t, res, match.HTTPResponse{
+		StatusCode: http.StatusOK,
+		JSON: []match.JSON{
+			match.JSONKeyPresent("event_id"),
+		},
+	})
+	return gjson.GetBytes(body, "event_id").Str
+}
+
+func syncUntilInvite(t *testing.T, alice *client.CSAPI) string {
```
kegsay, line 121: Needs a comment explaining why you can't use `client.SyncInvitedTo` - you don't know the room ID!

tests/csapi/admin_test.go
```diff
@@ -13,3 +21,111 @@ func TestCanRegisterAdmin(t *testing.T) {
...
+	t.Run("/send_server_notice as an admin is allowed", func(t *testing.T) {
+		sendServerNotice(t, admin, reqBody, nil)
+
+	})
+	t.Run("Alice is invited to the server alert room", func(t *testing.T) {
+		sendServerNotice(t, admin, reqBody, nil)
+		syncUntilInvite(t, alice)
+	})
+	t.Run("Alice cannot reject the invite", func(t *testing.T) {
+		sendServerNotice(t, admin, reqBody, nil)
```
kegsay, line 57: Why are we constantly sending server notices? These subtests will run sequentially.

kegsay (approved):
Excellent battery of tests!

tests/csapi/admin_test.go
S7evinK: I was running them in parallel first, but remembered that this might race since there can only be one active notice room. Fixed.

kegsay (approved):

PR #328, opened 2022-03-01
Update complement to use setuptools entrypoints

DMRobertson:
Crossref: matrix-org/synapse#12118

I have made this script a proper entry point, so it appears as an suffixless executable.


DMRobertson:
This has been tested on https://github.com/matrix-org/synapse/pull/12118 and approved. I'm going to merge this now so that Synapse's CI will succeed on develop.

PR #331, opened 2022-03-04
Remove knocking build flag and Synapse's experimental config option

anoadragon453:
As knocking has been stable for a while now (and is even in [a released version of the spec](https://spec.matrix.org/v1.2/client-server-api/#knocking-on-rooms)), it is now certainly overdue to remove the build flag guarding Complement's knocking tests.

cc @matrix-org/dendrite-core

Recommended to review commit-by-commit.


tests/knocking_test.go
```diff
@@ -1,7 +1,9 @@
-// +build msc2403
+// Knocking is not yet implemented on Dendrite
+//go:build !dendrite_blacklist
```
anoadragon453, line 2: This line was automatically added by my IDE. I assume it's correct to include?

kegsay: That's fine.

anoadragon453:
I believe this is just waiting for an approval on your side @kegsay?

anoadragon453:
Oh -- let's request review from them then.

clokep (approved):

PR #337, opened 2022-03-09
WIP: Attempt to get poetry working

DMRobertson:
See https://github.com/matrix-org/synapse/pull/12385

### DIFF
--- a/dockerfiles/Synapse.Dockerfile
+++ b/dockerfiles/Synapse.Dockerfile
@@ -24,7 +24,7 @@ COPY keys/* /ca/
 RUN openssl genrsa -out /conf/server.tls.key 2048
 
 # generate a signing key
-RUN generate_signing_key -o /conf/server.signing.key
+RUN /synapse/.venv/bin/generate_signing_key -o /conf/server.signing.key
 
 WORKDIR /data
 
--- a/dockerfiles/synapse/start.sh
+++ b/dockerfiles/synapse/start.sh
@@ -24,5 +24,4 @@ openssl x509 -req -in /conf/server.tls.csr \
   -CA /ca/ca.crt -CAkey /ca/ca.key -set_serial 1 \
   -out /conf/server.tls.crt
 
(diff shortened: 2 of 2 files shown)


dockerfiles/Synapse.Dockerfile
```diff
@@ -24,7 +24,7 @@ COPY keys/* /ca/
 RUN openssl genrsa -out /conf/server.tls.key 2048
 
 # generate a signing key
-RUN generate_signing_key -o /conf/server.signing.key
+RUN /synapse/.venv/bin/generate_signing_key -o /conf/server.signing.key
```
richvdh, line 27: can we arrange for `/synapse/.venv/bin` to be on the default PATH, so we don't need to worry about this?

dockerfiles/synapse/start.sh
```diff
@@ -24,5 +24,4 @@ openssl x509 -req -in /conf/server.tls.csr \
   -CA /ca/ca.crt -CAkey /ca/ca.key -set_serial 1 \
   -out /conf/server.tls.crt
 
-exec python -m synapse.app.homeserver -c /conf/homeserver.yaml "$@"
-
+exec /synapse/.venv/bin/python -m synapse.app.homeserver -c /conf/homeserver.yaml "$@"
```
richvdh, line 27: similarly, change this to:

```suggestion
exec synapse_homeserver -c /conf/homeserver.yaml "$@"
```

... and then it becomes independent of the poetry change. It's a good thing to do anyway, if we're now telling people to use the entrypoint, per https://github.com/matrix-org/synapse/pull/12337/files#diff-7b3ed02bc73dc06b7db906cf97aa91dec2b2eb21f2d92bc5caa761df5bbc168fR323.

DMRobertson: I'll do this separately.

DMRobertson:
As of https://github.com/matrix-org/synapse/pull/12385 we should once again be able to use the entrypoints without having to specify a full path.

PR #339, opened 2022-03-10
Add a method to login

S7evinK:
Adds a `Login` and `MustLogin` on the client, for easier logging in.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
 
 	"github.com/matrix-org/complement/internal/b"
 	"github.com/matrix-org/complement/internal/must"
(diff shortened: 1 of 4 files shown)


internal/client/client.go
```diff
@@ -321,6 +322,66 @@ func (c *CSAPI) RegisterUser(t *testing.T, localpart, password string) (userID,
...
+
+// Login tries to log in with the given password, returns the *http.Response for further inspection.
+func (c *CSAPI) Login(t *testing.T, userID, password string, opts ...LoginOpt) *http.Response {
+	reqBody := json.RawMessage(fmt.Sprintf(`{
+				"type": "m.login.password",
+				"identifier": {
+					"type": "m.id.user",
+					"user": "%s"
+				},
+				"password": "%s"
```
kegsay, line 358: This won't work if the password contains `"`.

kegsay (changes requested):
I don't want this. Logging into an account should not be necessary for the majority of tests because they should be pulling alice/bob/etc from blueprints. The only time this is actually needed are _for tests which check login_ and where it is actually relevant for the test e.g all of `apidoc_login_test.go` and changing password tests.

By exposing `MustLogin` and `Login` on the client we are suggesting to test writers that this is a-okay to do and should be done for their tests, which is _not_ the intention. Furthermore, the login API is more complicated that just user/pass (e.g token login) and these helper functions don't seem to account for that.

PR #343, opened 2022-03-14
Attempt to deflake `TestRoomCreationReportsEventsToMyself`

S7evinK:
We can't run `/sync` in parallel, as this can result in `Room creation reports m.room.member to myself` consuming events required for `Room creation reports m.room.create to myself` and thus failing the test.


S7evinK:
As I've learned, they can be run in parallel.

PR #353, opened 2022-03-29
Convert more tests to Complement

S7evinK:
Again adds new tests:
`tests/32room-versions.pl` - this seems to slow down the tests quite a bit (at least locally)
`tests/10apidoc/36room-levels.pl`
`tests/30rooms/01state.pl`


tests/csapi/apidoc_room_levels_test.go
```diff
@@ -0,0 +1,105 @@
...
+func TestRoomLevels(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+
+	successCount := 0
+	defer func() {
+		// sytest: Both GET and PUT work
+		if successCount != 2 {
+			t.Fatalf("expected GET and PUT to work")
```
kegsay, line 25: Why is this necessary? `t.Run("Parallel", ...` will block until the tests complete. Each test can fail the subtest which will have a knock-on effect and fail the outer test, so this seems entirely redundant? We never do this anywhere in Complement AFAICT.

tests/csapi/apidoc_room_levels_test.go
```diff
@@ -0,0 +1,105 @@
...
+		// sytest: PUT /rooms/:room_id/state/m.room.power_levels can set levels
+		t.Run("PUT /rooms/:room_id/state/m.room.power_levels can set levels", func(t *testing.T) {
+			t.Parallel()
+			roomID := alice.CreateRoom(t, map[string]interface{}{})
+			alice.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(alice.UserID, roomID))
+			res := alice.MustDoFunc(t, "GET", []string{"_matrix", "client", "v3", "rooms", roomID, "state", "m.room.power_levels"})
+
+			powerLevels := gjson.ParseBytes(must.ParseJSON(t, res.Body))
+			changedUser := client.GjsonEscape("@random-other-user:their.home")
+			alicePowerLevel := powerLevels.Get("users." + alice.UserID).Int()
```
kegsay, line 65: Technically `alice.UserID` should be gjson escaped as well.

tests/csapi/apidoc_room_levels_test.go
```diff
@@ -0,0 +1,105 @@
...
+				"users": map[string]int64{
+					alice.UserID: 100,
+				},
+			})
+			alice.MustDoFunc(t, "PUT", []string{"_matrix", "client", "v3", "rooms", roomID, "state", "m.room.power_levels"}, reqBody)
+			// absence of a 'users' key
+			reqBody = client.WithJSONBody(t, map[string]interface{}{})
+			alice.MustDoFunc(t, "PUT", []string{"_matrix", "client", "v3", "rooms", roomID, "state", "m.room.power_levels"}, reqBody)
+			// this should now give a 403 (not a 500)
+			res := alice.DoFunc(t, "PUT", []string{"_matrix", "client", "v3", "rooms", roomID, "state", "m.room.power_levels"}, reqBody)
```
kegsay, line 99: This isn't what the sytest is doing.
```perl
      # absence of an 'events' key
      matrix_put_room_state(
         $user,
         $room_id,
         type      => "m.room.power_levels",
         state_key => "",
         content   => {
            users => {
               $user->user_id => 100,
            },
         },
      )->then( sub {
         # absence of a 'users' key
         matrix_put_room_state(
            $user,
            $room_id,
            type      => "m.room.power_levels",
            state_key => "",
            content   => {
            },
         );
      })->then( sub {
         # this should now give a 403 (not a 500)
         matrix_put_room_state(
            $user,
            $room_id,
            type      => "m.room.power_levels",
            state_key => "",
            content   => {
               users => {},
            },
         ) -> main::expect_http_403;
      })
```

So the order is:
- PUT with users key with alice
- PUT with empty content
- PUT with users key present but empty -> 403

tests/csapi/room_versions_test.go
```diff
@@ -0,0 +1,193 @@
...
+	return cl.SendEventSynced(t, roomID, b.Event{
+		Type: "m.room.message",
+		Content: map[string]interface{}{
+			"msgtype": "m.text",
+			"body":    "hello world",
+		},
+	})
+}
+
+func joinRoomSynced(t *testing.T, cl *client.CSAPI, roomID, alias string) {
```
kegsay, line 179: These should probably go into the Client impl to be honest. For now leave them here though.

tests/csapi/room_versions_test.go
```diff
@@ -0,0 +1,193 @@
...
+		// iterate over all room versions
+		for v := range roomVersions {
+			roomVersion := v
+			// sytest: User can create and send/receive messages in a room with version $version
+			t.Run(fmt.Sprintf("User can create and send/receive messages in a room with version %s", roomVersion), func(t *testing.T) {
+				t.Parallel()
+				roomID := createRoomSynced(t, alice, map[string]interface{}{
+					"room_version": roomVersion,
+				})
+				alice.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(alice.UserID, roomID))
```
kegsay, line 36: Isn't the whole point of the `createRoomSynced` function to literally do this? Why do it again?

tests/csapi/room_versions_test.go
```diff
@@ -0,0 +1,193 @@
...
+				sendMessageSynced(t, alice, roomID)
+			})
+
+			userTypes := map[string]*client.CSAPI{
+				"local":  bob,
+				"remote": charlie,
+			}
+			for typ, joiner := range userTypes {
+				typ := typ
+				joiner := joiner
```
kegsay, line 51: Need comments to explain why you do this (else it'll take the last value only).

tests/csapi/room_versions_test.go
```diff
@@ -0,0 +1,193 @@
...
+			// sytest: Can reject invites over federation for rooms with version $version
+			t.Run(fmt.Sprintf("Can reject invites over federation for rooms with version %s", roomVersion), func(t *testing.T) {
+				t.Parallel()
+				roomID := createRoomSynced(t, alice, map[string]interface{}{
+					"room_version": roomVersion,
+					"invite":       []string{charlie.UserID},
+				})
+				charlie.MustSyncUntil(t, client.SyncReq{}, client.SyncInvitedTo(charlie.UserID, roomID))
+				charlie.LeaveRoom(t, roomID)
+			})
```
kegsay, line 131: No check for alice to see that the invite was rejected?

kegsay (changes requested):

kegsay:
Closing as stale.

Issue #290, opened 2022-01-21
Swap out golangci-lint GHA for manual one


kegsay:
We don't see line numbers / files on errors which is utterly useless. https://github.com/golangci/golangci-lint-action/issues/119

kegsay:
Also, we should make `camel_case_vars` produce an error. xref https://github.com/matrix-org/complement/pull/354#discussion_r840792976

PR #354, opened 2022-03-30
Add test for non-member state events in `/batch_send` `state_events_at_start`

MadLittleMods:
Add test for non-member state events in `/batch_send` `state_events_at_start`

Tests for https://github.com/matrix-org/synapse/issues/12110

Synapse changes: https://github.com/matrix-org/synapse/pull/12329

### DIFF
--- a/tests/msc2716_test.go
+++ b/tests/msc2716_test.go
@@ -1382,6 +1382,23 @@ func validateBatchSendRes(t *testing.T, c *client.CSAPI, roomID string, batchSen
 	expectedEventIDOrder = append(expectedEventIDOrder, reversed(historicalEventIDs)...)
 	expectedEventIDOrder = append(expectedEventIDOrder, insertionEventID)
 
+	// Make sure the historical events appear in scrollback without jumping back
+	// in time specifically.
+	fullMessagesRes := c.MustDoFunc(t, "GET", []string{"_matrix", "client", "r0", "rooms", roomID, "messages"}, client.WithContentType("application/json"), client.WithQueries(url.Values{
+		"dir":   []string{"b"},
+		"limit": []string{"100"},
+	}))
+	must.MatchResponse(t, fullMessagesRes, match.HTTPResponse{
+		JSON: []match.JSON{
+			matcherJSONEventIDArrayInOrder("chunk",
+				expectedEventIDOrder,
+				historicalEventFilter,
+			),
+		},
+	})
+
+	// Validate state after we paginate `/messages` to avoid any potential 404 if
+	// the server hasn't backfilled here yet
 	if validateState {
 		contextRes := c.MustDoFunc(t, "GET", []string{"_matrix", "client", "r0", "rooms", roomID, "context", expectedEventIDOrder[0]}, client.WithContentType("application/json"), client.WithQueries(url.Values{
 			"limit": []string{"0"},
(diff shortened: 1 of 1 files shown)


tests/msc2716_test.go
```diff
@@ -1342,36 +1383,12 @@ func validateBatchSendRes(t *testing.T, c *client.CSAPI, roomID string, batchSen
...
-		messagesRes := c.MustDoFunc(t, "GET", []string{"_matrix", "client", "r0", "rooms", roomID, "messages"}, client.WithContentType("application/json"), client.WithQueries(url.Values{
-			// Go backwards (newest -> oldest) (same direction as if you were using scrollback)
-			"dir": []string{"b"},
-			// From the newest-in-time event in the historical batch
-			"from": []string{batchStartPaginationToken},
-			// We are aiming to scrollback to the oldest-in-time event from the
-			// historical batch
-			"limit": []string{fmt.Sprintf("%d", len(expectedEventIDOrder))},
-			// We add these options to the filter so we get member events in the state field
-			"filter": []string{"{\"lazy_load_members\":true,\"include_redundant_members\":true}"},
```
MadLittleMods, line 1365: `/messages` was only returning member state events and only because we passed this special filter. Now that we're testing with non-member state events, we need an endpoint which will show us them.

We can instead simplify and just use `/context` directly 🎉

tests/msc2716_test.go
```diff
@@ -305,6 +305,47 @@ func TestImportHistoricalMessages(t *testing.T) {
...
+
+			roomID := as.CreateRoom(t, createPublicRoomOpts)
+			alice.JoinRoom(t, roomID, nil)
+
+			// Create the "live" event we are going to insert our historical events next to
+			eventIDsBefore := createMessagesInRoom(t, alice, roomID, 1, "eventIDsBefore")
+			eventIdBefore := eventIDsBefore[0]
+			timeAfterEventBefore := time.Now()
+
+			state_events_at_start := createJoinStateEventsForBatchSendRequest([]string{virtualUserID}, timeAfterEventBefore)
```
kegsay, line 319: `camelCase` please.

tests/msc2716_test.go
kegsay: Bear in mind that `/context` does not work over federation. That is, if server A is in room X and sends 100 messages, then server B joins room X and hits `/context` for an event early in the timeline (so not backfilled) it will 404.

kegsay (approved):

tests/msc2716_test.go
MadLittleMods: Is there a linter to use to avoid these kinds of mistakes?

tests/msc2716_test.go
MadLittleMods: I moved the `validateState` block (which calls `/context`) after we call `/messages` which will backfill any of the potential gaps 👍

tests/msc2716_test.go
kegsay: There is, but I don't think it's built into CI. Added it to https://github.com/matrix-org/complement/issues/290

kegsay (approved):
Merge whenever you want!

MadLittleMods:
Thanks for the review @kegsay 🐬

Was waiting for https://github.com/matrix-org/synapse/pull/12329 to merge 🚢

PR #361, opened 2022-04-11
Log the body of unexpected HTTP requests

richvdh:
I've observed intermittent test failures due to `/send` requests arriving when
none are expected. To help narrow down what is going on, let's log the content
of any unexpected requests.

### DIFF
--- a/internal/federation/server.go
+++ b/internal/federation/server.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
+	"io/ioutil"
 	"math/big"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


internal/federation/server.go
```diff
@@ -94,7 +95,8 @@ func NewServer(t *testing.T, deployment *docker.Deployment, opts ...func(*Server
 	})
 	srv.mux.NotFoundHandler = http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		if srv.UnexpectedRequestsAreErrors {
-			t.Errorf("Server.UnexpectedRequestsAreErrors=true received unexpected request to server: %s %s", req.Method, req.URL.Path)
+			body, _ := ioutil.ReadAll(req.Body)
+			t.Errorf("Server.UnexpectedRequestsAreErrors=true received unexpected request to server: %s %s\n%s", req.Method, req.URL.Path, string(body))
```
DMRobertson, line 99: I'm assuming `string(io.ReadCloser)` is safe and isn't going to fall over with e.g. a sequence of bytes that's not e.g. a valid UTF-8 byte stream?

DMRobertson: (I did try looking at the go docs but couldn't see where `string(io.ReadCloser)` would be described or implemented)

DMRobertson (approved):

internal/federation/server.go
richvdh: (given we're failing the test anyway at this point, I don't think it matters too much.)

PR #364, opened 2022-04-18
Port "registration accepts non-ascii passwords" from sytest

H-Shay:
As the title states. Adds "registration accepts non-ascii passwords" to Complement from the sytest file `11register.pl`.


H-Shay:
I may be missing something, but AFICT the dendrite failures are unrelated to the test that I added. That test appears to be passing, and the failing tests seem to be unrelated to the change I made.

tests/csapi/registration_test.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+func Test11Register(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	unauthedClient := deployment.Client(t, "hs1", "")
+	// sytest: registration accepts non-ascii passwords
+	t.Run("Registration accepts non-ascii passwords", func(t *testing.T) {
+		reqJson := json.RawMessage(`{"username": "test_user", "password": "übers3kr1t", "device_id": "xyzzy", "initial_device_display_name": "display_name"}`)
+		resp := unauthedClient.DoFunc(t, "POST", []string{"_matrix", "client", "r0", "register"}, client.WithRawBody(reqJson))
```
kegsay, line 22: Prefer using `map[string]interface{}` over raw JSON bodies as that will handle escaping for you. You do this further down with `reqBody` and `client.WithJSONBody`

tests/csapi/registration_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"testing"
+
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
+	"github.com/tidwall/gjson"
+)
+
+func Test11Register(t *testing.T) {
```
kegsay, line 15: Please don't include 11 in this. Complement doesn't use the numbering scheme sytest does. We already have some registration tests so put this code as a subtest of the existing function please.

kegsay (approved):
Otherwise LGTM

PR #368, opened 2022-05-06
Add regression test when modifying events_default

kegsay:
xref https://github.com/matrix-org/gomatrixserverlib/pull/306

### DIFF
--- a/tests/csapi/power_levels_test.go
+++ b/tests/csapi/power_levels_test.go
@@ -18,7 +18,7 @@ func TestDemotingUsersViaUsersDefault(t *testing.T) {
 	roomID := alice.CreateRoom(t, map[string]interface{}{
 		"preset": "public_chat",
 		"power_level_content_override": map[string]interface{}{
-			"users_default": 100, // the default is 100
+			"users_default": 100, // the default is 0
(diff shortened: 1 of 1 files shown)


tests/csapi/power_levels_test.go
```diff
@@ -0,0 +1,42 @@
...
+func TestDemotingUsersViaUsersDefault(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+
+	roomID := alice.CreateRoom(t, map[string]interface{}{
+		"preset": "public_chat",
+		"power_level_content_override": map[string]interface{}{
+			"users_default": 100, // the default is 100
```
neilalexander, line 21: The comment should probably be `// the default is 0`.

neilalexander (approved):

tests/csapi/power_levels_test.go
kegsay, line 21:
```suggestion
			"users_default": 100, // the default is 0
```

PR #373, opened 2022-05-18
More tests for partial-state joins

richvdh:
Adds some more testcases for faster joins - in particular, that we can (initial) `/sync` during a faster join, and that `/members` blocks until the resync completes.

Depends on https://github.com/matrix-org/synapse/pull/12498.

Review commit-by-commit: it has a couple of refactoring commits.


squahtx (approved):
lgtm

PR #375, opened 2022-05-24
Add key upload tests

S7evinK:
Again, the commit history should give an overview of what is added.
This also updates `maunium.net/go/mautrix` to fix an index out of range issue when generating keys.


tests/federation_upload_keys_test.go
```diff
@@ -0,0 +1,140 @@
...
+			must.MatchResponse(t, resp, match.HTTPResponse{
+				StatusCode: http.StatusOK,
+				JSON: []match.JSON{
+					match.JSONKeyMissing("one_time_keys." + client.GjsonEscape(alice.UserID)),
+				},
+			})
+		})
+
+		// sytest: Can query remote device keys using POST
+		t.Run("Can query remote device keys using POST", func(t *testing.T) {
```
kegsay, line 70: This isn't the same as the sytest, which does stuff with displaynames.

tests/csapi/upload_keys_test.go
```diff
@@ -0,0 +1,208 @@
...
+		t.Run("Can upload device keys", func(t *testing.T) {
+			reqBody := client.WithJSONBody(t, map[string]interface{}{
+				"device_keys":   deviceKeys,
+				"one_time_keys": oneTimeKeys,
+			})
+			resp := alice.MustDoFunc(t, "POST", []string{"_matrix", "client", "v3", "keys", "upload"}, reqBody)
+			must.MatchResponse(t, resp, match.HTTPResponse{
+				StatusCode: http.StatusOK,
+				JSON: []match.JSON{
+					match.JSONKeyEqual("one_time_key_counts.signed_curve25519", float64(1)),
```
kegsay, line 36: It's unclear why this should be `signed_curve25519`, and in this way the sytest is clearer. Perhaps instead loop the `oneTimeKeys` map.

tests/csapi/upload_keys_test.go
```diff
@@ -0,0 +1,208 @@
...
+				},
+			})
+		})
+
+		// sytest: Rejects invalid device keys
+		t.Run("Rejects invalid device keys", func(t *testing.T) {
```
kegsay, line 42: Please please please keep useful sytest comments when porting things over. You don't explain in this test how this request is invalid :/ 

The sytest is clearer: `# algorithms, keys and signatures are required fields, but missing`

kegsay (approved):

kegsay:
Which defaults to false - https://github.com/matrix-org/synapse/blob/116a4c8340b729ffde43be33df24d417384cb28b/synapse/config/federation.py#L48

PR #378, opened 2022-05-24
Test homeserver restarts during a partial state join

squahtx:
Tests https://github.com/matrix-org/synapse/issues/12642


internal/docker/deployer.go
```diff
@@ -212,12 +219,14 @@ func deployImage(
...
-					HostIP: "127.0.0.1",
+					HostIP:   "127.0.0.1",
+					HostPort: strconv.Itoa(port1),
 				},
 			},
 			nat.Port("8448/tcp"): []nat.PortBinding{
 				{
-					HostIP: "127.0.0.1",
+					HostIP:   "127.0.0.1",
+					HostPort: strconv.Itoa(port2),
```
squahtx, line 229: We must provide an explicit port, otherwise we'll get a different random port when the container is restarted. Which breaks everything.

tests/federation_room_join_partial_state_test.go
```diff
@@ -182,6 +182,62 @@ func TestPartialStateJoin(t *testing.T) {
...
+		// attempts to sync should block. Fire off a goroutine to try it.
+		syncResponseChan := make(chan gjson.Result)
+		defer close(syncResponseChan)
+		go func() {
+			response, _ := alice.MustSync(t, client.SyncReq{})
+			syncResponseChan <- response
+		}()
+
+		// wait for the state_ids request to arrive
+		psjResult.AwaitStateIdsRequest(t)
```
squahtx, line 216: `AwaitStateIdsRequest` cannot be used twice.

erikjohnston (approved):

tests/federation_room_join_partial_state_test.go
squahtx: I'm tempted to drop the second usage rather than try to rejig things.

internal/docker/deployer.go
```diff
@@ -391,6 +348,39 @@ func deployImage(
 	return d, nil
 }
 
+// Picks two free ports on localhost. Does not reserve them in any way.
+// The returned ports must be used before the next call to `allocateHostPorts`,
+// otherwise the same pair of ports may be returned.
```
kegsay, line 353: We cannot guarantee that as homeservers are deployed in parallel.

kegsay (changes requested):
My main problem with this lies with the usage of an explicit port. We cannot guarantee this as:
 - Tests in the same file can run in parallel via `.Parallel()`
 - Tests in different directories can run in parallel (modifiable via `go test -p N`)

We need to preserve that ability to ensure we can grow the test case library without suffering increasingly long test times, so we cannot add features which would compromise that. I've asked for a bit more information than just "which breaks everything".

internal/docker/deployer.go
```diff
@@ -417,6 +407,99 @@ func copyToContainer(docker *client.Client, containerID, path string, data []byt
...
+
+	for _, hsDep := range dep.HS {
+		err := dep.Deployer.Docker.ContainerStop(ctx, hsDep.ContainerID, &dep.Config.SpawnHSTimeout)
+		if err != nil {
+			return fmt.Errorf("failed to restart container %s: %s", hsDep.ContainerID, err)
+		}
+
+		// Remove the container from the network. If we don't do this,
+		// (re)starting the container fails with an error like
+		// "Error response from daemon: endpoint with name complement_fed_1_fed.alice.hs1_1 already exists in network complement_fed_alice".
```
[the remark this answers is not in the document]
kegsay: It doesn't seem like it's possible, so if this works then great, let's do that for now.

internal/docker/deployer.go
squahtx: `BaseURL` and `FedBaseURL` of `HomeserverDeployment` become incorrect because they include the port number. We can fix up those URLs. But any `CSAPI` client instances will continue to use the previous port number and that's not so easy to fix up.

internal/docker/deployer.go
squahtx: That's a good point, I didn't realise that we ran tests in parallel at the time.

internal/docker/deployer.go
kegsay: That seems like a reasonably easy thing to fix? There's several possible solutions:
 - Add documentation to `Restart()` which explicitly states that any clients created with this deployment need to be recreated. Or...
 - Add a callback hook to `client.CSAPI` to update the base URL and get test authors to call that callback with the updated URL. Or...
 - Remember all `client.CSAPI` instances created via `func (d *Deployment) Client(t *testing.T, hsName, userID string) *client.CSAPI` and automatically call the callback hook when the deployment is restarted to re-point port numbers.

The last option is the most preferable because:
 - All the complexity is in one place and does not need to be repeated in tests.
 - Test authors don't need to know or care that the port number changes from under them when they call `deployment.Restart()`.

In addition, I would change the function signature of restart from:
```go
func (dep *Deployment) Restart() error
```
to
```go
func (dep *Deployment) Restart(t *testing.T)
```
which will fail the test if the restart fails (returns an error).

At present, this PR cannot be accepted because it will introduce flakiness because these port numbers will race: we have to let the underlying OS decide the ports.

squahtx: I've split the PR and made the changes in #396. Let me know if I've missed anything.

PR #379, opened 2022-05-25
Test re-syncing state from other servers listed in `/send_join` reply

squahtx:
Tests https://github.com/matrix-org/synapse/pull/12812

### DIFF
--- a/internal/federation/handle.go
+++ b/internal/federation/handle.go
@@ -129,11 +129,11 @@ func SendJoinRequestsHandler(s *Server, w http.ResponseWriter, req *http.Request
 
 	authEvents := room.AuthChainForEvents(stateEvents)
 
+	// get servers in room *before* the join event
+	serversInRoom := room.ServersInRoom()
+
 	// insert the join event into the room state
 	room.AddEvent(event)
-
-	// servers in room: just us. TODO(faster_joins): this may not be correct
-	serversInRoom := []string{s.serverName}
 
 	// return state and auth chain
 	b, err := json.Marshal(gomatrixserverlib.RespSendJoin{
--- a/internal/federation/handle.go
+++ b/internal/federation/handle.go
@@ -45,6 +45,7 @@ func MakeJoinRequestsHandler(s *Server, w http.ResponseWriter, req *http.Request
 		Type:       "m.room.member",
 		StateKey:   &userID,
 		PrevEvents: []string{room.Timeline[len(room.Timeline)-1].EventID()},
+		Depth:      room.Timeline[len(room.Timeline)-1].Depth() + 1,
 	}
 	err := builder.SetContent(map[string]interface{}{"membership": gomatrixserverlib.Join})
 	if err != nil {
--- a/internal/federation/handle.go
+++ b/internal/federation/handle.go
@@ -45,7 +45,6 @@ func MakeJoinRequestsHandler(s *Server, w http.ResponseWriter, req *http.Request
 		Type:       "m.room.member",
 		StateKey:   &userID,
 		PrevEvents: []string{room.Timeline[len(room.Timeline)-1].EventID()},
-		Depth:      room.Timeline[len(room.Timeline)-1].Depth() + 1,
 	}
 	err := builder.SetContent(map[string]interface{}{"membership": gomatrixserverlib.Join})
 	if err != nil {
--- a/internal/federation/server.go
+++ b/internal/federation/server.go
@@ -119,6 +119,11 @@ func NewServer(t *testing.T, deployment *docker.Deployment, opts ...func(*Server
 	return srv
 }
 
+// Room returns the given room
+func (s *Server) Room(roomID string) *ServerRoom {
+	return s.rooms[roomID]
+}
+
 // Return the server name of this federation server. Only valid AFTER calling Listen() - doing so
 // before will produce an error.
 //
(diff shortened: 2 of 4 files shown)


internal/federation/handle.go
```diff
@@ -45,6 +45,7 @@ func MakeJoinRequestsHandler(s *Server, w http.ResponseWriter, req *http.Request
 		Type:       "m.room.member",
 		StateKey:   &userID,
 		PrevEvents: []string{room.Timeline[len(room.Timeline)-1].EventID()},
+		Depth:      room.Timeline[len(room.Timeline)-1].Depth() + 1,
```
squahtx, line 48: This is a drive-by fix and doesn't affect the outcome of the test.

internal/federation/server_room.go
```diff
@@ -97,24 +97,38 @@ func (r *ServerRoom) AuthChain() (chain []*gomatrixserverlib.Event) {
...
 	}
-	// find them in the timeline
-	for _, tev := range r.Timeline {
-		if chainMap[tev.EventID()] {
-			chain = append(chain, tev)
-		}
-	}
+
 	return
 }
```
squahtx, line 133: Without this change, the complement homeserver excludes hs1's membership event from the auth chain sent to hs2. hs2 is unhappy and tries to make `/event_auth` requests, which the complement homeserver is not set up to handle.

internal/federation/handle.go
kegsay: Please don't mix this with tests where possible. The set of reviewers for $feature are different from the set of reviewers for handling Complement internals.

internal/federation/server.go
```diff
@@ -119,6 +119,11 @@ func NewServer(t *testing.T, deployment *docker.Deployment, opts ...func(*Server
 	return srv
 }
 
+// Room returns the given room
+func (s *Server) Room(roomID string) *ServerRoom {
```
kegsay, line 123: This is not required. You call `MustJoinRoom` in the test which returns the `*ServerRoom` already.

tests/federation_room_join_partial_state_test.go
```diff
@@ -182,6 +182,104 @@ func TestPartialStateJoin(t *testing.T) {
...
+				},
+				// hs1 may send us presence when alice syncs
+				func(e gomatrixserverlib.EDU) {},
+			),
+		)
+		cancelListener := server.Listen()
+		defer cancelListener()
+
+		// join complement to the public room
+		server.MustJoinRoom(t, deployment, "hs1", roomID, server.UserID("bob"))
```
kegsay, line 217: This returns a `*ServerRoom` so you don't need to then retrieve it with the new `server.Room(roomID)` function on `:245`.

tests/federation_room_join_partial_state_test.go
```diff
@@ -182,6 +182,104 @@ func TestPartialStateJoin(t *testing.T) {
...
+
+		// the client-side requests should still be waiting
+		select {
+		case <-syncResponseChan:
+			t.Fatalf("hs2 sync completed before state resync complete")
+		default:
+		}
+
+		// reply to hs2 with a bogus /state_ids response
+		sendResponseWaiter.Finish()
```
kegsay, line 268: This could have a clearer name, e.g `sendStateIDsResponse`, as otherwise it's unclear which response you mean and need to add comments to explain it.

kegsay (approved):
Overall this is an excellent PR! Thanks!

tests/federation_room_join_partial_state_test.go
squahtx: I'll rename them to follow the naming in `partialStateJoinResult`:
```go
type partialStateJoinResult struct {
	cancelListener                   func()
	Server                           *federation.Server
	ServerRoom                       *federation.ServerRoom
	fedStateIdsRequestReceivedWaiter *Waiter
	fedStateIdsSendResponseWaiter    *Waiter
}
```

squahtx:
@kegsay Thanks for reviewing. Could you re-review the latest changes?

tests/federation_room_join_partial_state_test.go
```diff
@@ -182,6 +182,103 @@ func TestPartialStateJoin(t *testing.T) {
...
+		server := federation.NewServer(t, deployment,
+			federation.HandleKeyRequests(),
+			federation.HandlePartialStateMakeSendJoinRequests(),
+			federation.HandleEventRequests(),
+			federation.HandleTransactionRequests(
+				func(e *gomatrixserverlib.Event) {
+					t.Fatalf("Received unexpected PDU: %s", string(e.JSON()))
+				},
+				// hs1 may send us presence when alice syncs
+				func(e gomatrixserverlib.EDU) {},
```
richvdh, line 210: I think you can just write `nil` if there's nothing to do for EDUs.
```suggestion
				nil,
```

tests/federation_room_join_partial_state_test.go
```diff
@@ -182,6 +182,103 @@ func TestPartialStateJoin(t *testing.T) {
...
+			),
+		)
+		cancelListener := server.Listen()
+		defer cancelListener()
+
+		// join complement to the public room
+		room := server.MustJoinRoom(t, deployment, "hs1", roomID, server.UserID("bob"))
```
richvdh, line 217: worth noting there is also a `bob@hs1` created by the blueprint. It might be better to pick a fourth username.

richvdh (approved):
lgtm other than a few tweaks

tests/federation_room_join_partial_state_test.go
squahtx: I went for `david` instead.

richvdh (approved):
lgtm, and I need to build stuff on top of it, so I'm going to go ahead and merge it.

PR #380, opened 2022-05-27
Add sync timeline tests

S7evinK:
^ Title
Also adds the possibility to use a custom transaction ID when sending events.

### DIFF
--- a/internal/b/blueprints.go
+++ b/internal/b/blueprints.go
@@ -91,10 +91,11 @@ type ApplicationService struct {
 }
 
 type Event struct {
-	Type     string
-	Sender   string
-	StateKey *string
-	Content  map[string]interface{}
+	Type          string
+	Sender        string
+	StateKey      *string
+	Content       map[string]interface{}
+	TransactionID *string
 
 	/* The following fields are ignored in blueprints as clients are unable to set them.
 	 * They are used with federation.Server.
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -169,6 +169,10 @@ func (c *CSAPI) SendEventSynced(t *testing.T, roomID string, e b.Event) string {
 	t.Helper()
 	c.txnID++
 	paths := []string{"_matrix", "client", "r0", "rooms", roomID, "send", e.Type, strconv.Itoa(c.txnID)}
+	if e.TransactionID != nil {
+		c.txnID--
+		paths = []string{"_matrix", "client", "r0", "rooms", roomID, "send", e.Type, *e.TransactionID}
+	}
 	if e.StateKey != nil {
 		paths = []string{"_matrix", "client", "r0", "rooms", roomID, "state", e.Type, *e.StateKey}
 	}
--- a/tests/csapi/sync_timeline_test.go
+++ b/tests/csapi/sync_timeline_test.go
@@ -1,0 +1,365 @@
+package csapi_tests
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"reflect"
+	"testing"
+
+	"github.com/tidwall/gjson"
+
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
+)
+
+func TestSyncTimeline(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+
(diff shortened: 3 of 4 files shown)


internal/client/client.go
```diff
@@ -169,6 +169,10 @@ func (c *CSAPI) SendEventSynced(t *testing.T, roomID string, e b.Event) string {
 	t.Helper()
 	c.txnID++
 	paths := []string{"_matrix", "client", "r0", "rooms", roomID, "send", e.Type, strconv.Itoa(c.txnID)}
+	if e.TransactionID != nil {
+		c.txnID--
```
kegsay, line 173: We make no guarantees of always sending sequential txn IDs, so no need to `--` this.

internal/b/blueprints.go
```diff
@@ -91,10 +91,11 @@ type ApplicationService struct {
...
 type Event struct {
-	Type     string
-	Sender   string
-	StateKey *string
-	Content  map[string]interface{}
+	Type          string
+	Sender        string
+	StateKey      *string
+	Content       map[string]interface{}
+	TransactionID *string
```
kegsay, line 98: I dislike this being here because it falls into the same area as the fields below:

> The following fields are ignored in blueprints as clients are unable to set them.

Really we should have an embedded struct to cleanly separate `b.Event` (i.e blueprint events) and client event e.g `client.Event` - the former has very few customisable fields due to it being run via the instruction runner, but the latter has many more fields to adjust.

tests/csapi/sync_timeline_test.go
```diff
@@ -0,0 +1,365 @@
...
+						"limit": 2,
+					},
+				},
+			})
+			if err != nil {
+				t.Fatalf("unable to marshal filter: %v", err)
+			}
+
+			filterID := createFilter(t, alice, reqBody, alice.UserID)
+			roomID := alice.CreateRoom(t, struct{}{})
```
kegsay, line 53: Never do this as Synapse and Dendrite have different defaults when there is no preset. Always say if you want this to be public/private.

tests/csapi/sync_timeline_test.go
```diff
@@ -0,0 +1,365 @@
...
+			if !timeline.Get("events").Exists() || !timeline.Get("limited").Exists() || !timeline.Get("prev_batch").Exists() {
+				t.Fatalf("missing required field in timeline")
+			}
+			events := timeline.Get("events").Array()
+			wantEventCount := 2
+			assertEqual(t, len(events), wantEventCount)
+			assertEqual(t, events[0].Get("event_id").Str, eventIDs[0])
+			assertEqual(t, events[0].Get("content.body").Str, "Test message 0")
+			assertEqual(t, events[1].Get("event_id").Str, eventIDs[1])
+			assertEqual(t, events[1].Get("content.body").Str, "Test message 1")
```
kegsay, line 70: Alternatively just use `must.EqualStr(t, a, b)` https://github.com/matrix-org/complement/blob/main/internal/must/must.go#L136

tests/csapi/sync_timeline_test.go
```diff
@@ -0,0 +1,365 @@
...
+				t.Fatalf("unable to marshal filter: %v", err)
+			}
+
+			filterID := createFilter(t, alice, reqBody, alice.UserID)
+			roomID := alice.CreateRoom(t, struct{}{})
+			eventIDs := sendMessages(t, alice, roomID, "m.room.message", "Test message ", 2)
+			res, _ := alice.MustSync(t, client.SyncReq{Filter: filterID})
+			room := res.Get("rooms.join." + client.GjsonEscape(roomID))
+			if !room.Get("timeline").Exists() || !room.Get("state").Exists() || !room.Get("ephemeral").Exists() {
+				t.Fatalf("missing required field in response")
```
kegsay, line 58: Why aren't you using JSON matchers?
```go
roomKey := "rooms.join." + client.GjsonEscape(roomID)
must.MatchResponse(t, res match.HTTPResponse{
    JSON: []match.JSON{
        match.JSONKeyPresent(roomKey + ".timeline"),
        match.JSONKeyPresent(roomKey + ".state"),
        match.JSONKeyPresent(roomKey + ".ephemeral"),
    }
})
```
You get much better and clearer error messages than just a vague "missing required field in response" which doesn't even say which field...

kegsay (changes requested):
I think you forgot the `must` and `match` packages exist.

tests/csapi/sync_timeline_test.go
S7evinK: `MustSync` returns a `gjson.Result` not `*http.Response` which is what `must.MatchResponse` expects. Otherwise I'd have used that. :)

kegsay:
Marked as stale; this PR will be closed in 7 days if there isn't any further progress. Please re-open when you get time.

Issue #6, opened 2020-06-28
Allow os.Setenv to pass through to docker containers


kegsay:
So authors can do stuff like DENDRITE_TRACE_INTERNAL=1

reivilibre:
I'm not sure I understand the title of this issue (`os.Setenv` seems to be a function for setting environment vars?).

However I would also like the ability to pass environment variables through to Synapse's container.
Do we just pass all the variables in, or should we be a bit more cautious (a typical process/shell usually has a lot of ambient environment variables; I don't want to interfere with the reproducibility too much)?
As a first thought: if we care about that, they could be prefixed with a common prefix — not sure what your preference is there?

PR #382, opened 2022-05-27
Enable arbitrary env vars to be propagated to Complement containers

kegsay:
Fixes https://github.com/matrix-org/complement/issues/6

This PR adds a new complement env var called `COMPLEMENT_SHARE_ENV_PREFIX`.
This can be set to any string e.g `FOO_`. If it is set, Complement will
look at the host environment for any env var with this prefix. If it finds
it, it will remove the prefix and add it to the Complement container.

For example, with a host env:
```
COMPLEMENT_SHARE_ENV_PREFIX=FOO_
FOO_DENDRITE_TRACE_HTTP=1
```
This will cause `DENDRITE_TRACE_HTTP=1` to be set in the Complement container.

This addresses 2 desires:
 - The ability to ad-hoc set debugging env vars when a test is failing
   e.g to use `os.Setenv("FOO_DENDRITE_TRACE_HTTP", "1")` in a test to get more logging.
 - The ability to pass through arbitrary data to a Synapse container, as requested
   by @reivilibre

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -22,14 +22,15 @@ type HostMount struct {
 }
 
 type Complement struct {
-	BaseImageURI          string
-	BaseImageArgs         []string
-	DebugLoggingEnabled   bool
-	AlwaysPrintServerLogs bool
-	BestEffort            bool
-	SpawnHSTimeout        time.Duration
-	KeepBlueprints        []string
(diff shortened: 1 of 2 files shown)


reivilibre (approved):
LGTM — lovely thank you :)

PR #383, opened 2022-05-27
Set the depth of join events returned from `/make_join`

squahtx:


internal/federation/handle.go
```diff
@@ -45,6 +45,7 @@ func MakeJoinRequestsHandler(s *Server, w http.ResponseWriter, req *http.Request
 		Type:       "m.room.member",
 		StateKey:   &userID,
 		PrevEvents: []string{room.Timeline[len(room.Timeline)-1].EventID()},
+		Depth:      room.Timeline[len(room.Timeline)-1].Depth() + 1,
```
DMRobertson, line 48: Strictly speaking this is supposed to saturate at an upper limit. The value of that upper limit is a [point of contention](https://github.com/matrix-org/matrix-spec/issues/1085).

(This is almost certainly fine as it is though.)

kegsay (approved):

PR #384, opened 2022-05-27
Update the README for local running.

tommie:
By mounting /root/.cache/go-build, we don't need to pre-cache
binaries, and it also helps when large changes are being made to
Dendrite. Without this, the compile cache only remembers the changes
when the ComplementLocal.Dockerfile was last built.

The example is now easier to get started with.

Also fixes a small lint issue for Go 1.16 (which is what Docker images use.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,20 +88,24 @@ If you're looking to run against a custom Dockerfile, it must meet the following
 
 If you want to write Complement tests _and_ hack on a homeserver implementation at the same time it can be very awkward
 to have to `docker build` the image all the time. To resolve this, Complement support "host mounts" which mount a directory
-from the host to the container. This is set via `COMPLEMENT_HOST_MOUNTS`:
+from the host to the container. This is set via `COMPLEMENT_HOST_MOUNTS`, on the form `HOST:CONTAINER[:ro][;...]` where
+`:ro` makes the mount read-only.
 
+For example, for Dendrite on Linux with the default location of `$GOPATH`, do a one-time setup:
+
+```shellsession
+$ git clone https://github.com/matrix-org/dendrite ../dendrite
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -88,21 +88,25 @@ If you're looking to run against a custom Dockerfile, it must meet the following
...
-from the host to the container. This is set via `COMPLEMENT_HOST_MOUNTS`:
+from the host to the container. This is set via `COMPLEMENT_HOST_MOUNTS`, on the form `HOST:CONTAINER[:ro][;...]` where
+`:ro` makes the mount read-only.
 
+For example, for Dendrite on Linux with the default location of `$GOPATH`, do a one-time setup:
+
+```shellsession
+$ git clone https://github.com/matrix-org/dendrite ../dendrite
+$ (cd ../dendrite && docker build -t complement-dendrite-local -f build/scripts/ComplementLocal.Dockerfile .)
+$ mkdir -p /tmp/complement-$LOGNAME/go-build
```
kegsay, line 99: Why are we using `$LOGNAME`? Can we simplify this so we don't need it please?

kegsay (approved):
So my understanding is that by adding this mount directory we don't need to constantly re-snapshot `complement-dendrite-local` as the changes stack up to improve compile times, because this directory includes the build outputs? I did not know this and that is awesome if so :D

Minor comment but otherwise LGTM

tommie:
> So my understanding is that by adding this mount directory we don't need to constantly re-snapshot `complement-dendrite-local` as the changes stack up to improve compile times, because this directory includes the build outputs? I did not know this and that is awesome if so :D

Yes, exactly. :) AFAIK, it contains both intermediate and final build output, so it's a big win to have persistent. I've made the change to remove pre-building from Dockerfile, and that's what I've been using to test SSO.

PR #387, opened 2022-05-31
Add logging to SendFederationRequest

DMRobertson:
This mimics the logging that's already present in the
`loggedRoundTripper.RoundTrip` for the client-server client.

An example from a WIP test (see the line starting `server.go:247`)

```
=== RUN   TestCreateEventCannotHaveParentEvents
    auth_rules_test.go:46: Deploy times: 4.719938126s blueprints, 3.007339507s containers
    client.go:549: GET hs1/_matrix/client/r0/capabilities => 200 OK (2.694078ms)
    server.go:165: Creating room !0:host.docker.internal:37287 with version 9
    client.go:549: POST hs1/_matrix/client/r0/join/!0:host.docker.internal:37287 => 200 OK (395.481311ms)
    server.go:247: GET hs1/_matrix/federation/v1/event/%24r3RILL05BCgmsoUBGmERO7iHrJBzeD4zemPrTenSDe4 => 2xx (no error) (147.16758ms)
    server.go:103: Server.UnexpectedRequestsAreErrors=false received unexpected request to server: PUT /_matrix/federation/v1/send/1654014160986 - sending 404 which may cause the HS to backoff from Complement
2022/05/31 17:22:43 ============================================
```


internal/federation/server.go
```diff
@@ -224,7 +231,20 @@ func (s *Server) SendFederationRequest(deployment *docker.Deployment, req gomatr
...
-	return httpClient.DoRequestAndParseResponse(context.Background(), httpReq, resBody)
+
+	start := time.Now()
+	err = httpClient.DoRequestAndParseResponse(context.Background(), httpReq, resBody)
+
+	var outcome string
+	if respError, ok := err.(gomatrix.RespError); ok {
+		outcome = respError.ErrCode
+	} else if httpError, ok := err.(gomatrix.HTTPError); ok {
+		outcome = strconv.Itoa(httpError.Code)
```
kegsay, line 242: The status code alone isn't enough. You need to print the HTTP response body. Print `.Contents` - https://github.com/matrix-org/gomatrixserverlib/blob/f72a635100606c4e772a60d9028e9fe1c98850e3/client.go#L539

internal/federation/server.go
```diff
@@ -224,7 +231,20 @@ func (s *Server) SendFederationRequest(deployment *docker.Deployment, req gomatr
 	}
 
 	httpClient := gomatrixserverlib.NewClient(gomatrixserverlib.WithTransport(&docker.RoundTripper{Deployment: deployment}))
-	return httpClient.DoRequestAndParseResponse(context.Background(), httpReq, resBody)
+
+	start := time.Now()
+	err = httpClient.DoRequestAndParseResponse(context.Background(), httpReq, resBody)
+
+	var outcome string
+	if respError, ok := err.(gomatrix.RespError); ok {
```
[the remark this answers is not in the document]
DMRobertson: Oh I see, `RespError` is the struct representing a Matrix error json blob. (I was just assuming that it was somehow a peer of HTTPError.

internal/federation/server.go
DMRobertson: (happy to be persuaded otherwise though.)

DMRobertson:
Sorry for the delay on this. I've lost a lot of the context, but if I was motivated enough to propose the change I may as well see it through.

internal/federation/server.go
kegsay: > it's the test's responsibility to print out more detailed errors

This is a great way to get no detailed errors, because people either forget, don't know how, or can't spend the time to add such detail. Complement is designed to give **good** errors by default, and detailed errors with `COMPLEMENT_DEBUG=1` set. Please ensure the response body is logged if `COMPLEMENT_DEBUG=1` is set.

kegsay (approved):
This is an improvement even if we can do even better on logging.

PR #388, opened 2022-06-06
Document use with `ufw`

### LINKED ISSUE #370: Document that using a firewall can cause tests to fail in non-obvious ways mentioning `host.docker.internal` (and how to work around)
I've been using the `ufw` firewall on my machine 'for security purposes'. Turns out this was blocking connections from the containers to `host.docker.internal` and causing some Synapse tests to fail.

Frustratingly, Synapse's error messages were pretty hopeless at giving any details (`ConnectingCancelledError: HostnameAddress(hostname=b'host.docker.internal', port=33145)` is not particularly insightful :/).

Nevertheless, I imagine this is not an uncommon setup and so it would be nice to bring attention to this as part of the instructions for running Complement?

I didn't know what `host.docker.internal` is; I now see that it that it should resolve to the address of the host (although it didn't resolve in an ad-hoc container, so I'm not sure).

I'll take a crack at this; ideally once I've found the right rules to work around it rather than disabling the whole thing.

richvdh:
Fixes #370.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,21 @@ options will work, so to just run 1 named test and include a timeout for the tes
 $ COMPLEMENT_BASE_IMAGE=complement-dendrite:latest go test -timeout 30s -run '^(TestOutboundFederationSend)$' -v ./tests/...
 ```
 
+### Potential conflict with firewall software
+
+The homeserver in the test image needs to be able to make requests to the mock
+homeserver hosted by Complement itself, which may be blocked by firewall
+software. This will manifest with a subset of the tests (mostly those to do
+with federation) inexplicably failing.
+
+To solve this, you will need to configure your firewall to allow such requests.
+
+If you are using [ufw](https://code.launchpad.net/ufw), this can be done with:
+
+```sh
+sudo ufw allow in on in_br-+
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,6 +36,21 @@ options will work, so to just run 1 named test and include a timeout for the tes
...
+homeserver hosted by Complement itself, which may be blocked by firewall
+software. This will manifest with a subset of the tests (mostly those to do
+with federation) inexplicably failing.
+
+To solve this, you will need to configure your firewall to allow such requests.
+
+If you are using [ufw](https://code.launchpad.net/ufw), this can be done with:
+
+```sh
+sudo ufw allow in on in_br-+
```
richvdh, line 51: worth mentioning here: docker, as used by complement, uses a bridge interface with a name like `br-54a9b2364d63` rather than the regular `docker0`, because we tell it to create a dedicated network with `docker network create` (or the equivalent).

squahtx, line 51: typo?
```suggestion
sudo ufw allow in on br-+
```

squahtx (approved):
Tested it and it makes complement work for me (after the typo fix)

PR #389, opened 2022-06-06
Use `host-gateway` to ensure servers can reach host

DMRobertson:
This fixes [connectivity problems I was having when running complement](https://github.com/matrix-org/complement/issues/370#issuecomment-1147687179). Oddly, noone else seems to have found this issue, and I don't really understand the underlying machinations.

I'm not sure if this will cause complement to fail in environments whose docker versions are older than 20.10.

### DIFF
--- a/internal/docker/deployer.go
+++ b/internal/docker/deployer.go
@@ -177,10 +177,11 @@ func deployImage(
 	var err error
 
 	if runtime.GOOS == "linux" {
-		// By default docker for linux does not expose this, so do it now.
-		// When https://github.com/moby/moby/pull/40007 lands in Docker 20, we should
-		// change this to be  `host.docker.internal:host-gateway`
-		extraHosts = []string{HostnameRunningComplement + ":172.17.0.1"}
+		// Ensure that the homservers under test can contact the host, so they can
+		// interact with a complement-controlled test server.
+		// Note: this feature of docker landed in Docker 20.10,
+		// see https://github.com/moby/moby/pull/40007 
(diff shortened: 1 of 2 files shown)


erikjohnston:
> I'm not sure if this will cause complement to fail in environments whose docker versions are older than 20.10.

Don't suppose there is an easy way to test the docker version?

internal/docker/deployer.go
```diff
@@ -177,10 +177,11 @@ func deployImage(
 	var err error
 
 	if runtime.GOOS == "linux" {
-		// By default docker for linux does not expose this, so do it now.
-		// When https://github.com/moby/moby/pull/40007 lands in Docker 20, we should
-		// change this to be  `host.docker.internal:host-gateway`
-		extraHosts = []string{HostnameRunningComplement + ":172.17.0.1"}
+		// Ensure that the homservers under test can contact the host, so they can
```
reivilibre, line 180:
```suggestion
		// Ensure that the homeservers under test can contact the host, so they can
```

DMRobertson:
> Don't suppose there is an easy way to test the docker version?

Erik means: can we interrogate the docker version at runtime?

It looks like this is exposed as `docker.dockerversion.Version`. Presumably this is some string that needs to be parsed. https://pkg.go.dev/github.com/docker/docker@v20.10.17+incompatible/dockerversion

kegsay (commented):
Please add docs to the README to say you need Docker 20+, then LGTM

PR #392, opened 2022-06-10
Gracefully shut down container when we want logs

erikjohnston:
This gives the container a chance to flush out any pending logs, otherwise we can end up with truncated logs.

### DIFF
--- a/internal/docker/deployer.go
+++ b/internal/docker/deployer.go
@@ -153,8 +153,8 @@ func (d *Deployer) Destroy(dep *Deployment, printServerLogs bool) {
 		if printServerLogs {
 			// If we want the logs we gracefully stop the containers to allow
 			// the logs to be flushed.
-			x := 1 * time.Second
-			err := d.Docker.ContainerStop(context.Background(), hsDep.ContainerID, &x)
+			timeout := 1 * time.Second
+			err := d.Docker.ContainerStop(context.Background(), hsDep.ContainerID, &timeout)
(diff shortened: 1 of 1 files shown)


internal/docker/deployer.go
```diff
@@ -151,13 +151,23 @@ func (d *Deployer) Deploy(ctx context.Context, blueprintName string) (*Deploymen
 func (d *Deployer) Destroy(dep *Deployment, printServerLogs bool) {
 	for _, hsDep := range dep.HS {
 		if printServerLogs {
+			// If we want the logs we gracefully stop the containers to allow
+			// the logs to be flushed.
+			x := 1 * time.Second
+			err := d.Docker.ContainerStop(context.Background(), hsDep.ContainerID, &x)
```
squahtx:
```suggestion
			timeout := 1 * time.Second
			err := d.Docker.ContainerStop(context.Background(), hsDep.ContainerID, &timeout)
```

squahtx (approved):
1 suggestion, otherwise looks good to me!


I wonder if homeservers might sometimes take more than a second to shut down. Synapse certainly does for production instances.

erikjohnston:
> 1 suggestion, otherwise looks good to me!
> 
> I wonder if homeservers might sometimes take more than a second to shut down. Synapse certainly does for production instances.

Yeah, though I really only care about keeping them going long enough to flush the logs from the tests.

PR #393, opened 2022-06-13
Fix flakey `TestRoomReceiptsReadMarkers`

erikjohnston:
We can't do the tests in parallel, as sending a `/receipt` request after
a `/read_markers` request on the same event returns an error. (While
`/read_markers` will return 200 in the reversed situation).

We fix it by splitting the test into two.


squahtx (approved):
one comment, otherwise lgtm

PR #396, opened 2022-06-14
Add a function to restart a deployment

squahtx:
Split out from #378.

### DIFF
--- a/internal/docker/deployer.go
+++ b/internal/docker/deployer.go
@@ -176,6 +176,37 @@ func (d *Deployer) Destroy(dep *Deployment, printServerLogs bool) {
 	}
 }
 
+// Restart a homeserver deployment.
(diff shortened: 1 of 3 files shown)


internal/docker/deployer.go
```diff
@@ -301,89 +338,21 @@ func deployImage(
...
-		if err == nil {
-			break
-		}
-	}
+	baseURL, fedBaseURL, err := waitForPorts(ctx, docker, containerID)
 	if err != nil {
 		return stubDeployment, fmt.Errorf("%s : image %s : %w", contextStr, imageID, err)
 	}
+	inspect, err := docker.ContainerInspect(ctx, containerID)
+	if err != nil {
```
kegsay, line 346: I don't think this will ever be hit? `waitForPorts` does `inspect, err = docker.ContainerInspect(ctx, containerID)` initially and then bails if there is an `err`. In that case, we would've returned already on `:342`.

kegsay (approved):

PR #397, opened 2022-06-14
Test homeserver restarts during a partial state join

squahtx:
Tests https://github.com/matrix-org/synapse/issues/12642

Split out from #378.


squahtx:
@erikjohnston I've split out the test part from #378, which you've already reviewed.

erikjohnston (approved):

PR #401, opened 2022-06-30
Test events arriving during faster joins

richvdh:


tests/federation_room_join_partial_state_test.go
```diff
@@ -128,6 +128,58 @@ func TestPartialStateJoin(t *testing.T) {
...
+				"body":    "Message",
+			},
+		})
+		psjResult.ServerRoom.AddEvent(event)
+		psjResult.Server.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{event.JSON()}, nil)
+		t.Logf("Derek sent event event ID %s", event.EventID())
+
+		// give Derek's event a chance to land, otherwise we will race with the faster-join completing.
+		// TODO: find a better way to do this. Or hope it will be fixed by https://github.com/matrix-org/synapse/issues/13007
+		time.Sleep(time.Second)
```
squahtx, line 158: I'm having trouble seeing how https://github.com/matrix-org/synapse/issues/13007 would help here?

squahtx (approved):
The internal bits look good.
The TODOs in the test are a little sad, but I think it's still useful to get the test in.

tests/federation_room_join_partial_state_test.go
```diff
@@ -128,6 +128,58 @@ func TestPartialStateJoin(t *testing.T) {
 		t.Logf("Alice sent event event ID %s", eventID)
 	})
 
+	// we should be able to receive events over federation during the resync
+	t.Run("CanReceiveEventsDuringPartialStateJoin", func(t *testing.T) {
```
kegsay, line 132: You can use spaces you know.

kegsay (approved):

tests/federation_room_join_partial_state_test.go
richvdh: so the problem is that, without this sleep the persistence of the incoming event races with the un-partial-stating. Before the fix to #13007, that results in a foreign key violation (because we try to record the newly-received event as having partial state after the room is un-partial-stated).

Now that the fix to #13007 has landed, we don't get an FK violation, but the event still isn't persisted immediately.

Anyway, this test is aiming to ensure that events received during the actual partial state situation are processed correctly, so I'm going to keep the sleep, but update the comment a bit.

richvdh:
I've updated this a bit. @squahtx would you mind taking another look?

PR #405, opened 2022-07-06
Add test to make sure you can find imported events before room creation - MSC3030

MadLittleMods:
Add test to make sure you can find imported events before room creation

Synapse changes: https://github.com/matrix-org/synapse/pull/13197

MSC3030

### DIFF
--- a/tests/msc3030_test.go
+++ b/tests/msc3030_test.go
@@ -16,7 +16,6 @@ import (
 
 	"github.com/matrix-org/complement/internal/b"
 	"github.com/matrix-org/complement/internal/client"
-	"github.com/sirupsen/logrus"
 	"github.com/tidwall/gjson"
 )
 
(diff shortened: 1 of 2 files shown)


tests/msc3030_test.go
```diff
@@ -127,6 +131,38 @@ func TestJumpToDateEndpoint(t *testing.T) {
...
+			t.Run("when looking backwards before the room was created, should be able to find event that was imported", func(t *testing.T) {
+				t.Parallel()
+				timeBeforeRoomCreation := time.Now()
+				roomID, _, _ := createTestRoom(t, alice)
+
+				// Join from the application service bridge user
+				as.JoinRoom(t, roomID, []string{"hs1"})
+
+				// Import a message in the room before the room was created. We have to
+				// use an application service user because they are the only ones
```
MadLittleMods, line 144: Re-commented so I can apply

```suggestion
				// use an application service user because they are the only one
```

PR #406, opened 2022-07-06
Add test to make sure we can paginate after getting remote event from timestamp to event endpoint

MadLittleMods:
Add test to make sure we can paginate after getting remote event from timestamp to event endpoint

Synapse changes: https://github.com/matrix-org/synapse/pull/13205


MadLittleMods:
Thanks for the review @DMRobertson and @kegsay 🐁

PR #408, opened 2022-07-08
Add homerunner integration tests

kegsay:
Uses homerunner-client which also serves to make sure the client works.


cmd/homerunner/test/test.sh
```diff
@@ -0,0 +1,26 @@
...
+export HOMERUNNER_PORT=5544
+export HOMERUNNER_SPAWN_HS_TIMEOUT_SECS=30
+
+# build and run homerunner
+go build ..
+echo 'Running homerunner'
+./homerunner &
+HOMERUNNER_PID=$!
+# knife homerunner when this script finishes
+trap "kill $HOMERUNNER_PID" EXIT
```
Half-Shot, line 12: TIL `trap` is a thing. That would have been SO handy for the ci scripts.

cmd/homerunner/test/test.mjs
```diff
@@ -0,0 +1,36 @@
...
+
+const run = async () => {
+    const client = new Homerunner.Client();
+    console.log("homerunner base url:", client.baseUrl);
+    console.log("creating homeserver...");
+    const blueprint = await client.create({
+        base_image_uri: "complement-dendrite",
+        blueprint_name: "one_to_one_room",
+    });
+    console.log(blueprint);
```
Half-Shot, line 11: I presume this is intentionally left in, if it is can we have some context before it like:

```suggestion
    console.log("Client.create responded with", blueprint);
```

Half-Shot (approved):
Looks fantastic, I suppose we need to hook this up to some CI in homerunner-client after this is merged.

PR #409, opened 2022-07-12
Ignore presence EDUs during partial state join test

richvdh:
Hopefully fixes https://github.com/matrix-org/synapse/issues/13256

### DIFF
--- a/tests/federation_room_join_test.go
+++ b/tests/federation_room_join_test.go
@@ -459,6 +459,9 @@ func TestSendJoinPartialStateResponse(t *testing.T) {
 
 	srv := federation.NewServer(t, deployment,
 		federation.HandleKeyRequests(),
+
+		// accept incoming presence transactions, etc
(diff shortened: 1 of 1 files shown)


DMRobertson (approved):

tests/federation_room_join_test.go
```diff
@@ -459,6 +459,9 @@ func TestSendJoinPartialStateResponse(t *testing.T) {
 
 	srv := federation.NewServer(t, deployment,
 		federation.HandleKeyRequests(),
+
+		// accept incoming presence transactions, etc
+		federation.HandleTransactionRequests(nil, nil),
```
squahtx, line 464: some explanation of where the presence is coming from would be nice!

richvdh: that would require me to figure that out. Tbh I just cargo-culted all the other tests in this file...

PR #411, opened 2022-07-13
Use `/v3` URLs instead of `/r0`

neilalexander:
We had quite a lot of tests still using `/r0` URLs.

### DIFF
--- a/internal/federation/handle.go
+++ b/internal/federation/handle.go
@@ -366,7 +366,7 @@ func HandleMediaRequests(mediaIds map[string]func(w http.ResponseWriter)) func(*
 			}
 		})
 
-		// Note: The spec says to use r0, but implementations rely on /v1 working for federation requests as a legacy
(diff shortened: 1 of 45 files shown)


internal/federation/handle.go
```diff
@@ -368,8 +368,9 @@ func HandleMediaRequests(mediaIds map[string]func(w http.ResponseWriter)) func(*
 
 		// Note: The spec says to use r0, but implementations rely on /v1 working for federation requests as a legacy
```
richvdh, line 369: s/r0/
```suggestion
		// Note: The spec says to use /v3, but implementations rely on /v1 and /r0 working for federation requests as a legacy
```

richvdh: I wonder if we actually still need to support /v1 here.

richvdh (approved):
lgtm other than comment tweak

PR #415, opened 2022-07-21
Retry /publicRooms a few times to allow for propagation delay before a new room appears in the room directory

reivilibre:
This is needed in Synapse because the room directory is updated in the background. Part of https://github.com/matrix-org/synapse/issues/13161.
It's particularly worse when using workers, because the room directory is updated in a different process and CPU contention increases the likelihood of hitting the race.

### DIFF
--- a/tests/csapi/apidoc_room_state_test.go
+++ b/tests/csapi/apidoc_room_state_test.go
@@ -3,6 +3,7 @@ package csapi_tests
 import (
 	"net/url"
 	"testing"
+	"time"
 
 	"github.com/tidwall/gjson"
 
@@ -10,6 +11,8 @@ import (
 	"github.com/matrix-org/complement/internal/client"
 	"github.com/matrix-org/complement/internal/match"
 	"github.com/matrix-org/complement/internal/must"
+
+	"net/http"
 )
 
 func TestRoomState(t *testing.T) {
@@ -104,26 +107,31 @@ func TestRoomState(t *testing.T) {
 				"preset":     "public_chat",
 			})
 
-			res := authedClient.MustDoFunc(t, "GET", []string{"_matrix", "client", "v3", "publicRooms"})
-			foundRoom := false
+			authedClient.MustDoFunc(t, "GET", []string{"_matrix", "client", "v3", "publicRooms"},
+				client.WithRetryUntil(time.Second, func(res *http.Response) bool {
+					foundRoom := false
 
-			must.MatchResponse(t, res, match.HTTPResponse{
-				JSON: []match.JSON{
-					match.JSONKeyPresent("chunk"),
-					match.JSONArrayEach("chunk", func(val gjson.Result) error {
-						gotRoomID := val.Get("room_id").Str
-						if gotRoomID == roomID {
-							foundRoom = true
-							return nil
-						}
-						return nil
-					}),
-				},
-			})
+					must.MatchResponse(t, res, match.HTTPResponse{
+						JSON: []match.JSON{
+							match.JSONKeyPresent("chunk"),
+							match.JSONArrayEach("chunk", func(val gjson.Result) error {
+								gotRoomID := val.Get("room_id").Str
+								if gotRoomID == roomID {
+									foundRoom = true
+									return nil
+								}
+								return nil
+							}),
+						},
+					})
 
-			if !foundRoom {
-				t.Errorf("failed to find room with id: %s", roomID)
-			}
+					if !foundRoom {
+						t.Logf("failed to find room with id: %s", roomID)
+					}
+
+					return foundRoom
+				}),
+			)
 		})
 		// sytest: GET /directory/room/:room_alias yields room ID
 		t.Run("GET /directory/room/:room_alias yields room ID", func(t *testing.T) {
--- a/tests/knocking_test.go
+++ b/tests/knocking_test.go
@@ -14,6 +14,8 @@ import (
 	"testing"
 	"time"
 
+	"net/http"
+
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/tidwall/gjson"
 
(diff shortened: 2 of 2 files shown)


internal/client/client.go
```diff
@@ -453,6 +453,31 @@ func (c *CSAPI) MustDoFunc(t *testing.T, method string, paths []string, opts ...
...
+	var res *http.Response
+	res = nil
+	for retryNum := 0; retryNum < 5; retryNum++ {
+		res = c.DoFunc(t, method, paths, opts...)
+		if res.StatusCode < 200 || res.StatusCode >= 300 {
+			defer res.Body.Close()
+			body, _ := ioutil.ReadAll(res.Body)
+			t.Fatalf("CSAPI.MustDoFuncUntil response return non-2xx code: %s - body: %s", res.Status, string(body))
+			return res
+		}
```
DMRobertson, line 468: Two points of confusion:

1. I think this means you can't use `MustDoFuncUntil` to repeat a successful call until it fails. E.g. imagine a test where:
  - join a room
  - someone bans you
  - repeatedly try to change your name in the room until the ban takes place, at which point the name change fails.
2. I think this means that you can't use `MustDoFuncUntil` to repeat a failing call until it suceeds. E.g. imagine a test where:
   - fail to join a restricted room because you're not in the right space
   - join the space
   - repeatedly join the restricted room until it succeeds.

Though thinking about it some more, perhaps those ought be different functions e.g. some sort of `DoFuncUntil`.

internal/client/client.go
```diff
@@ -453,6 +453,31 @@ func (c *CSAPI) MustDoFunc(t *testing.T, method string, paths []string, opts ...
...
+
+		success := checkFunc(res)
+		if t.Failed() || success {
+			return res
+		}
+
+		time.Sleep(500 * time.Millisecond)
```
DMRobertson, line 475: If the average race is resolved quicker than this, say 100ms, then this is a little pessimistic. Might exponential backoff (rather than linear) make more sense? (e.g. 10ms delay, quadrupling every retry)

DMRobertson (commented):
Drive-by thoughts. No idea what a Waiter is (https://pkg.go.dev/github.com/aws/aws-sdk-go/private/waiter#Waiter?) but if it's more idiomatic then that might be the way to go.

internal/client/client.go
reivilibre: That's fair; not sure what better name to put on it (I've tried to keep the names consistent with what's there; I'd personally say those names aren't very intuitive either but they have docstrings).
At the very least, describing your caveats in the docstring would probably help though.

internal/client/client.go
```diff
@@ -453,6 +454,40 @@ func (c *CSAPI) MustDoFunc(t *testing.T, method string, paths []string, opts ...
...
+// MustDoFuncUntil is the same as MustDoFunc but retries up to 5 times,
+// with exponential backoff (10 ms, 40 ms, 160 ms, ...) whilst the predicate is not satisfied.
+//
+// It will retry a successful call (2xx HTTP response code) until the predicate
+// described by `checkFunc` is satisfied, timing out after 5 retries.
+//
+// Caveats:
+// 1. You can't use MustDoFuncUntil to repeat a successful call until it fails.
+// 2. You can't use MustDoFuncUntil to repeat a failing call until it suceeds.
+func (c *CSAPI) MustDoFuncUntil(t *testing.T, method string, paths []string, checkFunc func(*http.Response) bool, opts ...RequestOpt) *http.Response {
```
kegsay, line 466: Please don't add new functions which are basically the same as other functions. Add a `RequestOpt`.

internal/client/client.go
```diff
@@ -453,6 +454,40 @@ func (c *CSAPI) MustDoFunc(t *testing.T, method string, paths []string, opts ...
...
+	for retryNum := 0; retryNum < 5; retryNum++ {
+		res = c.DoFunc(t, method, paths, opts...)
+		if res.StatusCode < 200 || res.StatusCode >= 300 {
+			defer res.Body.Close()
+			body, _ := ioutil.ReadAll(res.Body)
+			t.Fatalf("CSAPI.MustDoFuncUntil response return non-2xx code: %s - body: %s", res.Status, string(body))
+			return res
+		}
+
+		success := checkFunc(res)
```
kegsay, line 479: Please don't conflate performing operations (e.g HTTP requests) with verifying the results of said operation (assertions). This ends up making inflexible code.

tests/csapi/apidoc_room_state_test.go
```diff
@@ -104,26 +106,29 @@ func TestRoomState(t *testing.T) {
...
-				JSON: []match.JSON{
-					match.JSONKeyPresent("chunk"),
-					match.JSONArrayEach("chunk", func(val gjson.Result) error {
-						gotRoomID := val.Get("room_id").Str
-						if gotRoomID == roomID {
-							foundRoom = true
+			authedClient.MustDoFuncUntil(t, "GET", []string{"_matrix", "client", "v3", "publicRooms"}, func(res *http.Response) bool {
+				foundRoom := false
+
+				must.MatchResponse(t, res, match.HTTPResponse{
```
kegsay, line 112: You can't use `MatchResponse` in this way as it will fail the test if the response does not pass the given matchers.

kegsay (changes requested):
What we want is something which will repeat the action until some condition is satisfied, at which point it will return the response. This response can then have assertions applied to it. Currently, this PR conflates the two: the presence of the room ID is both the condition to satisfy _and_ the assertion. This doesn't compose well because:
 - You cannot use assertion helpers (like matchers) or anything which will fail the test as part of your code to repeat until the condition is satisfied, as the first failure will fail the test.
 - You cannot say things like "do this request until you see this field, then give me the response so I can inspect the fields more deeply".

internal/client/client.go
kegsay: This is too inflexible. Some endpoints may take longer than others to update, it needs to be configurable how long to keep trying for.

tests/csapi/apidoc_room_state_test.go
reivilibre: (NB Diff is outdated but basic structure remains the same, so will reply to comment)

I believe this is what we want, though (or at least I'm used to it from SyTest); if the response is ever malformed then we want to fail the test immediately without waiting for it to be correct.

Put another way: The retries are waiting for the expected data to appear in the response, but that doesn't mean that any responses before that are allowed to omit required fields.
As a result, you end up with 3 outcomes: success, temporary fail (i.e. retry) and permanent fail.

As for how these currently seem to map to code:
Success = `return true`
Temporary fail = `return false`
Permanent fail = `t.FailNow`/`t.Fatalf`

erikjohnston (approved):
Seems legit to me. Will defer to @kegsay that this is the Right Way

tests/csapi/apidoc_room_state_test.go
```diff
@@ -104,26 +107,31 @@ func TestRoomState(t *testing.T) {
...
+					must.MatchResponse(t, res, match.HTTPResponse{
+						JSON: []match.JSON{
+							match.JSONKeyPresent("chunk"),
+							match.JSONArrayEach("chunk", func(val gjson.Result) error {
+								gotRoomID := val.Get("room_id").Str
+								if gotRoomID == roomID {
+									foundRoom = true
+									return nil
+								}
+								return nil
```
kegsay, line 123: This is a slight misuse of matchers, given you always return `nil` it will always succeed even if the room isn't found, and then you side-effect `foundRoom` to determine retry-ness. It's not something that the matchers were designed for, but in practice it seems to read pretty well.

kegsay (approved):
LGTM

tests/csapi/apidoc_room_state_test.go
reivilibre: You're right, though I will say 'don't blame me, it was there before!' :-P

I suspect this sort of pattern will crop up again, so if you get any better ideas then it may be worth putting them in

PR #417, opened 2022-07-21
Add WithRetryUntil request opt which can be used with any CSAPI call

kegsay:
This will repeat a request until the condition is satisfied.

Intended use: https://github.com/matrix-org/complement/pull/415


reivilibre (approved):
LGTM — thanks for polishing this idea up into something more idiomatic for Complement.

PR #418, opened 2022-07-21
Regression test for infinite-loop-in-resync

richvdh:
A rather complicated test which demonstrated that the infinite loop was in https://github.com/matrix-org/synapse/issues/13001 was possible.

~~The test currently fails: it's getting stuck waiting for soft-fail, which is something @squahtx is working on.~~

### DIFF
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -580,21 +580,6 @@ func TestPartialStateJoin(t *testing.T) {
 
 		// Alice has now joined the room, and the server is syncing the state in the background.
 
-		// utility function to build a regular event in the test room
-		makeTimelineEvent := func(body string) *gomatrixserverlib.Event {
-			event := psjResult.Server.MustCreateEvent(t, psjResult.ServerRoom, b.Event{
-				Type:   "m.room.message",
-				Sender: psjResult.Server.UserID("charlie"),
-				Content: map[string]interface{}{
-					"msgtype": "m.text",
-					"body":    body,
-				},
-			})
-			psjResult.ServerRoom.AddEvent(event)
-			t.Logf("Created event %s: %s", body, event.EventID())
-			return event
-		}
-
 		// utility function to wait for a given event to arrive at the remote server.
 		// This works simply by polling /event until we get a 200.
 		awaitEventArrival := func(eventID string) {
@@ -617,7 +602,7 @@ func TestPartialStateJoin(t *testing.T) {
 		}
 
 		// here's the first event which we *ought* to un-partial-state, but won't
-		lateEvent := makeTimelineEvent("late event")
+		lateEvent := psjResult.CreateMessageEvent(t, "charlie", nil)
 
 		// next, we want to create 100 outliers. So, charlie creates 100 state events, and
 		// then persuades the system under test to create a backwards extremity using those events as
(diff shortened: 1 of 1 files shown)


tests/federation_room_join_partial_state_test.go
```diff
@@ -511,6 +513,192 @@ func TestPartialStateJoin(t *testing.T) {
 			t.Errorf("Did not find derek's m.room.member event in gappy /sync response: %s", err)
 		}
 	})
+
+	// regression test for https://github.com/matrix-org/synapse/issues/13001
+	//
+	// There wass an edge case where, if we initially receive lots of events as outliers,
```
kegsay, line 519:
```suggestion
	// There was an edge case where, if we initially receive lots of events as outliers,
```

tests/federation_room_join_partial_state_test.go
```diff
@@ -511,6 +513,192 @@ func TestPartialStateJoin(t *testing.T) {
...
+				t.Fatalf("GET /event failed with %d: %s", res.StatusCode, string(eventResBody))
+			}
+			t.Fatalf("timeout waiting for event %s to be received", eventID)
+		}
+
+		// here's the first event which we *ought* to un-partial-state, but won't
+		lateEvent := makeTimelineEvent("late event")
+
+		// next, we want to create 100 outliers. So, charlie creates 100 state events, and
+		// then persuades the SUT to create a backwards extremity using those events as
```
kegsay, line 571: Unusual acronym 'SUT'?

richvdh: "system under test". will reword.

richvdh:
seems like this would benefit from some of the utility functions added in https://github.com/matrix-org/complement/pull/419. I'll update it once that lands.

richvdh:
This is now updated to use the utility functions added in #419. It includes the changes in #428.

tests/federation_room_join_partial_state_test.go
```diff
@@ -563,6 +565,128 @@ func TestPartialStateJoin(t *testing.T) {
 			t.Errorf("Did not find derek's m.room.member event in gappy /sync response: %s", err)
 		}
 	})
+
+	// regression test for https://github.com/matrix-org/synapse/issues/13001
+	//
+	// There was an edge case where, if we initially receive lots of events as outliers,
+	// and they then get de-outliered as partial state events,
+	t.Run("Resync works with many prev_events with partial state", func(t *testing.T) {
```
squahtx, line 573: Are we missing the other half of the comment?

tests/federation_room_join_partial_state_test.go
```diff
@@ -563,6 +565,128 @@ func TestPartialStateJoin(t *testing.T) {
...
+				}
+				if res.StatusCode == 404 && gjson.GetBytes(eventResBody, "errcode").String() == "M_NOT_FOUND" {
+					t.Logf("Fetching event %s failed with M_NOT_FOUND; will retry", eventID)
+					time.Sleep(1000 * time.Millisecond)
+					continue
+				}
+				t.Fatalf("GET /event failed with %d: %s", res.StatusCode, string(eventResBody))
+			}
+			t.Fatalf("timeout waiting for event %s to be received", eventID)
+		}
```
squahtx, line 602: It was suggested that we use `WithRetryUntil` here: https://github.com/matrix-org/complement/pull/419#discussion_r928706407 , which looks like identical logic with a different timeout.

We might also want to make this helper available to other tests (and deduplicate the copy from #419)

tests/federation_room_join_partial_state_test.go
squahtx: Also can we make the name of the test more specific? something along the lines of persisting chains of prev_events out of order maybe

richvdh: I've had a go at fixing the name of the test - perhaps it's better?

tests/federation_room_join_partial_state_test.go
richvdh: good idea, thanks. Done.

tests/federation_room_join_partial_state_test.go
squahtx: Looks alright!

squahtx (approved):
lgtm!

PR #419, opened 2022-07-21
Add more tests for receiving events while resyncing for a partial join

squahtx:
Add tests for three cases where a homeserver receives a latest event with missing prev events while it is still resyncing state after a partial join.

Namely:
```
... <-- M <-- A <-- B
```
```
        +---------+
        v          \
... <-- M <-- A <-- B
```
and
```
        +---------+
        v          \
... <-- M <-- A <-- B <-- C
```
where event M is the last event the homeserver knows about and the rightmost event is the one it receives.

The last case tests the code path mentioned in https://github.com/matrix-org/synapse/issues/13002

Best reviewed commit by commit.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "go.testEnvVars": {
-        "COMPLEMENT_BASE_IMAGE": "complement-synapse"
-    }
-}
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -142,15 +142,7 @@ func TestPartialStateJoin(t *testing.T) {
 		federation.HandleEventAuthRequests()(psjResult.Server)
 
 		// derek sends an event in the room
-		event := psjResult.Server.MustCreateEvent(t, psjResult.ServerRoom, b.Event{
-			Type:   "m.room.message",
-			Sender: psjResult.Server.UserID("derek"),
-			Content: map[string]interface{}{
-				"msgtype": "m.text",
-				"body":    "Message",
-			},
-		})
-		psjResult.ServerRoom.AddEvent(event)
+		event := psjResult.CreateMessageEvent(t, "derek", nil)
 		psjResult.Server.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{event.JSON()}, nil)
 		t.Logf("Derek sent event event ID %s", event.EventID())
 
@@ -618,6 +610,28 @@ func (psj *partialStateJoinResult) Destroy() {
 	if psj.cancelListener != nil {
 		psj.cancelListener()
 	}
+}
+
+// send a message into the room without letting the homeserver under test know about it.
+func (psj *partialStateJoinResult) CreateMessageEvent(t *testing.T, senderLocalpart string, prevEventIDs []string) *gomatrixserverlib.Event {
+	var prevEvents interface{}
+	if prevEventIDs == nil {
+		prevEvents = nil
+	} else {
+		prevEvents = prevEventIDs
+	}
+
+	event := psj.Server.MustCreateEvent(t, psj.ServerRoom, b.Event{
+		Type:   "m.room.message",
+		Sender: psj.Server.UserID(senderLocalpart),
+		Content: map[string]interface{}{
+			"msgtype": "m.text",
+			"body":    "Message",
+		},
+		PrevEvents: prevEvents,
+	})
+	psj.ServerRoom.AddEvent(event)
+	return event
 }
 
 // wait for a /state_ids request for the test room to arrive
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -632,6 +632,8 @@ func testReceiveEventDuringPartialStateJoin(
 	select {
 	case <-time.After(1 * time.Second):
 		t.Logf("/state_ids request for event %s blocked as expected", event.EventID())
+		// read from the channel and discard the result, so that the goroutine above doesn't block
+		// indefinitely on the send
 		defer func() { <-stateIdsResultChan }()
 		break
 	case stateIDsResult := <-stateIdsResultChan:
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -592,6 +592,8 @@ type partialStateJoinResult struct {
 	ServerRoom                       *federation.ServerRoom
 	fedStateIdsRequestReceivedWaiter *Waiter
 	fedStateIdsSendResponseWaiter    *Waiter
+	// the set of events for which we will not block `/state` or `/state_ids` requests.
+	fedStateIdsAllowedEvents map[string]bool
 }
 
 // beginPartialStateJoin spins up a room on a complement server,
@@ -627,6 +629,7 @@ func beginPartialStateJoin(t *testing.T, deployment *docker.Deployment, joiningU
 	// some things for orchestration
 	result.fedStateIdsRequestReceivedWaiter = NewWaiter()
 	result.fedStateIdsSendResponseWaiter = NewWaiter()
+	result.fedStateIdsAllowedEvents = make(map[string]bool)
 
 	// create the room on the complement server, with charlie and derek as members
 	roomVer := joiningUser.GetDefaultRoomVersion(t)
@@ -642,10 +645,17 @@ func beginPartialStateJoin(t *testing.T, deployment *docker.Deployment, joiningU
 
 	// register a handler for /state_ids requests, which finishes fedStateIdsRequestReceivedWaiter, then
 	// waits for fedStateIdsSendResponseWaiter and sends a reply
-	handleStateIdsRequests(t, result.Server, result.ServerRoom, result.fedStateIdsRequestReceivedWaiter, result.fedStateIdsSendResponseWaiter)
+	handleStateIdsRequests(
+		t,
+		result.Server,
+		result.ServerRoom,
+		result.fedStateIdsRequestReceivedWaiter,
+		result.fedStateIdsSendResponseWaiter,
+		result.fedStateIdsAllowedEvents,
+	)
 
 	// a handler for /state requests, which sends a sensible response
-	handleStateRequests(t, result.Server, result.ServerRoom, nil, nil)
+	handleStateRequests(t, result.Server, result.ServerRoom, nil, nil, nil)
 
 	// have joiningUser join the room by room ID.
 	joiningUser.JoinRoom(t, result.ServerRoom.RoomID, []string{result.Server.ServerName()})
@@ -693,6 +703,12 @@ func (psj *partialStateJoinResult) CreateMessageEvent(t *testing.T, senderLocalp
 	return event
 }
 
+// allow a /state_ids request for a given event to complete before FinishStateRequest has been called.
+// only applies to new incoming requests, and not any currently blocked ones.
+func (psj *partialStateJoinResult) AllowStateRequestForEvent(eventID string) {
+	psj.fedStateIdsAllowedEvents[eventID] = true
+}
+
 // wait for a /state_ids request for the test room to arrive
 func (psj *partialStateJoinResult) AwaitStateIdsRequest(t *testing.T) {
 	psj.fedStateIdsRequestReceivedWaiter.Waitf(t, 5*time.Second, "Waiting for /state_ids request")
@@ -709,7 +725,7 @@ func (psj *partialStateJoinResult) FinishStateRequest() {
 // if sendResponseWaiter is not nil, we will Wait() for it to finish before sending the response.
 func handleStateIdsRequests(
 	t *testing.T, srv *federation.Server, serverRoom *federation.ServerRoom,
-	requestReceivedWaiter *Waiter, sendResponseWaiter *Waiter,
+	requestReceivedWaiter *Waiter, sendResponseWaiter *Waiter, allowedEvents map[string]bool,
 ) {
 	srv.Mux().Handle(
 		fmt.Sprintf("/_matrix/federation/v1/state_ids/%s", serverRoom.RoomID),
@@ -719,7 +735,8 @@ func handleStateIdsRequests(
 			if requestReceivedWaiter != nil {
 				requestReceivedWaiter.Finish()
 			}
-			if sendResponseWaiter != nil {
+			if !allowedEvents[queryParams["event_id"][0]] &&
+				sendResponseWaiter != nil {
 				sendResponseWaiter.Waitf(t, 60*time.Second, "Waiting for /state_ids request")
 			}
 			t.Logf("Replying to /state_ids request")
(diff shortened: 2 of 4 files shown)


tests/federation_room_join_partial_state_test.go
```diff
@@ -620,6 +786,34 @@ func (psj *partialStateJoinResult) Destroy() {
...
 }
 
+// send a message into the room without letting the homeserver under test know about it.
+func (psj *partialStateJoinResult) CreateMessageEvent(t *testing.T, senderLocalpart string, prevEventIDs []string) *gomatrixserverlib.Event {
+	var prevEvents interface{}
+	if prevEventIDs == nil {
+		prevEvents = nil
+	} else {
+		prevEvents = prevEventIDs
+	}
```
squahtx, line 796: We do this because `interface{}(nil)` is not equal to `[]string(nil)`

tests/federation_room_join_partial_state_test.go
```diff
@@ -513,6 +558,117 @@ func TestPartialStateJoin(t *testing.T) {
...
+		break
+	case respStateIDs := <-stateIdsResponseChan:
+		if respStateIDs == nil {
+			t.Logf("/state_ids request for event %s returned 403 as expected", event.EventID())
+		} else {
+			// since we have not yet given the homeserver the full state at the join event and allowed
+			// the partial join to complete, it can't possibly know the full state at the last event.
+			// While it may be possible for the response to be correct by some accident of state res,
+			// the homeserver is still wrong in spirit.
+			t.Fatalf("/state_ids request for event %s did not block when it should have", event.EventID())
```
squahtx, line 639: This part tests https://github.com/matrix-org/synapse/issues/13002 . If Synapse does not correctly flag that it does not have partial state in the `CanReceiveEventsWithMissingPrevWithHalfMissingPrevsDuringPartialStateJoin` test case, it will erroneously return state (which was derived via state res and only happens to be correct) instead of blocking.

tests/federation_room_join_partial_state_test.go
```diff
@@ -142,75 +146,116 @@ func TestPartialStateJoin(t *testing.T) {
...
+
+		// the HS will make a /get_missing_events request for the missing prev event of event B
+		handleGetMissingEventsRequests(t, psjResult.Server, psjResult.ServerRoom, []*gomatrixserverlib.Event{eventA})
+
+		// send event B to hs1
+		testReceiveEventDuringPartialStateJoin(t, deployment, alice, psjResult, eventB)
+	})
+
+	// we should be able to receive events with missing prevs over federation during the resync
+	t.Run("CanReceiveEventsWithMissingPrevWithHalfMissingPrevsDuringPartialStateJoin", func(t *testing.T) {
```
squahtx, line 225: Suggestions for better names are welcome.

tests/federation_room_join_partial_state_test.go
```diff
@@ -489,6 +490,51 @@ func testReceiveEventDuringPartialStateJoin(
 		t.Fatalf("GET /event failed with %d: %s", res.StatusCode, string(eventResBody))
 	}
 
+	// fire off a /state_ids request for the last event.
+	// it must either:
+	//   * block because the homeserver does not have full state at the last event
+	//   * or 403 because the homeserver does not have full state yet and does not consider the
+	//     Complement homeserver to be in the room
```
richvdh, line 604: shouldn't we know which, and test for that specifically?

tests/federation_room_join_partial_state_test.go
```diff
@@ -489,6 +490,51 @@ func testReceiveEventDuringPartialStateJoin(
...
+			t.Errorf("/state_ids request returned non-200: %s", err)
+			return
+		}
+		stateIdsResponseChan <- &respStateIDs
+	}()
+
+	select {
+	case <-time.After(1 * time.Second):
+		t.Logf("/state_ids request for event %s blocked as expected", event.EventID())
+		defer func() { <-stateIdsResponseChan }()
```
richvdh, line 631: what does this defer thing do?

tests/federation_room_join_partial_state_test.go
```diff
@@ -489,6 +490,51 @@ func testReceiveEventDuringPartialStateJoin(
...
+			httpErr, ok := err.(gomatrix.HTTPError)
+			t.Logf("%v", httpErr)
+			if ok && httpErr.Code == 403 {
+				stateIdsResponseChan <- nil
+				return
+			}
+			t.Errorf("/state_ids request returned non-200: %s", err)
+			return
+		}
+		stateIdsResponseChan <- &respStateIDs
```
richvdh, line 625: rather than matching for the 403 here and sending a `nil` magic value down the channel, I suggest you define 

```go
type stateIDsResult struct {
    RespStateIDs gomatrixserverlib.RespStateIDs
    Error error
}
```

... and then send that back down the channel, and do the matching in the main thread.

tests/federation_room_join_partial_state_test.go
```diff
@@ -642,10 +645,17 @@ func beginPartialStateJoin(t *testing.T, deployment *docker.Deployment, joiningU
...
 	// register a handler for /state_ids requests, which finishes fedStateIdsRequestReceivedWaiter, then
 	// waits for fedStateIdsSendResponseWaiter and sends a reply
-	handleStateIdsRequests(t, result.Server, result.ServerRoom, result.fedStateIdsRequestReceivedWaiter, result.fedStateIdsSendResponseWaiter)
+	handleStateIdsRequests(
+		t,
+		result.Server,
+		result.ServerRoom,
+		result.fedStateIdsRequestReceivedWaiter,
+		result.fedStateIdsSendResponseWaiter,
+		result.fedStateIdsAllowedEvents,
```
richvdh, line 761: I think it might be easier to flip this around, and have handleStateIdsRequests only match /state_ids requests for the last event before the join event? That's what I did in #418, in https://github.com/matrix-org/complement/pull/418/commits/687ad360910260a0cf6167cf3df36a97694297b4.

tests/federation_room_join_partial_state_test.go
richvdh: maybe "can receive events with some missing grandparents" or something?

tests/federation_room_join_partial_state_test.go
```diff
@@ -221,6 +221,43 @@ func TestPartialStateJoin(t *testing.T) {
...
+		// initially, hs1 only knows about event M.
+		// we send only event C to hs1.
+		eventM := psjResult.ServerRoom.CurrentState("m.room.member", alice.UserID)
+		eventA := psjResult.CreateMessageEvent(t, "derek", []string{eventM.EventID()})
+		eventB := psjResult.CreateMessageEvent(t, "derek", []string{eventA.EventID(), eventM.EventID()})
+		eventC := psjResult.CreateMessageEvent(t, "derek", []string{eventB.EventID()})
+		t.Logf("%s's m.room.member event is %s", *eventM.StateKey(), eventM.EventID())
+		t.Logf("Derek sent event A with ID %s", eventA.EventID())
+		t.Logf("Derek sent event B with ID %s", eventB.EventID())
+		t.Logf("Derek sent event C with ID %s", eventC.EventID())
```
richvdh, line 250: s/sent/created/. He hasn't sent them to hs1 yet.

richvdh (commented):
looks generally good, a few suggestions.

tests/federation_room_join_partial_state_test.go
squahtx: when the `/state_ids` request blocks, the send of the response happens some time later, perhaps even after the test ends.
If we don't read from the channel, the goroutine above blocks on sending into the channel and complement gets very unhappy 10 minutes later because it's still not finished.

richvdh: right, so we're making sure we read from the channel so that the goroutine doesn't block. Can you add a comment to that effect?

tests/federation_room_join_partial_state_test.go
squahtx: That looks a lot cleaner. I've borrowed that commit and it works great. Do we want to split that change into a separate PR?

richvdh: yeah, I'll do that.

richvdh: #421

tests/federation_room_join_partial_state_test.go
squahtx: Right now only the last test case blocks instead of 403s. I'm somewhat reluctant to be more specific in the tests, because it feels like an implementation detail of synapse (which will change when we get around to https://github.com/matrix-org/synapse/issues/13288).

If you're really keen on being more specific, I'm happy to add a boolean parameter or suchlike. (It's not easy to move this part of the test out, since it needs to happen after the /event request above has succeeded)

richvdh: ah right. tbh I don't think either blocking or returning a 403 is correct behaviour, though I'm not quite sure what is.

It would be very useful to add a link to #13288 and say that this is somewhat temporary until we solve it properly.

tests/federation_room_join_partial_state_test.go
```diff
@@ -613,26 +623,28 @@ func testReceiveEventDuringPartialStateJoin(
...
 		}
-		stateIdsResponseChan <- &respStateIDs
 	}()
 
 	select {
 	case <-time.After(1 * time.Second):
 		t.Logf("/state_ids request for event %s blocked as expected", event.EventID())
-		defer func() { <-stateIdsResponseChan }()
+		// read from the channel and discard the result, so that the goroutine above doesn't block
+		// indefinitely on the send
```
richvdh, line 636: (I'm a bit surprised that closing the channel isn't sufficient for this, but fair enough)

richvdh (approved):
lgtm

tests/federation_room_join_partial_state_test.go
squahtx: hm, is there a better way to test whether synapse thinks an event has partial state?
this part of the test was intended to check that `_resolve_state_at_missing_prevs` calculates the partial state flag correctly

tests/federation_room_join_partial_state_test.go
squahtx: We get a `panic: send on closed channel` without this. To clarify:

no close and no read -> complement times out after 10 minutes
defer close + no read -> panic
defer close + defer read -> complement is happy

squahtx:
> > ```
> > // fire off a /state_ids request for the last event.
> > // it must either:
> > //   * block because the homeserver does not have full state at the last event
> > //   * or 403 because the homeserver does not have full state yet and does not consider the
> > //     Complement homeserver to be in the room
> > ```
> hm, is there a better way to test whether synapse thinks an event has partial state? this part of the test was intended to check that `_resolve_state_at_missing_prevs` calculates the partial state flag correctly and it sounds like it's not a good idea to rely on /state_ids

Before we merge, is there a better way to check that Synapse has calculated the partial state flag for an event correctly?

tests/federation_room_join_partial_state_test.go
```diff
@@ -513,6 +563,128 @@ func TestPartialStateJoin(t *testing.T) {
...
+	)
+	*/
+
+	// still, Alice should be able to see the event with an /event request. We might have to try it a few times.
+	start := time.Now()
+	for {
+		if time.Since(start) > time.Second {
+			t.Fatalf("timeout waiting for received event to be visible")
+		}
+		res := alice.DoFunc(t, "GET", []string{"_matrix", "client", "r0", "rooms", psjResult.ServerRoom.RoomID, "event", event.EventID()})
```
kegsay, line 591: You can now use [WithRetryUntil](https://github.com/matrix-org/complement/blob/main/internal/client/client.go#L459) in your `DoFunc` so you can remove the `for` loop etc. It would look a bit like:
```go
res := alice.DoFunc(t, "GET", []string{"_matrix", "client", "r0", "rooms", psjResult.ServerRoom.RoomID, "event", event.EventID()},
client.WithRetryUntil(time.Second, func(res *http.Response) bool {
    if res.StatusCode == 200 {
        return true
    }
    eventResBody := client.ParseJSON(t, res)
    if res.StatusCode == 404 && gjson.GetBytes(eventResBody, "errcode").String() == "M_NOT_FOUND" {
        return false
    }
    t.Fatalf("GET /event failed with %d: %s", res.StatusCode, string(eventResBody))
    return false
})
```

tests/federation_room_join_partial_state_test.go
kegsay: I don't really see why this is done in a goroutine at all - is this because `SendFederationRequest` can block indefinitely? There is a default [30s timeout](https://github.com/matrix-org/gomatrixserverlib/blob/fc2ea1e62e4606cfc415fb0da826ce69586a51d6/client.go#L39) - if that is too long then you need to send a custom context with a timeout here - https://github.com/matrix-org/complement/blob/d842670c55a2cdc61ec12e1640bc593b3f982542/internal/federation/server.go#L227 - currently this is `context.Background()` so it won't ever expire, but you can change that easily by doing:
```go
ctx, cancel := context.WithTimeout(context.Background(), time.Second)
defer cancel()
httpClient.DoRequestAndParseResponse(ctx, httpReq, resBody)
```

This will require a change to Complement's API to:
```go
func (s *Server) SendFederationRequest(ctx context.Context, deployment *docker.Deployment, req gomatrixserverlib.FederationRequest, resBody interface{}) error {
```

but that's fine.

kegsay (changes requested):
Use `client.WithRetryUntil` and add a `context` to `SendFederationRequest` and you should be able to remove a fair chunk of boilerplate here, and help other test authors.

tests/federation_room_join_partial_state_test.go
richvdh: none that I can immediately think of. The ideal thing would be to check that *after* the resync, /state_ids returns some result that would have been impossible without the information conveyed during the resync - but that sounds rather fiddly for a test (and, probably, a hard-to-understand test failure if it later turns out not to be true).

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "go.testEnvVars": {
+        "COMPLEMENT_BASE_IMAGE": "complement-synapse"
+    }
+}
```
richvdh, line 5: not sure if we should be commiting this?

richvdh (approved):
lgtm otherwise.

.vscode/settings.json
kegsay: Ideally not, as it will make Dendrite people sad.

kegsay (approved):

.vscode/settings.json
squahtx: Good spot! I must have added it by accident while merging in `main`.

squahtx:
Thank you both for the reviews!

tests/federation_room_join_partial_state_test.go
squahtx: Thanks for the suggestion, that's a lot cleaner!

PR #421, opened 2022-07-22
Faster joins tests: make request handlers more specific

richvdh:
I'm going to add a test which will involve *multiple* /state and /state_ids
requests, so we need to make the registered handlers more selective.


squahtx (approved):
lgtm!

PR #422, opened 2022-07-22
Fix tests to not depend on unspecced `room_id` parameter for each child room.

### LINKED ISSUE #416: Incorrectly tests the results of MSC2946 /hierarchy response
The tests for the `/hierarchy` endpoint are incorrect in that they assume each entry in `children_state` includes a `room_id` field. This was removed from MSC2946 at some point, but was never updated in complement.

Synapse no longer depends on this field as of matrix-org/synapse#12991, but still returns it. I assume Dendrite also returns it since those tests are skipped there.

I believe this would affect:

* [room_hierarchy_test.go](https://github.com/matrix-org/complement/blob/e43c1a0d5785e4728652079e3d118b77bbb4c41d/tests/room_hierarchy_test.go)
* [restricted_room_hierarchy_test.go](https://github.com/matrix-org/complement/blob/e43c1a0d5785e4728652079e3d118b77bbb4c41d/tests/restricted_room_hierarchy_test.go)

To test using Synapse I was running the following in my Synapse checkout:

```sh
$ COMPLEMENT_DIR=~/matrix/complement ./scripts-dev/complement.sh -run "TestClientSpacesSummary|TestClientSpacesSummaryJoinRules|TestFederatedClientSpaces|TestRestrictedRoomsSpacesSummaryLocal|TestRestrictedRoomsSpacesSummaryFederation"
```

clokep:
The `/hierarchy` response should not return a `room_id` in each entry in the `children_state` array. This ensures we're not using this (and are using the `room_id` from the parent object).

Fixes #416 

For cross-referencing:

* Removing the field from Synapse: [matrix-org/synapse#10946](https://github.com/matrix-org/synapse/issues/10946#issuecomment-1150206341)
* Removing the field from Dendrite: matrix-org/dendrite#2588

### DIFF
--- a/tests/room_hierarchy_test.go
+++ b/tests/room_hierarchy_test.go
@@ -37,6 +37,21 @@ func eventKey(srcRoomID, dstRoomID, evType string) string {
 	return srcRoomID + "|" + dstRoomID + "|" + evType
 }
 
+// Shared mapper function to return a structure comparison string for JSONCheckOff.
+func roomToChildrenMapper(r gjson.Result) interface{} {
+	roomId := r.Get("room_id").Str
+
+	result := ""
+	for i, res := range r.Get("children_state").Array() {
+		if i != 0 {
+			result += ";"
+		}
+		result += eventKey(roomId, res.Get("state_key").Str, res.Get("type").Str)
+	}
+
+	return result
+}
+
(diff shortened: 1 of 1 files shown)


neilalexander (approved):
LGTM

kegsay (approved):

PR #423, opened 2022-07-22
Request review from Synapse & Dendrite teams.

clokep:
I'm not sure we necessarily want this, but it seems like the reasonable thing to do for this repo?


DMRobertson (approved):
SGTM (I always do this manually on complement PRs)

kegsay (approved):

PR #426, opened 2022-07-25
WIP feature dirs

kegsay:
Somewhat arbitrary split of tests to see what happens on GHA.

Not mergable in its current state because:
 - Duplication in `main_test.go`
 - `/tests/csapi` isn't evenly balanced (330s on my machine, vs <200s for all the others)


kegsay:
Seems no quicker really, at least with the tests split this way.

PR #428, opened 2022-07-26
Make `handleGetMissingEventsRequests` more general

richvdh:
A few changes:
 * make it check the requested `latest_events` to check that we are replying to the request we think we are
 * Use the structures in `gomatrixserverlib` rather than defining our own
 * General cleanups

### DIFF
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -14,17 +14,16 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"reflect"
 	"strconv"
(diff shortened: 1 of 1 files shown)


squahtx (commented):
Was the `serverRoom *federation.ServerRoom` -> `roomID string` change reverted by accident? It looked sensible

Otherwise lgtm

richvdh:
> Was the serverRoom *federation.ServerRoom -> roomID string change reverted by accident? It looked sensible

I decided it made it match the handleStateRequest things better to leave it as serverRoom.

PR #429, opened 2022-07-27
partial state tests: factor out room creation

richvdh:
Factor the creation of the server and the room from `beginPartialStateJoin`. It was handy there, but is too restrictive for tests that need to mess about with the room before the join.

~~Based on #418~~


squahtx (approved):
The last two commits look reasonable!

PR #430, opened 2022-07-28
Faster joins: test that events stay rejected

richvdh:
Regression test for https://github.com/matrix-org/synapse/issues/13411


squahtx (approved):
Looks good!

The extra event to make `testReceiveEvent...` work is a little sad. It's probably not worth the effort to eliminate right now.

PR #432, opened 2022-08-02
Add test ensuring that /make_join and /send_join are rejected during a partial join

reivilibre:
Lines up with https://github.com/matrix-org/synapse/pull/13416

### DIFF
--- a/internal/federation/handle.go
+++ b/internal/federation/handle.go
@@ -39,9 +39,9 @@ func MakeJoinRequestsHandler(s *Server, w http.ResponseWriter, req *http.Request
 	}
 
 	makeJoinResp, err := MakeRespMakeJoin(s, room, userID)
-	if err != "" {
+	if err != nil {
 		w.WriteHeader(500)
-		w.Write([]byte(err))
+		w.Write([]byte(fmt.Sprintf("complement: HandleMakeSendJoinRequests %s", err)))
 		return
 	}
 
@@ -57,7 +57,7 @@ func MakeJoinRequestsHandler(s *Server, w http.ResponseWriter, req *http.Request
 
 // MakeRespMakeJoin makes the response for a /make_join request, without verifying any signatures
 // or dealing with HTTP responses itself.
-func MakeRespMakeJoin(s *Server, room *ServerRoom, userID string) (resp gomatrixserverlib.RespMakeJoin, err string) {
+func MakeRespMakeJoin(s *Server, room *ServerRoom, userID string) (resp gomatrixserverlib.RespMakeJoin, err error) {
 	// Generate a join event
 	builder := gomatrixserverlib.EventBuilder{
 		Sender:     userID,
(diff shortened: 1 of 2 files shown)


internal/federation/handle.go
```diff
@@ -38,37 +38,52 @@ func MakeJoinRequestsHandler(s *Server, w http.ResponseWriter, req *http.Request
...
+		w.WriteHeader(500)
+		w.Write([]byte(err))
+		return
+	}
+
+	// Send it
+	res := map[string]interface{}{
+		"event":        makeJoinResp.JoinEvent,
+		"room_version": makeJoinResp.RoomVersion,
+	}
```
richvdh, line 52: pretty sure you can just `json.Marshall(makeJoinResp)` rather than copying to this intermediate map

tests/federation_room_join_partial_state_test.go
```diff
@@ -806,6 +806,128 @@ func TestPartialStateJoin(t *testing.T) {
 			},
 		)
 	})
+
+	// when the server is in the middle of a partial state join, it should not accept
+	// /make_join because it can't give a full answer.
+	t.Run("DoesntAnswerMakeJoinDuringPartialJoin", func(t *testing.T) {
```
richvdh, line 812: seems like we should use spaces here. Also s/DoesntAnswer/Rejects/ ?

tests/federation_room_join_partial_state_test.go
```diff
@@ -806,6 +806,128 @@ func TestPartialStateJoin(t *testing.T) {
...
+		//   testServer1 (a Complement test server) with @charlie:<server name>
+		//     This is the server that will create the room originally.
+		//   testServer2 (another Complement test server) with @daniel:<server name>
+		//     This is the server that will try to join the room via hs2,
+		//     but only after using hs1 to /make_join (as otherwise we have no way
+		//     of being able to build a request to /send_join)
+		//
+		// We use a blueprint with two servers rather than using two test servers
+		// because Complement test servers can't send requests to each other
+		// (their names only resolve within docker containers)
```
richvdh, line 876: Looks like this comment is outdated?

richvdh (changes requested):

clokep:
Do we need similar tests for knocking?

internal/federation/handle.go
reivilibre: Thank you, that worked.

reivilibre:
> Do we need similar tests for knocking?

good point. I'll defer this to another PR because this one is already growing enough. Once this one is done, doing one for knocking should likely be easy to get approved.

internal/federation/handle.go
```diff
@@ -38,37 +38,52 @@ func MakeJoinRequestsHandler(s *Server, w http.ResponseWriter, req *http.Request
...
 		PrevEvents: []string{room.Timeline[len(room.Timeline)-1].EventID()},
 		Depth:      room.Timeline[len(room.Timeline)-1].Depth() + 1,
 	}
-	err := builder.SetContent(map[string]interface{}{"membership": gomatrixserverlib.Join})
-	if err != nil {
-		w.WriteHeader(500)
-		w.Write([]byte("complement: HandleMakeSendJoinRequests make_join cannot set membership content: " + err.Error()))
+	err2 := builder.SetContent(map[string]interface{}{"membership": gomatrixserverlib.Join})
+	if err2 != nil {
+		err = "complement: HandleMakeSendJoinRequests make_join cannot set membership content: " + err2.Error()
```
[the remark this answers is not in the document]
reivilibre: Thanks, good tool to have in the toolbox — I was really struggling with this part and knew that `err2` had to be a silly workaround.

Let me know if you spot any other improvements I should make here

richvdh (approved):
lgtm

PR #434, opened 2022-08-04
Improve error handling when spinning up homeservers

richvdh:
Some minor tweaks to the error handling here

### DIFF
--- a/internal/docker/deployer.go
+++ b/internal/docker/deployer.go
@@ -431,7 +431,7 @@ func waitForContainer(ctx context.Context, docker *client.Client, hsDep *Homeser
 		iterCount += 1
 		if time.Now().After(stopTime) {
 			lastErr = fmt.Errorf("timed out checking for homeserver to be up: %s", lastErr)
-			break
+			return
 		}
 		inspect, err := docker.ContainerInspect(ctx, hsDep.ContainerID)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


internal/docker/deployer.go
```diff
@@ -423,17 +423,15 @@ func waitForPorts(ctx context.Context, docker *client.Client, containerID string
 }
 
 // Waits until a homeserver deployment is ready to serve requests.
-func waitForContainer(ctx context.Context, docker *client.Client, hsDep *HomeserverDeployment, stopTime time.Time) (iterCount int, err error) {
-	var lastErr error = nil
-
+func waitForContainer(ctx context.Context, docker *client.Client, hsDep *HomeserverDeployment, stopTime time.Time) (iterCount int, lastErr error) {
```
H-Shay, line 426: I am not great with Go so forgive my questions:
Does the function signature need to change now that it is no longer returning `iterCount` and `lastErr`? 
Also, it looks like `iterCount` is just incremented and doesn't do anything else, if it's not being returned anymore does it make sense to just ditch that variable?

kegsay: This function is using naked return values and named return args. See https://go.dev/tour/basics/7

So it is returning data, and by calling it `lastErr` you remove the need to ass `err = lastErr`.

kegsay (approved):
LGTM!

internal/docker/deployer.go
richvdh: yup as kegsay says: we *are* still returning `iterCount` and `lastErr`, it's just happening automatically instead of explicitly.

https://go.dev/doc/effective_go#named-results explains this too.

(This is one of many things about Go I'm not entirely convinced about, but in some cases like this it does seem to be useful)

H-Shay: Thanks all for the explanations, I appreciate it.

PR #437, opened 2022-08-09
read receipts: ensure that read receipts show up via /sync

sumnerevans:


tests/csapi/apidoc_room_receipts_test.go
```diff
@@ -5,56 +5,72 @@ import (
...
+		Content: map[string]interface{}{
+			"msgtype": "m.text",
+			"body":    "Hello world!",
+		},
+	})
+
+	return roomID, eventID
+}
+
+func syncHasReadReceipt(roomID string) client.SyncCheckOpt {
```
kegsay, line 30: It would be nice if this took an event ID to check which event you want a receipt for.

kegsay (approved):
LGTM!

PR #438, opened 2022-08-10
Add tests to check that `/login` creates a `device_lists.changed` entry

S7evinK:
Adds a test to verify that a new login creates a `device_lists.changed` entry and that uploading OTKs works as expected.


tests/csapi/keychanges_test.go
```diff
@@ -0,0 +1,111 @@
...
+func TestKeyChangesLocal(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+	password := "$uperSecretPassword"
+	bob := deployment.RegisterUser(t, "hs1", "bob", password, false)
+	unauthedClient := deployment.Client(t, "hs1", "")
+
+	t.Run("Parallel", func(t *testing.T) {
```
kegsay, line 26: Given we only run a single test here, drop this?

tests/csapi/keychanges_test.go
```diff
@@ -0,0 +1,111 @@
...
+		t.Run("New login should create a device_lists.changed entry", func(t *testing.T) {
+			mustUploadKeys(t, bob)
+
+			roomID := alice.CreateRoom(t, map[string]interface{}{"preset": "private_chat"})
+			alice.InviteRoom(t, roomID, bob.UserID)
```
kegsay, line 31: Should also work with a public chat so you can skip the invite? Would be fractionally faster.

kegsay (approved):

PR #441, opened 2022-08-10
Wait for received events using /sync in faster room joins tests

squahtx:
https://github.com/matrix-org/synapse/pull/13477 unblocks lazy-loading
`/sync`s while a room has partial state, which allows us to wait for
received events using `/sync`.

"Resync completes even when events arrive before their prev_events" is
now the only test that waits for events using `/event`, since the
outliers do not appear in the `/sync` timeline.

### DIFF
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -1001,19 +1001,26 @@ func testReceiveEventDuringPartialStateJoin(
 	must.CheckOffAll(t, gotState, expectedState)
 }
 
-// awaitEventArrival waits for alice to be able to see a given event
-func awaitEventArrival(t *testing.T, timeout time.Duration, alice *client.CSAPI, roomID string, eventID string) {
-	/* TODO: check that a lazy-loading sync can see the event. Currently this doesn't work, because /sync blocks.
-	 * https://github.com/matrix-org/synapse/issues/13146
-	alice.MustSyncUntil(t,
+// awaitEventViaSync waits for alice to be able to see a given event via an incremental lazy-loading
+// /sync and returns the new sync token after
+func awaitEventViaSync(t *testing.T, alice *client.CSAPI, roomID string, eventID string, syncToken string) string {
+	// check that a lazy-loading sync can see the event
+	syncToken = alice.MustSyncUntil(t,
 		client.SyncReq{
+			Since:  syncToken,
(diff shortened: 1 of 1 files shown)


squahtx:
> "Resync completes even when events arrive before their prev_events" is
> now the only test that waits for events using `/event`, since the
> outliers do not appear in the `/sync` timeline.

@richvdh Is this expected behaviour?

kegsay (approved):
LGTM - though @richvdh is probably better qualified to review this.

erikjohnston:
@squahtx is this just waiting for @richvdh's input now?

richvdh:
> > "Resync completes even when events arrive before their prev_events" is
> > now the only test that waits for events using `/event`, since the
> > outliers do not appear in the `/sync` timeline.
> 
> @richvdh Is this expected behaviour?

It's expected that outliers don't come down `/sync`, if that's what you're asking?

tests/federation_room_join_partial_state_test.go
```diff
@@ -65,6 +65,17 @@ func TestPartialStateJoin(t *testing.T) {
 		return serverRoom
 	}
 
+	// getSyncToken gets the latest sync token
+	getSyncToken := func(t *testing.T, alice *client.CSAPI) string {
+		_, syncToken := alice.MustSync(t,
+			client.SyncReq{
+				Filter:        buildLazyLoadingSyncFilter(nil),
```
richvdh, line 72: I think you probably don't need a `Filter`?

richvdh (approved):
lgtm!

tests/federation_room_join_partial_state_test.go
squahtx: Won't that make the function block if we decide to use it midway through a partial join?

richvdh: yes, but I didn't think this was used in the middle of a partial-state join?

(indeed, I thought initialsyncs in the middle of partial-state joins blocked irrespective of the lazy-loading flag. I'm probably wrong though)

squahtx: It isn't currently used in the middle of a partial-state join, but I would like to keep the option open.

lazy-loading initial syncs don't (or no longer?) block during a partial-state join, otherwise one of the new lazy-loading tests in #442 would be getting stuck.

richvdh: all makes sense, thanks!

PR #442, opened 2022-08-10
Add lazy-loading `/sync` tests for faster room joins

squahtx:
Test that lazy-loading `/sync`s do not block when receiving events from
remote senders. Also test that lazy-loading `/sync`s return the
appropriate memberships for event senders.

---

Contains some changes from #440 and #441.


squahtx:
The three new tests have a lot of duplicate code. Suggestions for tidying them up are welcome.

tests/federation_room_join_partial_state_test.go
```diff
@@ -337,6 +337,155 @@ func TestPartialStateJoin(t *testing.T) {
...
+					"limit": 1,
+				}),
+			},
+		)
+
+		err := client.SyncTimelineHasEventID(serverRoom.RoomID, event1.EventID())(alice.UserID, syncRes)
+		if err == nil {
+			t.Errorf("gappy /sync returned the first event unexpectedly")
+		}
+
```
erikjohnston, line 429: I wonder if we should also assert that the `limited` flag is set? I don't think it matters much for this test, but in general I think it might avoid potential pitfalls to explicitly assert that.

erikjohnston (approved):

tests/federation_room_join_partial_state_test.go
squahtx: I've added a check for the `limited` flag.

squahtx:
I've removed the objectionable parts of #440.

@kegsay Can you check the changes to `internal/client/client.go`?

erikjohnston (approved):

PR #444, opened 2022-08-11
Add E2EE Client

S7evinK:
This PR adds `CSAPIEncrypted`, which uses [mautrix-go](https://github.com/mautrix/go) to handle sending/receiving encrypted events.


internal/client/client_encrypted.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+		if err != nil {
+			resp.Error = err
+		}
+		enc.encChan <- resp
+	}
+}
+
+// Sync loops to keep syncing the client with the homeserver by calling the /sync endpoint.
+func (enc *CSAPIEncrypted) Sync(t *testing.T) {
```
kegsay, line 168: I don't see this looping, it just does it once?

internal/client/state_store.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	_, ok := room.State[event.StateEncryption]
+	return ok
+}
+
+// FindSharedRooms returns a list of room IDs that the given user ID is also a member of.
+func (ss *StateStore) FindSharedRooms(userID id.UserID) []id.RoomID {
+	sharedRooms := make([]id.RoomID, 0)
+	for roomID, room := range ss.Storer.Rooms {
+		if room.GetMembershipState(userID) != event.MembershipLeave {
```
kegsay, line 46: What about invite? That isn't shared surely. Same for Ban. Only Join should apply here.

tests/csapi/encryption_test.go
```diff
@@ -0,0 +1,95 @@
...
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+	bob := deployment.Client(t, "hs1", "@bob:hs1")
+
+	encAlice := client.MustCreateEncryptedClient(t, alice)
+	encBob := client.MustCreateEncryptedClient(t, bob)
+
+	// Create a new room and invite Bob
+	roomResp, err := encAlice.Client.CreateRoom(&mautrix.ReqCreateRoom{
```
kegsay, line 27: Please try to use the Complement version where possible, we don't want to have to maintain 2 `CreateRoom` flavours in tests.

tests/csapi/encryption_test.go
```diff
@@ -0,0 +1,95 @@
...
+	encBob.MustSyncUntil(t, client.SyncReq{}, client.SyncTimelineHasEventID(roomID, string(resp.EventID)))
+
+	time.Sleep(time.Second)
+
+	// Send an encrypted message using mautrix.Client
+	wantMessage := &event.MessageEventContent{
+		MsgType: event.MsgText,
+		Body:    "Hello world!",
+	}
+	// This sends an encrypted message
```
kegsay, line 65: Redundant comment

tests/csapi/encryption_test.go
```diff
@@ -0,0 +1,95 @@
...
+	sendMessageResp, err := encAlice.SendEncryptedMessage(roomResp.RoomID, wantMessage)
+	if err != nil {
+		t.Fatalf("failed to send message event: %s", err)
+	}
+	time.Sleep(time.Second)
```
kegsay, line 70: Do not sleep. If you absolutely have to, it MUST have justifications.

kegsay (changes requested):
It looks to be a good start but there a few things I'd like to see architecture wise.
- Use Complement `CSAPI` and not go-mautrix's client. Don't even make that public.
- Likewise, hide _all_ the mautrix bits from the test writer: they shouldn't need to know or care that we are using mautrix behind the scenes. Don't make it part of our public test API.
- It's hard to know if the message is in fact encrypted after it has been sent, we're relying a lot on go-mautrix and a lot less on manual verification (e.g the `m.room.encrypted` fields on https://spec.matrix.org/latest/client-server-api/#mroomencrypted )
- I dislike the need to start a state machine (and needing to /sync to pull in the few events we need to send encrpyted messages) - surely this can be replaced with 2x CSAPI calls (the `m.room.encryption` event and then the joined members in the room?) It would be much nicer _and_ more reliable if we can just poll that information just prior to using it.

tests/csapi/encryption_test.go
```diff
@@ -0,0 +1,63 @@
...
+	})
+
+	// Send an encrypted text message
+	msgType := "m.text"
+	wantBody := "Hello World!"
+	encAlice.SendEncryptedMessageSynced(t, roomID, msgType, wantBody)
+
+	// Wait for an encrypted event with Bob
+	t.Log("Waiting for encrypted event")
+	msg := encBob.MustSyncUntilEncryptedMessage(t)
```
kegsay, line 56: Weird API shape. This is going to return on _every_ encrypted message and not ones for this room? Also weird that this is returning the text "Hello World!" and _not_ the `content` section which is what is encrypted? How would you test that the `msgType` was set correctly?

tests/federation_encryption_test.go
```diff
@@ -0,0 +1,63 @@
...
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+	bob := deployment.Client(t, "hs2", "@bob:hs2")
+
+	encAlice := client.MustCreateEncryptedClient(t, alice)
+	encBob := client.MustCreateEncryptedClient(t, bob)
+
+	// start syncing using mautrix.Client, needed to update the olmMachine store
+	go encAlice.Sync(t)
+	go encBob.Sync(t)
+	defer encAlice.StopSync()
```
kegsay, line 26: I really dislike this from an API perspective because it's so easy to forget to do this (either sync, sync in a goroutine, defer stop on _both_ clients, etc). It would be better if we encapsulated this in `MustCreateEncryptedClient` where we just return `encAlice, finish` or something and then just `defer finish()`, which is similar to the usage of `federation.Server`:
```go
	cancel := srv.Listen()
	defer cancel()
```

tests/federation_encryption_test.go
```diff
@@ -0,0 +1,63 @@
...
+		Sender:   encAlice.UserID,
+		Content: map[string]interface{}{
+			"algorithm": "m.megolm.v1.aes-sha2",
+		},
+	})
+
+	// Send an encrypted text message
+	msgType := "m.text"
+	wantBody := "Hello World!"
+	encAlice.SendEncryptedMessageSynced(t, roomID, msgType, wantBody)
```
kegsay, line 52: This hides too much complexity. It's nice that we have a one-liner to send encrypted messages. Let's keep that. _However_, it is hard to see how we can test E2EE more thoroughly with this, as things like uploading/downloading OTKs, syncing device lists, etc is completely hidden from the user. Ideally there would be escape hatches where you can choose to do more complex things, but by default it Just Works.

internal/client/client.go
```diff
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"github.com/matrix-org/gomatrixserverlib"
+	_ "github.com/mattn/go-sqlite3"
```
kegsay, line 22: We are using our own state store, so we can nuke this right?

internal/client/client_encrypted.go
```diff
@@ -0,0 +1,233 @@
...
+	stateStore := StateStore{storer: mautrix.NewInMemoryStore()}
+
+	// Open SQL connection
+	db, err := sql.Open("sqlite3", ":memory:")
+	if err != nil {
+		t.Fatalf("failed to open sql.DB: %s", err)
+	}
+
+	accountID := csapi.UserID + "-" + csapi.DeviceID
+	sqlCryptoStore := crypto.NewSQLCryptoStore(db, "sqlite3", accountID, id.DeviceID(csapi.DeviceID), []byte(csapi.DeviceID+"pickle"), defaultOptions.logger)
```
kegsay, line 74: Having spoken to tulir, we can use `NewGobStore` safely so long as we don't write to disk as you can lose data if the write fails. Please can we use that but yank out the save() code, so either copy https://github.com/mautrix/go/blob/v0.12.1/crypto/store.go#L354 and comment out the `save()` code or write to `/dev/null`. I'd much rather use an in-memory impl than pull in a SQLite dep and add more moving parts.

internal/client/client_encrypted.go
```diff
@@ -0,0 +1,233 @@
...
+	// Handle general events, e.g. to update the olmMachine and stateStore
+	syncer.OnSync(encAPI.syncCallback(t))
+
+	// Handle membership events
+	syncer.OnEventType(event.StateMember, func(_ mautrix.EventSource, evt *event.Event) {
+		encAPI.olmMachine.HandleMemberEvent(evt)
+	})
+
+	// Handle encrypted events
+	syncer.OnEventType(event.EventEncrypted, encAPI.encryptedEventHandler())
```
kegsay, line 106: This should almost certainly be done using Complement's API and not mautrix's syncer with a channel. Expose a public function on `CSAPIEncrypted` to `DecryptEvent(...)` which then invokes the handler code.

internal/client/client_encrypted.go
```diff
@@ -0,0 +1,233 @@
...
+		return crypto.AcceptRequest, encAPI
+	}*/
+
+	syncer := cl.Syncer.(*mautrix.DefaultSyncer)
+	// Handle general events, e.g. to update the olmMachine and stateStore
+	syncer.OnSync(encAPI.syncCallback(t))
+
+	// Handle membership events
+	syncer.OnEventType(event.StateMember, func(_ mautrix.EventSource, evt *event.Event) {
+		encAPI.olmMachine.HandleMemberEvent(evt)
```
kegsay, line 102: We need to be treating this in the same way that we treat the federation server API wise. In other words, by default do _nothing fancy_. Expose optional _opt-in_ mechanics for "automatically keeping member lists up to date" and so on.

kegsay (commented):
This is looking much better, but can be even better with a bit more time. I'm happy to contribute to this as well when I get some time, if you don't address my comments before then. This is looking promising!

kegsay:
Closing as stale.

PR #451, opened 2022-08-18
[server notices] wait for the join to come down sync first

S7evinK:
Not exactly sure if we should fix the issue here or in Dendrite:
In Dendrite, the server notice event might be persisted before the join event, but this results in Dendrite missing the server notice event, because the initial sync sets the stream position to something > the server notice event. So in further incremental syncs we wouldn't receive the server notice event.

This now waits for the join to come down sync and does another initial sync, deflaking the test in Dendrite.


S7evinK:
Should now be fixed in Dendrite - https://github.com/matrix-org/dendrite/pull/2657

PR #452, opened 2022-08-18
Fix partial state join tests for incorrectly accepted/rejected state

squahtx:
These broke during the merge of #441.


richvdh (approved):
lgtm

PR #453, opened 2022-08-22
Add a test testing that an `m.login.password` flow is available on `/deactivate`

reivilibre:
Would have caught https://github.com/matrix-org/synapse/issues/13560 before it was too late.

I intend to also copy this test for other endpoints that use user-interactive authentication.


tests/csapi/account_deactivate_test.go
```diff
@@ -16,6 +20,41 @@ func TestDeactivateAccount(t *testing.T) {
...
+
+	// Ensure that the first step, in which the client queries the server's user-interactive auth flows, returns
+	// at least one auth flow involving a password.
+	t.Run("Password flow is available", func(t *testing.T) {
+		reqBody := client.WithJSONBody(t, map[string]interface{}{})
+		res := authedClient.DoFunc(t, "POST", []string{"_matrix", "client", "v3", "account", "deactivate"}, reqBody)
+		must.MatchResponse(t, res, match.HTTPResponse{
+			StatusCode: 401,
+			JSON: []match.JSON{
+				match.JSONArrayEach(
```
reivilibre, line 32: This feels awful. I'm pretty sure these matchers should be able to stack recursively, but I'm not sure how to — the matcher seems to want a function rather than another matcher as its argument.

tests/csapi/account_deactivate_test.go
```diff
@@ -16,6 +20,41 @@ func TestDeactivateAccount(t *testing.T) {
...
+								foundStage = true
+							}
+						}
+						if !foundStage {
+							return fmt.Errorf("No m.login.password stage found.")
+						}
+						return nil
+					},
+				),
+			},
```
DMRobertson, line 54: I think this is going to test that every flow contains an `m.login.password` stage.

I think we instead want to test that at least one flow contains an `m.login.password` stage?

DMRobertson: Though having said that: must a homeserver offer password-based authentication at all? I'm not sure that's a requirement of the spec!

Maybe it'd be easier to assert that
- we get a response object containing at least one flow
- each flow consists of >= 1 stages

Would that have detected the Synapse screwup? (I think even the status code check would suffice, but...)

tests/csapi/account_deactivate_test.go
```diff
@@ -16,6 +18,54 @@ func TestDeactivateAccount(t *testing.T) {
...
+			}
+
+			for _, stage := range stageList {
+				stageName, ok := stage.(string)
+				if !ok {
+					t.Fatalf("stage is not a string")
+					return
+				}
+				if stageName == "m.login.password" {
+					foundPasswordStage = true
```
erikjohnston, line 59: We should break from the outer loop here?

erikjohnston (approved):

PR #443, opened 2022-08-10
Draft: Federation `/messages` tests

MadLittleMods:
Federation `/messages` tests

Part of https://github.com/matrix-org/synapse/issues/13356 and https://github.com/matrix-org/synapse/pull/13440

## Dev notes

```
TEST_ONLY_IGNORE_POETRY_LOCKFILE=1 TEST_ONLY_SKIP_DEP_HASH_VERIFICATION=1 COMPLEMENT_DEBUG=1 COMPLEMENT_KEEP_BLUEPRINTS="fed.perf_many_messages.hs1" COMPLEMENT_ALWAYS_PRINT_SERVER_LOGS=1 COMPLEMENT_DIR=../complement ./scripts-dev/complement.sh -run TestMessagesOverFederation -p 1
```

```
TEST_ONLY_IGNORE_POETRY_LOCKFILE=1 TEST_ONLY_SKIP_DEP_HASH_VERIFICATION=1 COMPLEMENT_KEEP_BLUEPRINTS="perf_many_messages" COMPLEMENT_ALWAYS_PRINT_SERVER_LOGS=1 COMPLEMENT_DIR=../complement ./scripts-dev/complement.sh -run TestMessagesOverFederation -p 1
```

PR #454, opened 2022-08-22
Keep specific homeserver from blueprint

MadLittleMods:
Keep specific homeserver from blueprint. When testing in https://github.com/matrix-org/complement/pull/443, `hs1` has a room with many users with many messages in it which takes a while to construct in the blueprint which I want to stay cached, and `hs2` is my local Synapse built up over and over to test optimization changes when federating with `hs1`.

Split off from https://github.com/matrix-org/complement/pull/443

Usage:

```
COMPLEMENT_KEEP_BLUEPRINTS="fed.perf_many_messages.hs1" COMPLEMENT_DIR=../complement ./scripts-dev/complement.sh -run TestMessagesOverFederation -p 1
```

### Dev notes

Labels on container:

```
"Labels": {
    "access_token_@charlie:hs2": "syt_Y2hhcmxpZQ_rKpoAAPGqeKxFYkgmAkU_4PG5Hd",
    "complement_blueprint": "perf_many_messages",
    "complement_context": "fed.perf_many_messages.hs2",
    "complement_hs_name": "hs2",
    "complement_pkg": "fed",
    "desktop.docker.io/ports/8008/tcp": "127.0.0.1:0",
    "desktop.docker.io/ports/8448/tcp": "127.0.0.1:0",
    "device_id@charlie:hs2": "NINFGRNLCL",
    "org.opencontainers.image.documentation": "https://github.com/matrix-org/synapse/blob/master/docker/README.md",
    "org.opencontainers.image.licenses": "Apache-2.0",
    "org.opencontainers.image.source": "https://github.com/matrix-org/synapse.git",
    "org.opencontainers.image.url": "https://matrix.org/docs/projects/server/synapse"
}
```

### DIFF
--- a/cmd/account-snapshot/README.md
+++ b/cmd/account-snapshot/README.md
@@ -23,7 +23,7 @@ Then run Homerunner in single-shot mode: (this will take hours or days depending
 ```
 HOMERUNNER_SNAPSHOT_BLUEPRINT=blueprint.json ./homerunner
 ```
-This will execute the blueprint and commit the resulting images so you can push them to docker hub/gitlab. IMPORTANT: Make sure to set `HOMERUNNER_KEEP_BLUEPRINTS=your-blueprint-name` when running homerunner subsequently or **it will clean up the image**.
+This will execute the blueprint and commit the resulting images so you can push them to docker hub/gitlab. IMPORTANT: Make sure to set `HOMERUNNER_KEEP_BLUEPRINTS=your-blueprint-name` (TODO: Update) when running homerunner subsequently or **it will clean up the image**.
 
 #### Limitations
 
--- a/internal/docker/builder.go
+++ b/internal/docker/builder.go
@@ -124,16 +124,16 @@ func (d *Builder) removeImages() error {
 			d.log("Not cleaning up image with tags: %v", img.RepoTags)
 			continue
 		}
-		bprintName := img.Labels["complement_blueprint"]
+		contextStr := img.Labels[complementLabel]
 		keep := false
 		for _, keepBprint := range d.Config.KeepBlueprints {
-			if bprintName == keepBprint {
+			if contextStr == keepBprint {
 				keep = true
 				break
 			}
 		}
 		if keep {
-			d.log("Keeping image created from blueprint %s", bprintName)
+			d.log("Keeping image created from blueprint %s", contextStr)
 			continue
 		}
 		_, err = d.Docker.ImageRemove(context.Background(), img.ID, types.ImageRemoveOptions{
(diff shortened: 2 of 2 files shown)


internal/docker/builder.go
```diff
@@ -180,17 +180,32 @@ func (d *Builder) ConstructBlueprintIfNotExist(bprint b.Blueprint) error {
 	if err != nil {
 		return fmt.Errorf("ConstructBlueprintIfNotExist(%s): failed to ImageList: %w", bprint.Name, err)
 	}
-	if len(images) == 0 {
-		err = d.ConstructBlueprint(bprint)
+	var missingHomeservers []b.Homeserver
```
MadLittleMods, line 183: There are many ways to tackle this sort of behavior. What's the most idiomatic Complement way to accomplish this? Different argument name?

internal/docker/builder.go
```diff
@@ -124,16 +124,16 @@ func (d *Builder) removeImages() error {
 			d.log("Not cleaning up image with tags: %v", img.RepoTags)
 			continue
 		}
-		bprintName := img.Labels["complement_blueprint"]
+		contextStr := img.Labels[complementLabel]
```
MadLittleMods, line 127: Re-using the `contextStr` because it already packs in the blueprint name and homeserver name like `fed.perf_many_messages.hs1`

internal/docker/builder.go
```diff
@@ -124,16 +124,16 @@ func (d *Builder) removeImages() error {
...
 		keep := false
 		for _, keepBprint := range d.Config.KeepBlueprints {
-			if bprintName == keepBprint {
+			if contextStr == keepBprint {
```
MadLittleMods, line 130: We could add in some regex pattern matching here to support keeping all of the homeservers in a blueprint, `fed.perf_many_messages.*`

cmd/account-snapshot/README.md
```diff
@@ -23,7 +23,7 @@ Then run Homerunner in single-shot mode: (this will take hours or days depending
 ```
 HOMERUNNER_SNAPSHOT_BLUEPRINT=blueprint.json ./homerunner
 ```
-This will execute the blueprint and commit the resulting images so you can push them to docker hub/gitlab. IMPORTANT: Make sure to set `HOMERUNNER_KEEP_BLUEPRINTS=your-blueprint-name` when running homerunner subsequently or **it will clean up the image**.
+This will execute the blueprint and commit the resulting images so you can push them to docker hub/gitlab. IMPORTANT: Make sure to set `HOMERUNNER_KEEP_BLUEPRINTS=your-blueprint-name` (TODO: Update) when running homerunner subsequently or **it will clean up the image**.
```
MadLittleMods, line 26: Keeping this as a TODO to update until we settle on a way to accomplish this.

kegsay (changes requested):
Overall happy with the idea. We need to be kinder when breaking API compatibility however.

internal/docker/builder.go
MadLittleMods: This other comment is perfect, https://github.com/matrix-org/complement/pull/454#discussion_r975202948

Just looking for a way forward as I know the the quick and dirty way here wasn't good enough to merge.

kegsay:
What's the status with this?

PR #456, opened 2022-08-26
Hack around a race in the partial-state tests

richvdh:
Fixes https://github.com/matrix-org/synapse/issues/13564


squahtx (approved):
lgtm! it's great to have the flakiness fixed

PR #457, opened 2022-08-29
Tests to make sure right event is returned when all of the events have the same timestamps - MSC3030

MadLittleMods:
Tests to make sure right event is returned when all of the events have the same timestamps.

Synapse changes: https://github.com/matrix-org/synapse/pull/13658

Part of https://github.com/matrix-org/matrix-spec-proposals/pull/3030

### DIFF
--- a/tests/msc3030_test.go
+++ b/tests/msc3030_test.go
@@ -63,7 +63,7 @@ func TestJumpToDateEndpoint(t *testing.T) {
 			mustCheckEventisReturnedForTime(t, alice, roomID, eventB.AfterTimestamp, "f", "")
 		})
 
-		t.Run("should find next event topologically before given timestmap when all message timestamps are the same", func(t *testing.T) {
+		t.Run("should find next event topologically before given timestamp when all message timestamps are the same", func(t *testing.T) {
 			t.Parallel()
 			roomID, _, _ := createTestRoom(t, alice)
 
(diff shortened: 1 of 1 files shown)


tests/msc3030_test.go
```diff
@@ -63,6 +63,45 @@ func TestJumpToDateEndpoint(t *testing.T) {
...
+			as.JoinRoom(t, roomID, []string{"hs1"})
+
+			// Send a couple messages with the same timestamp after the other test
+			// messages in the room.
+			timeBeforeMessageCreation := time.Now()
+			sendMessageWithTimestamp(t, as, alice, roomID, timeBeforeMessageCreation, "messageWithSameTime1")
+			messageIDWithSameTime2 := sendMessageWithTimestamp(t, as, alice, roomID, timeBeforeMessageCreation, "messageWithSameTime2")
+
+			// Looking backwards from the time the messages were sent, we should find
+			// message2. A naive MSC3030 implementation that only sorts by timestamp,
```
squahtx, line 81:
```suggestion
			// message2. A naive MSC3030 implementation that only sorts by timestamp
```

tests/msc3030_test.go
```diff
@@ -63,6 +63,45 @@ func TestJumpToDateEndpoint(t *testing.T) {
 			mustCheckEventisReturnedForTime(t, alice, roomID, eventB.AfterTimestamp, "f", "")
 		})
 
+		t.Run("should find next event topologically before given timestmap when all message timestamps are the same", func(t *testing.T) {
```
squahtx, line 66: small typo:
```suggestion
		t.Run("should find next event topologically before given timestamp when all message timestamps are the same", func(t *testing.T) {
```

and the same for the other test too

tests/msc3030_test.go
```diff
@@ -139,28 +178,13 @@ func TestJumpToDateEndpoint(t *testing.T) {
 				timeBeforeRoomCreation := time.Now()
 				roomID, _, _ := createTestRoom(t, alice)
 
-				// Join from the application service bridge user
+				// Join from the application service bridge user so we can use to send
+				// some messages at a specific time.
```
squahtx, line 182: small typo
```suggestion
				// Join from the application service bridge user so we can use it to send
				// some messages at a specific time.
```

squahtx (approved):
lgtm!

MadLittleMods:
Thanks for the review @squahtx 🦝

PR #458, opened 2022-09-09
Fix panic in `HandleEventRequests` when room or event is not known

squahtx:

### DIFF
--- a/internal/federation/handle.go
+++ b/internal/federation/handle.go
@@ -324,6 +324,12 @@ func HandleEventRequests() func(*Server) {
 				}
 			}
 
+			if event == nil {
+				w.WriteHeader(404)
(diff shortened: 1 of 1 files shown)


internal/federation/handle.go
```diff
@@ -324,6 +324,12 @@ func HandleEventRequests() func(*Server) {
 				}
 			}
 
+			if event == nil {
+				w.WriteHeader(404)
+				w.Write([]byte(fmt.Sprintf(`complement: failed to find event: %s`, eventID)))
+				return
+			}
+
```
squahtx, line 332: The panic happens below at line 337, where we call `event.JSON()`.

neilalexander (approved):

squahtx:
Thanks for the fast review!

PR #460, opened 2022-09-10
LeaveRoom: send with empty JSON body for spec compliance

sumnerevans:
This change ensures that an empty JSON body is sent whenever `/rooms/:room_id/leave` is called.


internal/client/client.go
```diff
@@ -151,11 +151,12 @@ func (c *CSAPI) JoinRoom(t *testing.T, roomIDOrAlias string, serverNames []strin
...
 }
 
-// LeaveRoom joins the room ID, else fails the test.
+// LeaveRoom leaves the room ID, else fails the test.
 func (c *CSAPI) LeaveRoom(t *testing.T, roomID string) {
 	t.Helper()
 	// leave the room
-	c.MustDoFunc(t, "POST", []string{"_matrix", "client", "v3", "rooms", roomID, "leave"})
+    body := map[string]interface{}{}
+	c.MustDoFunc(t, "POST", []string{"_matrix", "client", "v3", "rooms", roomID, "leave"}, WithJSONBody(t, body))
```
DMRobertson, line 159: Is this meant to be indented?

DMRobertson (commented):
https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3roomsroomidleave isn't super clear about what clients should do if they want to /leave without giving a reason. Is there any text which suggest clients should send an empty body in that situation? (I would guess we ignored the body completely in spec 1.0)

internal/client/client.go
DMRobertson: Oh, I think it's a tabs versus spaces thing on line 158.

sumnerevans:
Section 1 says:

> For the default HTTP transport, all API calls use a Content-Type of application/json

DMRobertson (approved):
Many thanks!

richvdh:
The relevant synapse bug is https://github.com/matrix-org/synapse/issues/13388.

Issue #226, opened 2021-11-19
Could Complement be extended to load and run out-of-repository tests?


reivilibre:
The context here is that we 'need' some way to run integration tests for non-Matrix-standard features, such as those implemented by custom Synapse modules or maybe even Synapse-specific behaviours (if/where there is good reason).

Without knowing too much, Complement seems basically perfect for this use case, but it really wouldn't make sense to add these kinds of tests straight to Complement's repository.

It would be nice to have out-of-repository test suites (for example: alongside a Synapse module, say https://github.com/matrix-org/synapse-user-restrictions/, but you can also imagine weirder customer-specific modules).
Is Complement suitable for being extended this way? What would it take?

I'm not familiar with Go, so I can't easily suggest the right approach (I can speculate, but I'll refrain from doing so for now).

kegsay:
Yes, it is *possible* but Complement isn't quite yet there for out-of-repo tests yet because it lacks a public API. To elaborate..

Complement just runs some Go tests so it's possible to set up the test framework trivially: it's just `go test ./your-directory`. Complement *tests* call functions/structs that are inside `/internal` and that is a problem. "internal" is a special name in Go and it means imports only work if the package doing the importing is on the same path as the imported package, which won't be the case for you if you wanted to do out of repo tests.

If the directory `/internal` is renamed then it should be possible to just import packages from Complement and have it work, provided you have the bootstrapping code from `main_test.go` e.g https://github.com/matrix-org/complement/blob/master/tests/main_test.go#L30

Now that's an easy fix, but there's a reason *why* it's internal: the API surface isn't stable nor do we (in practice I) make any guarantees that it will remain so.

I can work on opening up the API to allow this use case though as it definitely seems desirable and useful. A brief checklist for this to be done:
 - All tests in `/tests` should not be referring to any imports in `/internal` e.g `"github.com/matrix-org/complement/internal/b"`
 - The  `b`,`must` and `match` packages have already been used heavily and are geared for public consumption so they can just be shifted out.
 - The remaining packages need to be audited to see how tests in practice use them. Back when the framework was set up we had very few tests, which isn't the case now. There will likely be common usage patterns which we should support, and hide implementation details much more to allow the inner workings of Complement to change without breaking all the tests.

I can start working on this over the next few days (I need to make some more quality of life improvements to Complement anyway around blacklists) but if you need a quick fix then you can just fork and rename `internal`.

reivilibre:
Thanks for your detailed response!
[...]

PR #462, opened 2022-09-14
Add a test to verify that redactions are sent out over federation without having the original event  

H-Shay:
This is a test to verify that synapse [#13813](https://github.com/matrix-org/synapse/pull/13813) works. Tests that redaction events are sent over federation even if the originating server does not have the event to be redacted. More discussion on this can be found at https://github.com/matrix-org/synapse/issues/12795.


tests/federation_redaction_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"time"
+)
+
+// test that a redaction is sent out over federation even if we don't have the original event
+func TestFederationRedactSendsWithoutEvent(t *testing.T) {
+	os.Setenv("COMPLEMENT_DEBUG", "1")
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+
+	roxy := deployment.RegisterUser(t, "hs1", "roxy", "verygoodpass", true)
```
DMRobertson, line 19: Any particular reason for using roxy over Alice from the blueprint?

tests/federation_redaction_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/matrix-org/complement/internal/federation"
+	"github.com/matrix-org/gomatrixserverlib"
+	"os"
+	"testing"
+	"time"
+)
+
+// test that a redaction is sent out over federation even if we don't have the original event
+func TestFederationRedactSendsWithoutEvent(t *testing.T) {
+	os.Setenv("COMPLEMENT_DEBUG", "1")
```
DMRobertson, line 15: I know this is a draft but I think you don't want this when we come to merge the test.

tests/federation_redaction_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// create username
+	charlie := srv.UserID("charlie")
+
+	ver := roxy.GetDefaultRoomVersion(t)
+
+	// the remote homeserver creates a public room
+	initalEvents := federation.InitialRoomEvents(ver, charlie)
+	plEvent := initalEvents[2]
+	plEvent.Content["redact"] = 0
```
DMRobertson, line 51: This confused me for a bit. Maybe a comment like "// allow anyone to redact" would help?

tests/federation_redaction_test.go
```diff
@@ -0,0 +1,87 @@
...
+	serverRoom.AddEvent(badEvent)
+
+	eventID := badEvent.EventID()
+	fullServerName := srv.ServerName()
+	eventToRedact := eventID + ":" + fullServerName
+
+	// the client sends a request to the local homeserver to send the redaction
+	roxy.SendRedaction(t, serverRoom.RoomID, b.Event{Type: wantEventType, Content: map[string]interface{}{
+		"msgtype": "m.room.redaction"},
+		Redacts: eventToRedact}, eventToRedact)
```
DMRobertson, line 75: Sanity check: what guarantees that the badEvent isn't sent by the complement HS to the HS under test? I guess `srv` won't automatically send out transactions unless we ask it to?

tests/federation_redaction_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// wait for redaction to arrive at remote homeserver
+	waiter.Wait(t, 1*time.Second)
```
DMRobertson, line 78: What's the difference between waiter.Wait and time.Sleep here?

tests/federation_redaction_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Check that the last event in the room is now the redaction
+	lastEvent := serverRoom.Timeline[len(serverRoom.Timeline)-1]
+	lastEventType := lastEvent.Type()
+	wantedType := "m.room.redaction"
+	if lastEventType != wantedType {
+		t.Fatalf("Incorrent event type, wanted m.room.redaction.")
```
DMRobertson, line 85: It's useful to include the incorrect type in this message as a clue for debugging the test. How about e.g.
```go
		t.Fatalf("Last event in room timeline has incorrect type %s, wanted %s.", lastEventType, wantedType)
```

internal/client/client.go
```diff
@@ -196,6 +196,13 @@ func (c *CSAPI) SendEventSynced(t *testing.T, roomID string, e b.Event) string {
 	return eventID
 }
 
+func (c *CSAPI) SendRedaction(t *testing.T, roomID string, e b.Event, eventID string) {
```
DMRobertson, line 199: I think this needs a docstring.

Otherwise seems reasonable to me, though @kegsay might want to check this fits with his view of which helper functions we want in complement.

tests/federation_redaction_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}
```
DMRobertson, line 86: Should we be checking that the redaction event we received is the redaction event send by roxy? To do that you'd want to make SendRedaction return the event ID of the newly-created redaction event and then compare event IDs.

tests/federation_redaction_test.go
H-Shay: I think this I wanted to have an admin account to make it easier to do redactions but ended up taking care of that in different way. I'll revert to the vanilla alice user.

tests/federation_redaction_test.go
H-Shay: D'oh! Thanks.

tests/federation_redaction_test.go
H-Shay: My probably unsatisfactory answers to this is that this test fails without the code changes/on develop (you would think that this test would still pass on develop if the badEvent is being sent by the complement HS, because the only reason the redact isn't being sent is that the original homeserver didn't have this event).  I also don't think this is occurring due to seeing the whole thing in action while extensively debugging but I recognize that's a pretty specious answer.

tests/federation_redaction_test.go
H-Shay: I believe that the `waiter.Wait` here is setting the time that the `federation.HandleTransactionRequests` funciton of `srv` waits for the desired PDU, whereas `time.Sleep` would just pause the whole test?

tests/federation_redaction_test.go
H-Shay: Sure why not? Done.

tests/federation_redaction_test.go
DMRobertson: Thanks, I'd missed the use of `waiter` on 29

H-Shay:
Irritatingly the synapse run is beset with flakes but the test in question is passing. I'm also unsure if this test is something that makes sense to merge, given that Dendrite does not currently implement this behavior? My main goal with this was a POC of synapse #13813

DMRobertson:
> I'm also unsure if this test is something that makes sense to merge, given that Dendrite does not currently implement this behavior? 

Good question. I guess the pertinent question is: is this behaviour (being able to redact something you've not seen yet) important enough for all HSes to do?

> My main goal with this was a POC of synapse #13813

Related: https://github.com/matrix-org/complement/issues/226

DMRobertson:
Discussed yesterday in a meeting.

> is this behaviour (being able to redact something you've not seen yet) important enough for all HSes to do?

Probably. It helps moderation (https://github.com/matrix-org/synapse/issues/12795). It also ensures that an event we send out locally is sent out over federation, which seems like a very sound property to uphold. (I hadn't grokked that redactions were being sent locally in #13813.)

We agreed that we'd like to land this in complement (skipping for dendrite if necessary).

DMRobertson (approved):

PR #463, opened 2022-09-15
Some changes to calls to Docker API for podman compatibility

MatMaul:
- Podman expects the network name specified when launching the container to match the name used when creating it even when the network ID is specified, the API spec is unclear about that
- Adding labels when committing a container needs to use `changes` query parameter instead of a POST json, which is unspecified
- Add a config to specify the hostname where Complement on the host can be reached when inside a container. Podman uses `host.containers.internal`

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -78,6 +78,9 @@ type Complement struct {
 	CAPrivateKey  *rsa.PrivateKey
 
 	BestEffort bool
+
+	// The hostname of Complement from the perspective of a Homeserver running inside a container
+	HostnameRunningComplement string
 }
 
 var hsRegex = regexp.MustCompile(`COMPLEMENT_BASE_IMAGE_(.+)=(.+)$`)
@@ -127,6 +130,13 @@ func NewConfigFromEnvVars(pkgNamespace, baseImageURI string) *Complement {
 	}
 	if cfg.PackageNamespace == "" {
 		panic("package namespace must be set")
+	}
+
+	HostnameRunningComplement := os.Getenv("COMPLEMENT_HOSTNAME_RUNNING_COMPLEMENT")
+	if HostnameRunningComplement != "" {
+		cfg.HostnameRunningComplement = HostnameRunningComplement
+	} else {
+		cfg.HostnameRunningComplement = "host.docker.internal"
 	}
 
 	return cfg
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -79,7 +79,11 @@ type Complement struct {
 
 	BestEffort bool
 
-	// The hostname of Complement from the perspective of a Homeserver running inside a container
+	// Name: COMPLEMENT_HOSTNAME_RUNNING_COMPLEMENT
+	// Default: host.docker.internal
+	// Description: The hostname of Complement from the perspective of a Homeserver running inside a container.
+	// This can be useful for container runtimes using another hostname to access the host from a container,
+	// like Podman that uses `host.containers.internal` instead.
 	HostnameRunningComplement string
 }
 
--- a/internal/docker/builder.go
+++ b/internal/docker/builder.go
@@ -406,6 +406,7 @@ func (d *Builder) deployBaseImage(blueprintName string, hs b.Homeserver, context
 	)
 }
 
+// Multilines label using Dockefile syntax is unsupported, let's inline \n instead
 func generateASRegistrationYaml(as b.ApplicationService) string {
 	return fmt.Sprintf("id: %s\\n", as.ID) +
 		fmt.Sprintf("hs_token: %s\\n", as.HSToken) +
@@ -421,8 +422,8 @@ func generateASRegistrationYaml(as b.ApplicationService) string {
 		"  aliases: []\\n"
 }
 
-// createNetworkIfNotExists creates a docker network and returns its id.
-// ID is guaranteed not to be empty when err == nil
+// createNetworkIfNotExists creates a docker network and returns its name.
+// Name is guaranteed not to be empty when err == nil
 func createNetworkIfNotExists(docker *client.Client, pkgNamespace, blueprintName string) (networkName string, err error) {
 	// check if a network already exists for this blueprint
 	nws, err := docker.NetworkList(context.Background(), types.NetworkListOptions{
--- a/internal/docker/builder.go
+++ b/internal/docker/builder.go
@@ -21,7 +21,6 @@ import (
 	"time"
 
 	"github.com/docker/docker/api/types"
-	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/stdcopy"
 	"github.com/docker/go-connections/nat"
@@ -32,8 +31,6 @@ import (
 )
 
 var (
-	// HostnameRunningComplement is the hostname of Complement from the perspective of a Homeserver.
-	HostnameRunningComplement = "host.docker.internal"
 	// HostnameRunningDocker is the hostname of the docker daemon from the perspective of Complement.
 	HostnameRunningDocker = "localhost"
 )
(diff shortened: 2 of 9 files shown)


internal/docker/builder.go
```diff
@@ -328,6 +328,10 @@ func (d *Builder) construct(bprint b.Blueprint) (errs []error) {
 		// Log again so we can see the timings.
 		d.log("%s: Stopped container: %s", res.contextStr, res.containerID)
 
+		for _, c := range changes {
+			d.log(c)
+		}
```
reivilibre, line 333: debug code left in? If you want to keep this, probably want to prefix the log line with something meaningful

reivilibre (approved):
Looks reasonable to me

internal/docker/builder.go
MatMaul: thanks, fixed :)

MatMaul:
It seems like the hack [here](https://github.com/matrix-org/complement/blob/main/internal/docker/deployer.go#L187) was only needed because of the fact that the network name was not the one used when launching the container.

internal/docker/builder.go
```diff
@@ -292,32 +289,32 @@ func (d *Builder) construct(bprint b.Blueprint) (errs []error) {
...
 			for k, v := range accessTokens {
-				labels["access_token_"+k] = v
+				changes = appendLabelToChanges(changes, "access_token_"+k, v)
 			}
 		}
 
 		deviceIDs := runner.DeviceIDs(res.homeserver.Name)
 		for userID, deviceID := range deviceIDs {
-			labels["device_id"+userID] = deviceID
+			changes = appendLabelToChanges(changes, "device_id"+userID, deviceID)
```
kegsay, line 311: This presumably is going to result in duplicate keys if there are two users with the same ID? Whereas previously we would just take the last value.

internal/docker/builder.go
```diff
@@ -336,9 +333,7 @@ func (d *Builder) construct(bprint b.Blueprint) (errs []error) {
 			Author:    "Complement",
 			Pause:     true,
 			Reference: "localhost/complement:" + res.contextStr,
-			Config: &container.Config{
-				Labels: labels,
-			},
+			Changes:   changes,
```
kegsay, line 336: I'd prefer it if we kept the map of `labels` and then just called `toChanges(labels) []string` or something.

internal/docker/deployer.go
```diff
@@ -184,14 +184,6 @@ func (d *Deployer) Restart(hsDep *HomeserverDeployment, cfg *config.Complement)
 		return fmt.Errorf("Restart: Failed to stop container %s: %s", hsDep.ContainerID, err)
 	}
 
-	// Remove the container from the network. If we don't do this,
-	// (re)starting the container fails with an error like
-	// "Error response from daemon: endpoint with name complement_fed_1_fed.alice.hs1_1 already exists in network complement_fed_alice".
-	err = d.Docker.NetworkDisconnect(ctx, d.networkID, hsDep.ContainerID, false)
-	if err != nil {
-		return fmt.Errorf("Restart: Failed to disconnect container %s: %s", hsDep.ContainerID, err)
-	}
```
kegsay, line 193: Why was this removed?

internal/docker/deployer.go
```diff
@@ -283,7 +275,7 @@ func deployImage(
 		Mounts:     mounts,
 	}, &network.NetworkingConfig{
 		EndpointsConfig: map[string]*network.EndpointSettings{
-			contextStr: {
+			"complement_" + pkgNamespace + "_" + blueprintName: {
```
kegsay, line 278: Why aren't we using the `contextStr` anymore?

internal/docker/labels.go
```diff
@@ -33,7 +33,7 @@ func asIDToRegistrationFromLabels(labels map[string]string) map[string]string {
 	asMap := make(map[string]string)
 	for k, v := range labels {
 		if strings.HasPrefix(k, "application_service_") {
-			asMap[strings.TrimPrefix(k, "application_service_")] = v
+			asMap[strings.TrimPrefix(k, "application_service_")] = strings.Replace(v, "\\n", "\n", -1)
```
kegsay, line 36: Why did you change the format of `\n` and then do a replace over it?

kegsay (changes requested):
In theory LGTM, but I have outstanding questions on why seemingly unrelated things were modified. The most dangerous is the blunt removal of network disconnection code, which the comments explained why it existed, and yet this PR just removes without any reason...

internal/docker/labels.go
MatMaul: Because the `change` is a Dockerfile directive and it seems like it can't be on several lines.
This allows generating this kind of change `LABEL "mylabel"="multiline\nvalue"`.

internal/docker/builder.go
MatMaul: Exact, I'll do what you suggested (use `labels` and then convert to `changes`) and this problem should be solved.

internal/docker/deployer.go
MatMaul: This needs to be the name of the network, which is not the case here.

We could also use the contextStr as the name of the network in `createNetworkIfNotExists`, or makes `createNetworkIfNotExists` returns both the ID and the network name and store both in `Deployer`.

The latter looks like the cleaner way to do it so leaning towards that.

internal/docker/deployer.go
MatMaul: Following your edit in the top comment, do you have more question on this one ?

internal/docker/deployer.go
MatMaul: Actually it seems like we can just use the name, ID is optional, at least with Podman.
I've push a change to return and use the network name instead of the ID, let's see if Docker is happy.

internal/docker/labels.go
MatMaul: I added comments in the code regarding that for clarity.

internal/docker/deployer.go
kegsay: Nope!

internal/config/config.go
```diff
@@ -78,6 +78,9 @@ type Complement struct {
 	CAPrivateKey  *rsa.PrivateKey
 
 	BestEffort bool
+
+	// The hostname of Complement from the perspective of a Homeserver running inside a container
+	HostnameRunningComplement string
```
kegsay, line 83: Please update the docstring to match other env vars in this struct. Use this format so we can automatically generate docs:
```go
// Name: COMPLEMENT_HOSTNAME_RUNNING_COMPLEMENT
// Default: host.docker.internal
// Description: Some useful description here,
// multi-line is fine.
```

Please be thorough and explain:
 - Why you might want to set this.
 - What value you might want to set this to in Podman as an example.

internal/docker/builder.go
```diff
@@ -351,11 +346,19 @@ func (d *Builder) construct(bprint b.Blueprint) (errs []error) {
 	return errs
 }
 
+func toChanges(labels map[string]string) []string {
+	changes := make([]string, 0)
```
kegsay, line 350: `var changes []string` is shorter, or `changes := make([]string, 0, len(labels))` to preallocate a capacity of the underlying array.

internal/docker/builder.go
```diff
@@ -351,11 +346,19 @@ func (d *Builder) construct(bprint b.Blueprint) (errs []error) {
...
+	for k, v := range labels {
+		changes = append(changes, fmt.Sprintf("LABEL \"%s\"=\"%s\"", k, v))
```
kegsay, line 352: Please document assumptions you're making here. Not all characters will be supported in this format, no? Which ones would be problematic?

kegsay (approved):
Otherwise LGTM

MatMaul:
@kegsay thanks for the review, I'll let you do a final round and merge. My Fedora is happy :) .

kegsay (approved):

PR #465, opened 2022-09-16
Add tests for EDUs during fast joins

MatMaul:
The typing test will fail until https://github.com/matrix-org/synapse/pull/13830 is merged.

Presence test is not testing the sync because it seems to be lost sometimes, I haven't investigated yet.
At least it doesn't fail and block the transaction like with typing.


tests/federation_room_join_partial_state_test.go
```diff
@@ -188,6 +188,249 @@ func TestPartialStateJoin(t *testing.T) {
...
+
+		content, _ := json.Marshal(map[string]interface{}{
+			"room_id": serverRoom.RoomID,
+			"user_id": derekUserId,
+			"typing":  true,
+		})
+		edu := gomatrixserverlib.EDU{
+			Type:    "m.typing",
+			Content: content,
+		}
```
DMRobertson, line 214: I was going to say "this doesn't match the spec for `m.typing` but then a wild https://github.com/matrix-org/matrix-spec/issues/1241 appeared.

tests/federation_room_join_partial_state_test.go
```diff
@@ -188,6 +188,249 @@ func TestPartialStateJoin(t *testing.T) {
...
+			client.SyncEphemeralHas(serverRoom.RoomID, func(result gjson.Result) bool {
+				if result.Get("type").Str != "m.typing" {
+					return false
+				}
+				for _, item := range result.Get("content").Get("user_ids").Array() {
+					if item.Str == derekUserId {
+						return true
+					}
+				}
+				return false
```
DMRobertson, line 229: This checks that at least one user in the list of mxids is Derek. It'd be nice to also check that no other user is typing, i.e. that the list of typing users is exactly `[derekUserId]`.

DMRobertson: There may be a way to express this more neatly using complement's `must` library or the helpers in `internal/match/json.go`. Being blunt, I don't really understand the distinction between the two and neither seems intended to be used directly with a `gjson.Result`. Asking in the complement room or picking Kegan's brains might be worth a shot.

tests/federation_room_join_partial_state_test.go
```diff
@@ -188,6 +188,249 @@ func TestPartialStateJoin(t *testing.T) {
...
+		// 	func(userID string, sync gjson.Result) error {
+		// 		for _, e := range sync.Get("presence").Get("events").Array() {
+		// 			t.Logf("lllll %s", e)
+		// 			if e.Get("sender").Str == derekUserId {
+		// 				return nil
+		// 			}
+		// 		}
+		// 		return fmt.Errorf("No presence update from %s", derekUserId)
+		// 	},
+		// )
```
DMRobertson, line 279: Do we suspect a Synapse bug here? If so, is this related to partial joins or a known defect with presence?

tests/federation_room_join_partial_state_test.go
```diff
@@ -188,6 +188,249 @@ func TestPartialStateJoin(t *testing.T) {
...
+			"user_id": derekUserId,
+			"typing":  true,
+		})
+		edu := gomatrixserverlib.EDU{
+			Type:    "m.typing",
+			Content: content,
+		}
+		psjResult.Server.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{}, []gomatrixserverlib.EDU{edu})
+
+		psjResult.FinishStateRequest()
```
squahtx, line 217: Given https://github.com/matrix-org/synapse/issues/13684, do we want to test that we _don't_ block on incoming EDUs?

eg.
```go
		event := psjResult.CreateMessageEvent(t, "charlie", nil)
		serverRoom.AddEvent(event)
		server.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{event.JSON()}, nil)
		awaitEventViaSync(t, alice, serverRoom.RoomID, event.EventID(), "")
```
ought to get stuck and fail the test if we block after the incoming EDU `/send`.

tests/federation_room_join_partial_state_test.go
MatMaul: I haven't looked yet at your pointers but I did learn since writing that code that I can use a json path in `Get`. Which is both nice and annoying since we are using `.` as a separator in types.
That plus your comment we should end up with a pretty simple equal expression here.

tests/federation_room_join_partial_state_test.go
MatMaul: Good idea indeed, I'll add that, thanks !

tests/federation_room_join_partial_state_test.go
MatMaul: I am not sure. I would suspect a bug since I vaguely remember seeing both results which is suspicious.
I am planning to have a look at this one, but haven't had the time yet. There is some tracking about my tests&fixes status for EDUs in https://github.com/matrix-org/synapse/issues/13684.

tests/federation_room_join_partial_state_test.go
DMRobertson: Done in 7331ebc

tests/federation_room_join_partial_state_test.go
```diff
@@ -188,6 +188,249 @@ func TestPartialStateJoin(t *testing.T) {
...
+		alice.MustSyncUntil(t,
```
DMRobertson, line 226: This doesn't check that Alice can see the typing notification _during_ the resync.

DMRobertson: 39c553d

tests/federation_room_join_partial_state_test.go
DMRobertson: I tweaked this a little bit in 39c553d.

I did look at `must`, but they mostly seemed to operate on byte arrays rather than `gjson.Result`s.

DMRobertson:
In the light of Sean's comment https://github.com/matrix-org/complement/pull/465#discussion_r976708066 I'd like to write a test which checks you can still see incoming messages during resync after the server receives a presence EDU.

tests/federation_room_join_partial_state_test.go
```diff
@@ -241,6 +242,296 @@ func TestPartialStateJoin(t *testing.T) {
...
+			"typing":  true,
+		})
+		edu := gomatrixserverlib.EDU{
+			Type:    "m.typing",
+			Content: content,
+		}
+		psjResult.Server.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{}, []gomatrixserverlib.EDU{edu})
+
+		// Alice should be able to see that Derek is typing (even though HS1 is resyncing).
+		alice.MustSyncUntil(t,
```
MatMaul, line 272: We should probably store the nextBatch token here and use it as a `since` in following syncs.

DMRobertson: Thanks. This is 73b2398

tests/federation_room_join_partial_state_test.go
```diff
@@ -220,6 +220,7 @@ func TestPartialStateJoin(t *testing.T) {
 
 	// we should be able to send events in the room, during the resync
 	t.Run("CanSendEventsDuringPartialStateJoin", func(t *testing.T) {
+		// See https://github.com/matrix-org/synapse/issues/12997")
```
erikjohnston, line 223:
```suggestion
		// See https://github.com/matrix-org/synapse/issues/12997
```

tests/federation_room_join_partial_state_test.go
```diff
@@ -241,6 +242,297 @@ func TestPartialStateJoin(t *testing.T) {
...
+			}),
+		)
+
+		// Alice should still be able to see incoming PDUs in the room during
+		// the resync; the earlier EDU shouldn't interfere with this.
+		// (See https://github.com/matrix-org/synapse/issues/13684)
+		event := psjResult.CreateMessageEvent(t, "charlie", nil)
+		serverRoom.AddEvent(event)
+		server.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{event.JSON()}, nil)
+		awaitEventViaSync(t, alice, serverRoom.RoomID, event.EventID(), "")
```
erikjohnston, line 294: Can we update `aliceNextBatch` here?

tests/federation_room_join_partial_state_test.go
```diff
@@ -241,6 +242,297 @@ func TestPartialStateJoin(t *testing.T) {
...
+		}
+		psjResult.Server.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{}, []gomatrixserverlib.EDU{edu})
+
+		alice.MustSyncUntil(t,
+			client.SyncReq{
+				Filter: buildLazyLoadingSyncFilter(nil),
+			},
+			func(userID string, sync gjson.Result) error {
+				for _, e := range sync.Get("presence").Get("events").Array() {
+					t.Logf("lllll %s", e)
```
erikjohnston, line 363: `lllll`?

tests/federation_room_join_partial_state_test.go
```diff
@@ -241,6 +242,297 @@ func TestPartialStateJoin(t *testing.T) {
...
+					return true
+				}
+				return false
+			}),
+		)
+		psjResult.FinishStateRequest()
+	})
+
+	// we should be able to receive device list update EDU over federation during the resync
+	t.Run("CanReceiveDeviceListUpdateDuringPartialStateJoin", func(t *testing.T) {
```
erikjohnston, line 479: I'm a bit dubious about this test TBH, I don't think it tests anything particularly useful as is and may just cause confusion.

tests/federation_room_join_partial_state_test.go
DMRobertson: yes. thanks, I hadn't noticed this.

DMRobertson: ed33411

tests/federation_room_join_partial_state_test.go
DMRobertson: errr, debug. you saw nothing.

DMRobertson: 21646b5

tests/federation_room_join_partial_state_test.go
DMRobertson: I don't fully follow this either. Shouldn't the device list update be send to Alice, even if she and Derek aren't in an encrypted room?

DMRobertson: As in: Alice should get told via /sync that Derek's device list has changed.

erikjohnston: In Synapse currently, yes, though its a bit of a grey area and we've certainly talked about only sending device list updates for E2EE enabled devices and/or users who share rooms.

OTOH, I think all the current integration tests assume this, so I don't see a problem assuming it here

Issue #351, opened 2022-03-24
Test MSC2675 / MSC3440


clokep:
Synapse has a variety of unit tests for MSC2675 / MSC3440 (and MSC2676 / MSC2677, but those are still unstable). Some of those would be good to be complement tests instead.

See https://github.com/matrix-org/synapse/blob/4df10d32148ae29f792afc68ff774bcbd1915cea/tests/rest/client/test_relations.py

clokep:
This is maybe blocked behind being able to actually test this? Which would need both matrix-org/complement#351 and matrix-org/complement#194.

PR #467, opened 2022-09-19
Add basic tests for /relations.

clokep:
Basic tests for the `/relations` endpoint from MSC2675.

Part #351.


clokep:
This is super simple, but I want to:

1. Get a base down.
2. Be sure that something lands as I'm not positive when I'll come back to this.

erikjohnston (approved):

PR #468, opened 2022-09-20
Documentation: completely rework how env vars are documented

kegsay:
Automatically generate the markdown based on line comments in `type Complement struct {...}`. Rejig `README.md` to include a link to `ENVIRONMENT.md`. Update sytest stats.

[Rendered docs](https://github.com/matrix-org/complement/blob/7110a1638ba21d45162cf454609aafede1b196ac/ENVIRONMENT.md)


cmd/gendoc/main.go
```diff
@@ -0,0 +1,132 @@
...
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"log"
+	"os"
+	"sort"
+	"strings"
+)
+
+var configPath = flag.String("config", "", "The path to internal/config/config.go")
```
S7evinK, line 15: Why not default to `internal/config/config.go`, which is probably the most used case?

S7evinK (approved):

PR #471, opened 2022-09-22
Partial State Join tests: Reuse one Alice deployment but register a different alice user for each test

reivilibre:
This improves runtime of the test file from 3m15s to 1m25s locally.

### DIFF
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -82,8 +82,7 @@ func TestPartialStateJoin(t *testing.T) {
 	// test that a regular /sync request made during a partial-state /send_join
 	// request blocks until the state is correctly synced.
 	t.Run("SyncBlocksDuringPartialStateJoin", func(t *testing.T) {
-		//userId := "@t1alice1:hs1"
-		alice := deployment.RegisterUser(t, "hs1", "t1alice", "secret", false)
+		alice := deployment.RegisterUser(t, "hs1", "t2alice", "secret", false)
 
 		server := createTestServer(t, deployment)
 		cancel := server.Listen()
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -76,12 +76,14 @@ func TestPartialStateJoin(t *testing.T) {
 		return syncToken
 	}
 
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+
 	// test that a regular /sync request made during a partial-state /send_join
 	// request blocks until the state is correctly synced.
 	t.Run("SyncBlocksDuringPartialStateJoin", func(t *testing.T) {
-		deployment := Deploy(t, b.BlueprintAlice)
-		defer deployment.Destroy(t)
-		alice := deployment.Client(t, "hs1", "@alice:hs1")
+		//userId := "@t1alice1:hs1"
+		alice := deployment.RegisterUser(t, "hs1", "t1alice", "secret", false)
 
 		server := createTestServer(t, deployment)
 		cancel := server.Listen()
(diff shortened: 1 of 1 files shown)


tests/federation_room_join_partial_state_test.go
```diff
@@ -76,12 +76,14 @@ func TestPartialStateJoin(t *testing.T) {
...
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+
 	// test that a regular /sync request made during a partial-state /send_join
 	// request blocks until the state is correctly synced.
 	t.Run("SyncBlocksDuringPartialStateJoin", func(t *testing.T) {
-		deployment := Deploy(t, b.BlueprintAlice)
-		defer deployment.Destroy(t)
-		alice := deployment.Client(t, "hs1", "@alice:hs1")
+		//userId := "@t1alice1:hs1"
```
squahtx, line 85: stray comment?

squahtx (commented):
The improvement to test runtime is absolutely brilliant. Thanks for improving it greatly!

The hardcoded incremental numbering is a bit dodgy. I can totally see us ending up with gaps, non-increasing and duplicate user IDs as various developers refactor the file later on.

We could consider doing something similar to the device list tracking tests, where we allocate incrementing user IDs (albeit in a thread-unsafe way):
https://github.com/matrix-org/complement/blob/4acae02e760ca4425e4a720f0f59445b3bbb21a9/tests/csapi/device_lists_test.go#L21-L27
It makes the user IDs less predictable for sure, but there's only one `@alice` per test so it's still reasonably unambiguous.

DMRobertson:
Is this the preferred idiom?  I don't know if having a deployment per-test is an Opinion that complement takes strongly.

(The speedup is highly attractive and I'm inclined to just merge it; but I'm wondering if this is just a workaround for "creating deployments is slow, i.e. treating symptom rather than cause)

reivilibre:
> Is this the preferred idiom? I don't know if having a deployment per-test is an Opinion that complement takes strongly.
> 
> (The speedup is highly attractive and I'm inclined to just merge it; but I'm wondering if this is just a workaround for "creating deployments is slow, i.e. treating symptom rather than cause)

FWIW this is based on a few-weeks-old suggestion from Kegan, who as far as I'm concerned may as well be *Complement-Man*. It's not the prettiest I suppose, but it seems OK since we still create different 'Complement' test homeservers and the users and rooms should be orthogonal.

> I can totally see us ending up with gaps, non-increasing and duplicate user IDs as various developers refactor the file later on.

I personally don't think gaps or non-increasing numbers matter much; sure they're a little ugly I suppose, but it means you always know what you're looking at in the logs.
Duplicate user IDs should be impossible because each test registers its own user. If they're duplicate, registration will fail with a pretty obvious error:
`{"errcode":"M_USER_IN_USE","error":"User ID already taken."}`

squahtx (approved):

tests/federation_room_join_partial_state_test.go
```diff
@@ -76,12 +76,13 @@ func TestPartialStateJoin(t *testing.T) {
...
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+
 	// test that a regular /sync request made during a partial-state /send_join
 	// request blocks until the state is correctly synced.
 	t.Run("SyncBlocksDuringPartialStateJoin", func(t *testing.T) {
-		deployment := Deploy(t, b.BlueprintAlice)
-		defer deployment.Destroy(t)
-		alice := deployment.Client(t, "hs1", "@alice:hs1")
+		alice := deployment.RegisterUser(t, "hs1", "t2alice", "secret", false)
```
clokep, line 85: `t2alice` seems duplicated, not sure if that's on purpose?

reivilibre: oh silly me, I committed it on accident to test what would happen if we did that accidentally :-)

PR #472, opened 2022-09-22
Unflake `Can query remote device keys using POST`

S7evinK:
This should hopefully unflake `Can query remote device keys using POST` on Dendrite, as we're now waiting for the initial key upload to propagate over federation first.


neilalexander (approved):

S7evinK:
Closed in favor of https://github.com/matrix-org/dendrite/pull/2733

PR #475, opened 2022-09-23
Faster room joins: Add happy path tests for device list tracking

squahtx:
---
Best reviewed commit by commit.
The first half of the commits overlap with #473.

PR #476, opened 2022-09-23
Faster room joins: Add abandoned join tests for device list tracking

squahtx:
This change tests device list tracking when a partial state room is left
before the partial state join completes. Leaving a partial state room
is not currently implemented in Synapse.

---

~~Built on top of #475.~~

### DIFF
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -2348,8 +2348,12 @@ func TestPartialStateJoin(t *testing.T) {
 			server.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{joinEvent.JSON()}, nil)
 			awaitEventViaSync(t, alice, room.RoomID, joinEvent.EventID(), syncToken)
 
-			// @elsie's device list ought to be cached after the first request.
-			// Try to bring their device list into the homeserver's cache, then test whether it really was cached.
+			// hs1 should now be tracking @elsie's device list. Enforce this in two steps:
+			// 1) Have Alice request Elsie's keys via the CS API and check
+			// that hs1 makes a federation request to serve Alice's request.
+			// 2) Repeat Alice's request and check that hs1 does _not_ make a
+			// second federation request. This proves that hs1 has cached the
+			// response from the first step.
 			syncToken = mustSyncUntilDeviceListsHas(t, alice, syncToken, "changed", server.UserID("elsie"))
 			mustQueryKeysWithFederationRequest(t, alice, userDevicesChannel, server.UserID("elsie"))
 			mustQueryKeysWithoutFederationRequest(t, alice, userDevicesChannel, server.UserID("elsie"))
@@ -2359,6 +2363,8 @@ func TestPartialStateJoin(t *testing.T) {
 			sendDeviceListUpdate("elsie")
 			mustSyncUntilDeviceListsHas(t, alice, syncToken, "changed", server.UserID("elsie"))
 			mustQueryKeysWithFederationRequest(t, alice, userDevicesChannel, server.UserID("elsie"))
+			// Again, hs1 should have cached @elsie's device list.
+			// hs1 should not require a second federation request if Alice rerequests @elsie's keys.
 			mustQueryKeysWithoutFederationRequest(t, alice, userDevicesChannel, server.UserID("elsie"))
 
 			// Finish the partial state join.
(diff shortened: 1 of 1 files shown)


tests/federation_room_join_partial_state_test.go
```diff
@@ -2405,6 +2405,77 @@ func TestPartialStateJoin(t *testing.T) {
 			mustSyncUntilDeviceListsHas(t, alice, syncToken, "left", server.UserID("elsie"))
 			mustQueryKeysWithFederationRequest(t, alice, userDevicesChannel, server.UserID("elsie"))
 		})
+
+		// test that device lists stop being tracked when leaving a partial state room before the
+		// partial state join completes.
+		t.Run("Device list no longer tracked when leaving partial state room", func(t *testing.T) {
+			t.Skip("Cannot yet leave a room during resync")
```
richvdh, line 2412: to be clear: if these are both skipped, does that mean you've not been able to try them?

It might be good to link to the bug that they are waiting for (https://github.com/matrix-org/synapse/issues/12802, I guess?)

tests/federation_room_join_partial_state_test.go
```diff
@@ -2405,6 +2405,77 @@ func TestPartialStateJoin(t *testing.T) {
...
+			room.AddEvent(joinEvent)
+			server.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{joinEvent.JSON()}, nil)
+			awaitEventViaSync(t, alice, room.RoomID, joinEvent.EventID(), syncToken)
+
+			// @elsie's device list ought to be cached.
+			syncToken = mustSyncUntilDeviceListsHas(t, alice, syncToken, "changed", server.UserID("elsie"))
+			mustQueryKeysWithFederationRequest(t, alice, userDevicesChannel, server.UserID("elsie"))
+			mustQueryKeysWithoutFederationRequest(t, alice, userDevicesChannel, server.UserID("elsie"))
+
+			// Leave the room
```
richvdh, line 2436:
```suggestion
			// alice aborts her join before the resync completes
```

tests/federation_room_join_partial_state_test.go
```diff
@@ -2405,6 +2405,77 @@ func TestPartialStateJoin(t *testing.T) {
...
+
+			// @elsie joins the room.
+			joinEvent := createJoinEvent(t, server, room, server.UserID("elsie"))
+			room.AddEvent(joinEvent)
+			server.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{joinEvent.JSON()}, nil)
+			awaitEventViaSync(t, alice, room.RoomID, joinEvent.EventID(), syncToken)
+
+			// @elsie's device list ought to be cached.
+			syncToken = mustSyncUntilDeviceListsHas(t, alice, syncToken, "changed", server.UserID("elsie"))
+			mustQueryKeysWithFederationRequest(t, alice, userDevicesChannel, server.UserID("elsie"))
```
richvdh, line 2433: this seems to be checking that the device list is *not* cached, in contrast to the comment?

squahtx: The first query (this one) brings the device list into the cache, and the second one tests that the device list is in the cache.

richvdh: so, when it says `// @elsie's device list ought to be cached.` ... we expect the device list to not be cached at that point? It doesn't seem very clear, either way.

squahtx: How does this sound?
```go
// @elsie's device list ought to be cached after the first request.
// Try to bring their device list into the homeserver's cache, then test whether it really was cached.
mustQueryKeysWithFederationRequest(...)
mustQueryKeysWithoutFederationRequest(...)
```

DMRobertson: Took me a moment to grok this as well. As I understand it, both `mustQueryKeys` functions query keys via the CSAPI. The `WithFederationRequest` checks that the server under test calls out over federation to the complement-managed server; the `Without` variant checks that this does not happen.

How about this?

```
The server under test should now be tracking @elsie's device list. Enforce this in two steps.
First, have Alice request Elsie's keys via the CS API and check that hs1 makes a federation request to serve Alice's request.
Then repeat Alice's request and check that hs1 does _not_ make a second federation request.
```

DMRobertson: I've had a go at rewriting this in af6228a. Please let me know what you think.

richvdh: yeah, this seems much clearer now.

richvdh (approved):
lgtm

PR #477, opened 2022-09-23
Faster room joins: Add edge case tests for device list tracking

squahtx:
Test device list tracking for users incorrectly thought to be in a
partial state room once the room transitions to full state.

---

Built on top of #476.


squahtx:
Expected to fail until the last bullet in https://github.com/matrix-org/synapse/issues/12993#issuecomment-1253785129 is done.

clokep:
@squahtx Does this just need Ci kicked (or maybe merge `main` while you're at it)?

squahtx:
> @squahtx Does this just need Ci kicked (or maybe merge `main` while you're at it)?

It does, and Rich has already done so.

erikjohnston (approved):

Issue #478, opened 2022-09-20
`Search_works_across_an_upgraded_room_and_its_predecessor` is flaky


DMRobertson:
https://github.com/matrix-org/synapse/actions/runs/3081667619/jobs/4982001564

Ported from sytest in https://github.com/matrix-org/complement/pull/464

richvdh:
> if we think a test is just plain wrong, raise in complement

Moving to the complement repo, then.

richvdh:
https://github.com/matrix-org/synapse/actions/runs/3130081166/jobs/5079975834

DMRobertson:
Another one here: https://github.com/matrix-org/synapse/actions/runs/3126059365/jobs/5071114994

PR #479, opened 2022-09-27
Allow search results to appear in multiple orders

richvdh:
Fixes https://github.com/matrix-org/complement/issues/478, I hope, though there may be other tests with the same problem.

### DIFF
--- a/tests/csapi/apidoc_search_test.go
+++ b/tests/csapi/apidoc_search_test.go
@@ -212,6 +212,7 @@ func TestSearch(t *testing.T) {
 					"msgtype": "m.text",
 				},
 			})
+			t.Logf("Sent message before upgrade with event ID %s", eventBeforeUpgrade)
 
 			upgradeBody := client.WithJSONBody(t, map[string]string{
 				"new_version": "9",
@@ -219,6 +220,7 @@ func TestSearch(t *testing.T) {
 			upgradeResp := alice.MustDoFunc(t, "POST", []string{"_matrix", "client", "v3", "rooms", roomID, "upgrade"}, upgradeBody)
 			body := must.ParseJSON(t, upgradeResp.Body)
 			newRoomID := must.GetJSONFieldStr(t, body, "replacement_room")
+			t.Logf("Replaced room %s with %s", roomID, newRoomID)
 
 			eventAfterUpgrade := alice.SendEventSynced(t, newRoomID, b.Event{
 				Type: "m.room.message",
@@ -227,6 +229,7 @@ func TestSearch(t *testing.T) {
 					"msgtype": "m.text",
 				},
 			})
+			t.Logf("Sent message after upgrade with event ID %s", eventAfterUpgrade)
 
 			searchRequest := client.WithJSONBody(t, map[string]interface{}{
 				"search_categories": map[string]interface{}{
@@ -245,7 +248,7 @@ func TestSearch(t *testing.T) {
 
 			expectedEvents := map[string]string{
 				eventBeforeUpgrade: "Message before upgrade",
-				eventAfterUpgrade: "Message after upgrade",
+				eventAfterUpgrade:  "Message after upgrade",
 			}
 			must.MatchResponse(t, resp, match.HTTPResponse{
 				StatusCode: http.StatusOK,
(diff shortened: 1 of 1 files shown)


squahtx (approved):
lgtm

S7evinK:
@richvdh RE your [comment](https://github.com/matrix-org/complement/issues/478#issuecomment-1258690805) 
> The problem here is that the complement test is assuming that the search results will be returned in a particular order, which I don't think is correct (and certainly the sytest that it was ported from did not make that assumption).

Looking at the deleted Sytest code, [it definitely did that assumption](
https://github.com/matrix-org/sytest/pull/1291/files#diff-930631cb47132515687c5b8c56994b6c6502fd1038f2974dbcece6f7bbae2bd0L273-L306).

S7evinK:
Aha.. Looking closer.. It might be related to the different message body? `message 1` vs `message before upgrade`?

richvdh:
> Looking at the deleted Sytest code, it definitely did that assumption

I don't think so. Look at https://github.com/matrix-org/sytest/blob/bebf431edf7311d5147139f44dab97131a784d29/tests/43search.pl#L287: it sets `$result` to the first "result" whose `event_id` matches `$event_id_one`, and then proceeds to do checks on `$result`. That means it will pass even if "message 1" appears second in the result list. It then goes on to do something similar for "message 2".

S7evinK:
Oh, true, didn't read that right.

kegsay (approved):

kegsay:
Though CI seems to be failing:
```
TestSearch/parallel/Search_works_across_an_upgraded_room_and_its_predecessor (1.09s)
      apidoc_search_test.go:208: SendEventSynced waiting for event ID $5_V3NGmyDWKZwYZLCdmopuvHpzes6ao4ILTml5aQAGo
      apidoc_search_test.go:223: SendEventSynced waiting for event ID $1xoBlX0wz7xtwUeiit0K72msQ0cC8wOIvsFe17e_Xck
      apidoc_search_test.go:250: MatchResponse JSONCheckOff: unexpected item  - http://localhost:49342/_matrix/client/v3/search => {"search_categories":{"room_events":{"results":[{"rank":0.5,"result":{"type":"m.room.message","room_id":"!zgnaljHeflTPgVGwXI:hs1","sender":"@alice:hs1","content":{"body":"Message before upgrade","msgtype":"m.text"},"origin_server_ts":1664277514508,"unsigned":{"age":733},"event_id":"$5_V3NGmyDWKZwYZLCdmopuvHpzes6ao4ILTml5aQAGo","user_id":"@alice:hs1","age":733},"context":{}},{"rank":0.5,"result":{"type":"m.room.message","room_id":"!ZrmSMkfiHQuBjbdUuD:hs1","sender":"@alice:hs1","content":{"body":"Message after upgrade","msgtype":"m.text"},"origin_server_ts":1664277515138,"unsigned":{"age":103},"event_id":"$1xoBlX0wz7xtwUeiit0K72msQ0cC8wOIvsFe17e_Xck","user_id":"@alice:hs1","age":103},"context":{}}],"count":2,"highlights":[]}}}
```

tests/csapi/apidoc_search_test.go
```diff
@@ -242,20 +242,32 @@ func TestSearch(t *testing.T) {
...
-					match.JSONKeyPresent(result0 + ".type"),
-					match.JSONKeyEqual(result0+".event_id", eventBeforeUpgrade),
-					match.JSONKeyEqual(result1+".event_id", eventAfterUpgrade),
-					match.JSONKeyEqual(result0+".content.body", "Message before upgrade"),
-					match.JSONKeyEqual(result1+".content.body", "Message after upgrade"),
+					match.JSONKeyArrayOfSize(sce+".results", 2),
+
+					// the results can be in either order: check that both are there and that the content is as expected
+					match.JSONCheckOff(sce+".results", []interface{} {eventBeforeUpgrade, eventAfterUpgrade}, func(res gjson.Result) interface{} {
+						return res.Get("event_id")
```
kegsay, line 258:
```suggestion
						return res.Get("event_id").Str
```

kegsay: This is the bug. The error message on CI is unhelpful because it's using `%s` not `%v`: https://github.com/matrix-org/complement/blob/41787c72bf6c9faed27c840a0473426eaa4f6e77/internal/match/json.go#L122

PR #480, opened 2022-09-27
Improve error messages from JSONCheckOff

richvdh:
Hopefully this will make the errors from `JSONCheckOff` and friends a bit more consistent and comprehensible.


kegsay (approved):
Thank you!

PR #482, opened 2022-09-27
Test pagination of the `/relations` endpoint.

clokep:
Builds on #467 to test pagination of `/relations` endpoint, also tests MSC3715 support (forward pagination).


squahtx (approved):
lgtm!

PR #487, opened 2022-09-30
Faster room joins: Fix bug in `awaitPartialStateJoinCompletion`

squahtx:
`/sync` is not a good way to wait for a room to transition to full
state, since it only waits for the full state at particular events
instead of the whole room. This caused flakiness in the "Device list
updates no longer reach departed servers after partial state join
completes" test.

Use `/members` instead to wait for partial state joins to complete.

---

Fixes https://github.com/matrix-org/synapse/issues/13977.

### DIFF
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -121,7 +121,14 @@ func TestPartialStateJoin(t *testing.T) {
 	) {
 		t.Helper()
 
-		user.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(user.UserID, room.RoomID))
+		// Use a `/members` request to wait for the room to be un-partial stated.
+		// We avoid using `/sync`, as it only waits (or used to wait) for full state at
+		// particular events, rather than the whole room.
+		user.MustDoFunc(
+			t,
+			"GET",
+			[]string{"_matrix", "client", "v3", "rooms", room.RoomID, "members"},
+		)
 		t.Logf("%s's partial state join to %s completed.", user.UserID, room.RoomID)
 	}
 


tests/federation_room_join_partial_state_test.go
```diff
@@ -121,7 +121,14 @@ func TestPartialStateJoin(t *testing.T) {
...
 
-		user.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(user.UserID, room.RoomID))
+		// Use a `/members` request to wait for the room to be un-partial stated.
+		// We avoid using `/sync`, as it only waits (or used to wait) for full state at
+		// particular events, rather than the whole room.
+		user.MustDoFunc(
+			t,
+			"GET",
+			[]string{"_matrix", "client", "v3", "rooms", room.RoomID, "members"},
+		)
```
erikjohnston, line 131: I wonder if we should expose a way for clients to know if the room is partial, rather than trying to infer it? It even sounds like something we might want to expose in the UI?

erikjohnston (approved):

tests/federation_room_join_partial_state_test.go
squahtx: That sounds possibly useful. Though I wonder how much the concept of a partial state room (as opposed to a partial state event) is specific to the synapse implementation.

I suppose there are at least three reasons `awaitPartialStateJoinCompletion` might be used:
* to wait for a particular event to have full state
* to wait for the "current" state of the room to be a full state
* to wait for all events in the room to have full state (implies the previous 2)

I'd guess that clients and most tests are primarily interested in the 2nd property and it doesn't matter if there's still a partial state event somewhere back in the dag (which synapse doesn't allow, but a theoretical homeserver implementation might).

PR #488, opened 2022-10-04
Add pushrule sync tests

S7evinK:
Adds the tests from tests/61push/06_push_rules_in_sync.pl

### DIFF
--- a/tests/csapi/push_test.go
+++ b/tests/csapi/push_test.go
@@ -1,7 +1,10 @@
 package csapi_tests
 
 import (
+	"fmt"
 	"testing"
+
+	"github.com/tidwall/gjson"
 
 	"github.com/matrix-org/complement/internal/b"
 	"github.com/matrix-org/complement/internal/client"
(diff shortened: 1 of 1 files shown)


tests/csapi/push_test.go
```diff
@@ -41,3 +44,93 @@ func TestPushRuleCacheHealth(t *testing.T) {
...
+		}
+		errChan <- nil
+	}()
+	// Wait for the sync to actually start, to avoid responding immediately
+	<-syncStarted
+	fn()
+
+	if err := <-errChan; err != nil {
+		t.Fatal(err)
+	}
```
squahtx, line 134: Is there a reason we are using a goroutine instead of calling `fn()`, then doing the sync?

squahtx: Thanks, that makes sense. The wait doesn't actually guarantee that the `/sync` has started before `fn()` completes, which confused me. Could we make it a little clearer?
Maybe `// Try to wait for the sync to actually start, so that we test wakeups`?

tests/csapi/push_test.go
```diff
@@ -41,3 +45,108 @@ func TestPushRuleCacheHealth(t *testing.T) {
...
+		}
+		errChan <- nil
+	}()
+	// Try to wait for the sync to actually start, so that we test wakeups
+	select {
+	case <-syncStarted:
+		fn()
+	case <-time.After(time.Second * 5):
+		// even though this should mostly be impossible, make sure we have a timeout
+		t.Fatalf("go routine didn't start")
```
squahtx, line 135:
```suggestion
		t.Fatalf("goroutine didn't start")
```

squahtx (approved):
Thanks for updating the comment. The PR largely looks good.
I completely missed that we neglect to close the channels, good spot.

I'm not fully convinced the timeouts are worth the additional complexity, especially if we think they are impossible to hit, but won't block the PR on it.

tests/csapi/push_test.go
S7evinK: > Thanks for updating the comment. The PR largely looks good. I completely missed that we neglect to close the channels, good spot.
> 
> I'm not fully convinced the timeouts are worth the additional complexity, especially if we think they are impossible to hit, but won't block the PR on it.

Given CI can sometimes be veeeery slow, I'll just keep them :) Thanks for the review!

PR #490, opened 2022-10-05
Faster room joins: Deflake events-arriving-before-prev_events test

squahtx:
Synapse can take a while to persist 100 outliers. Instead of doing one big wait for the last outlier to be persisted, do some smaller waits for intermediate outliers and a small wait for the last outlier.

ie. as long as the homeserver under test is making progress, we are happy to continue waiting.

Fixes https://github.com/matrix-org/synapse/issues/13777.


squahtx:
Thanks for the review!

PR #491, opened 2022-10-05
Generate unique room IDs when creating rooms

squahtx:
This ensures that room IDs are not re-used across tests, even if a
Complement server happens to re-use the same port as a previous one,
which reduces noise when searching through logs and prevents homeservers
from getting confused when multiple test cases re-use the same
homeserver deployment.

---

Fixes https://github.com/matrix-org/synapse/issues/13975 together with https://github.com/matrix-org/complement/pull/486.

### DIFF
--- a/internal/federation/server.go
+++ b/internal/federation/server.go
@@ -161,7 +161,12 @@ func (s *Server) MustMakeRoom(t *testing.T, roomVer gomatrixserverlib.RoomVersio
 	if !s.listening {
 		s.t.Fatalf("MustMakeRoom() called before Listen() - this is not supported because Listen() chooses a high-numbered port and thus changes the server name and thus changes the room ID. Ensure you Listen() first!")
 	}
-	roomID := fmt.Sprintf("!%d:%s", len(s.rooms), s.serverName)
+	// Generate a unique room ID, prefixed with an incrementing counter.
+	// This ensures that room IDs are not re-used across tests, even if a Complement server happens
+	// to re-use the same port as a previous one, which
+	//  * reduces noise when searching through logs and
+	//  * prevents homeservers from getting confused when multiple test cases re-use the same homeserver deployment.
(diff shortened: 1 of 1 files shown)


internal/federation/server.go
```diff
@@ -161,7 +161,12 @@ func (s *Server) MustMakeRoom(t *testing.T, roomVer gomatrixserverlib.RoomVersio
 	if !s.listening {
 		s.t.Fatalf("MustMakeRoom() called before Listen() - this is not supported because Listen() chooses a high-numbered port and thus changes the server name and thus changes the room ID. Ensure you Listen() first!")
 	}
-	roomID := fmt.Sprintf("!%d:%s", len(s.rooms), s.serverName)
+	// Generate a unique room ID, prefixed with an incrementing counter.
+	// This ensures that room IDs are not re-used across tests, even if a Complement server happens
+	// to re-use the same port as a previous one, which
+	//  * reduces noise when searching through logs and
+	//  * prevents homeservers from getting confused when multiple test cases re-use the same homeserver deployment.
+	roomID := fmt.Sprintf("!%d-%s:%s", len(s.rooms), util.RandomString(18), s.serverName)
```
DMRobertson, line 169: Do we ever delete from or otherwise truncate `s.rooms`? Is it possible that some race could mean that we reuse a value of `len(s.rooms)`? (I have no idea what sort of concurrency problems one has to ancipate in Go.)

On second thought: we were using `len(s.rooms)` before, so this can't be any worse.

DMRobertson (approved):

internal/federation/server.go
squahtx: > Do we ever delete from or otherwise truncate `s.rooms`?

I did a search and can't find anywhere where we do that. I'd expect the random part of the room ID to handle conflicts with `len(s.rooms)` unless we are unlucky to the degree of ~`1/2**107`.

kegsay (approved):

PR #493, opened 2022-10-06
Enable debug logging for TestWriteMDirectAccountData

richvdh:
Yet another straw to grasp at for debugging https://github.com/matrix-org/synapse/issues/13334

### DIFF
--- a/tests/direct_messaging_test.go
+++ b/tests/direct_messaging_test.go
@@ -28,6 +28,10 @@ func TestWriteMDirectAccountData(t *testing.T) {
 
 	alice := deployment.Client(t, "hs1", "@alice:hs1")
 	bob := deployment.Client(t, "hs1", "@bob:hs1")
+
+	// additional logging to debug https://github.com/matrix-org/synapse/issues/13334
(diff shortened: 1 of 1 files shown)


tests/direct_messaging_test.go
```diff
@@ -28,6 +28,10 @@ func TestWriteMDirectAccountData(t *testing.T) {
 
 	alice := deployment.Client(t, "hs1", "@alice:hs1")
 	bob := deployment.Client(t, "hs1", "@bob:hs1")
+
+	// additional logging to debug https://github.com/matrix-org/synapse/issues/13334
+	alice.Debug = true
+
```
DMRobertson, line 34: Was gonna ask if we wanna do the same thing for `bob`, but we never make any client-server requests on his behalf.

DMRobertson (approved):

tests/direct_messaging_test.go
richvdh: yeah, the problems all seem to be with requests from alice.

PR #496, opened 2022-10-11
Add tests for MSC3771 / MSC3773: threaded receipts & notifications

clokep:
Related to #351.

Tests for the bug fixed in matrix-org/synapse#14140 specifically, but also a bit more general.


tests/csapi/thread_notifications_test.go
```diff
@@ -0,0 +1,178 @@
...
+	// Mark the first event as read with a threaded receipt and check counts.
+	bob.MustDoFunc(t, "POST", []string{"_matrix", "client", "v3", "rooms", roomID, "receipt", "m.read", firstEventID}, client.WithJSONBody(t, map[string]interface{}{"thread_id": "main"}))
+	bob.MustSyncUntil(
+		t, client.SyncReq{Since: token},
+		client.SyncTimelineHas(roomID, func(r gjson.Result) bool {
+			return r.Get("event_id").Str == secondEventID
+		}),
+		syncHasUnreadNotifs(roomID, func(r gjson.Result, t gjson.Result) bool {
+			return r.Get("highlight_count").Num == 2 && r.Get("notification_count").Num == 3 && !t.Exists()
+		}),
```
MadLittleMods, line 122: We should explain these counts. "We expect .. to be unread"

tests/csapi/thread_notifications_test.go
```diff
@@ -0,0 +1,178 @@
...
+	threadFilter := `{"room":{"timeline":{"unread_thread_notifications":true}}}`
+
+	// Check the unthreaded and threaded counts.
+	bob.MustSyncUntil(
+		t, client.SyncReq{Since: token},
+		client.SyncTimelineHas(roomID, func(r gjson.Result) bool {
+			return r.Get("event_id").Str == secondEventID
+		}),
+		syncHasUnreadNotifs(roomID, func(r gjson.Result, t gjson.Result) bool {
+			return r.Get("highlight_count").Num == 2 && r.Get("notification_count").Num == 4 && !t.Exists()
```
MadLittleMods, line 99: Why is highlight count `2` when we only sent 1 highlight message according to the comments above?

MadLittleMods: I guess from `firstEventID` and `secondEventID` but we should indicate that it's also a highlight message for `bob` in the comments.

tests/csapi/thread_notifications_test.go
```diff
@@ -0,0 +1,178 @@
...
+			"msgtype": "m.text",
+			"body":    fmt.Sprintf("Thread response %s!", bob.UserID),
+			"m.relates_to": map[string]interface{}{
+				"event_id": firstEventID,
+				"rel_type": "m.thread",
+			},
+		},
+	})
+
+	// Send a highlight message and re-check counts.
```
MadLittleMods, line 80: In reference to the "check the highlight and notification counts from bob." comment above,

The "re-check counts." part of the comment here is a little weird given the comment above. We don't seem to check the counts in between these two comments. Sounds like we need to adjust these comments to reflect that

tests/csapi/thread_notifications_test.go
```diff
@@ -0,0 +1,178 @@
...
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+	bob := deployment.Client(t, "hs1", "@bob:hs1")
+
+	roomID := alice.CreateRoom(t, map[string]interface{}{"preset": "public_chat"})
+	bob.JoinRoom(t, roomID, nil)
+
+	token := bob.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(bob.UserID, roomID))
+
+	// Send messages as alice and then check the highlight and notification counts from bob.
+	firstEventID := alice.SendEventSynced(t, roomID, b.Event{
```
MadLittleMods, line 48:
```suggestion
	firstHighlightEventForBobID := alice.SendEventSynced(t, roomID, b.Event{
```

Related to https://github.com/matrix-org/complement/pull/496#discussion_r996208469

tests/csapi/thread_notifications_test.go
```diff
@@ -0,0 +1,178 @@
...
+		unreadThreadNotifications := topLevelSyncJSON.Get("rooms.join." + client.GjsonEscape(roomID) + ".unread_thread_notifications")
+		if !unreadNotifications.Exists() {
+			return fmt.Errorf("syncHasUnreadNotifs(%s): missing notifications", roomID)
+		}
+		if check(unreadNotifications, unreadThreadNotifications) {
+			return nil
+		}
+		return fmt.Errorf("syncHasUnreadNotifs(%s): check function did not pass: %v / %v", roomID, unreadNotifications.Raw, unreadThreadNotifications.Raw)
+	}
+}
```
MadLittleMods, line 26: We could craft much better errors if we changed the `check` function into a `expectedMap` and displayed a diff

```js
{
  main: {
    highlight_count: 2,
    notification_count: 4
  },
  [firstEventID]: {
    label: "firstEventIDThreadNotifications",
    highlight_count: 1,
    notification_count: 4
  }
}
```

 - `syncHasUnreadNotifs for main: Expected 2 / 4 but actual was 3  / 4 - Format is (highlights / notifications)`
 - `syncHasUnreadNotifs for firstEventIDThreadNotifications ($abcdef): Expected 1 / 2 but actual was 1 / 1 - Format is (highlights / notifications)`

tests/csapi/thread_notifications_test.go
clokep: No, we send a highlight both in the thread and out of the thread.

I don't see anything saying that there is only a single highlight message in the comments, but I'll try to make it a bit clearer.

tests/csapi/thread_notifications_test.go
clokep: I left this somewhat generic because:

1. It seems to fit better how check functions are done with complement.
2. I suspect we will eventually want to check multiple threads.

tests/csapi/thread_notifications_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/runtime"
+)
+
+func syncHasUnreadNotifs(roomID string, check func(gjson.Result, gjson.Result) bool) client.SyncCheckOpt {
+	return func(clientUserID string, topLevelSyncJSON gjson.Result) error {
+		unreadNotifications := topLevelSyncJSON.Get("rooms.join." + client.GjsonEscape(roomID) + ".unread_notifications")
+		unreadThreadNotifications := topLevelSyncJSON.Get("rooms.join." + client.GjsonEscape(roomID) + ".unread_thread_notifications")
+		if !unreadNotifications.Exists() {
+			return fmt.Errorf("syncHasUnreadNotifs(%s): missing notifications", roomID)
```
DMRobertson, line 19: Should this say "missing unread_notifications"?

tests/csapi/thread_notifications_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"testing"
+
+	"github.com/tidwall/gjson"
+
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/runtime"
+)
+
+func syncHasUnreadNotifs(roomID string, check func(gjson.Result, gjson.Result) bool) client.SyncCheckOpt {
```
DMRobertson, line 14: Might be worth a quick docstring (or whatever they're called in go) to describe what this does?

Something like this?

```
// Builds a `syncCheckOpt` which enforces that a sync result satisfies some `check` function 
// on the `unread_notifications` and `unread_thread_notifications` fields for the given room.
// The `unread_notifications` field must exist, or else the overall syncCheckOpt will be considered
// as failing.
```

tests/csapi/thread_notifications_test.go
```diff
@@ -0,0 +1,196 @@
...
+//
+// A<--B<--C
+// ^
+// +---D
+//
+// Where C and D generate highlight notifications.
+//
+// Notification counts and receipts are handled by bob.
+func TestThreadedReceipts(t *testing.T) {
+	runtime.SkipIf(t, runtime.Dendrite) // not supported
```
DMRobertson, line 44: Feels a bit icky to couple the test to the implementation, but I suppose we do this everywhere. Might be possible to do something cleaner with build tags, but  that might be a faff too (would have to e.g. update Synapse CI and complement.sh...)

This is probably fine as it is.

tests/csapi/thread_notifications_test.go
```diff
@@ -0,0 +1,196 @@
...
+	defer deployment.Destroy(t)
+
+	// Create a room with alice and bob.
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+	bob := deployment.Client(t, "hs1", "@bob:hs1")
+
+	roomID := alice.CreateRoom(t, map[string]interface{}{"preset": "public_chat"})
+	bob.JoinRoom(t, roomID, nil)
+
+	token := bob.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(bob.UserID, roomID))
```
DMRobertson, line 55: perhaps call this `bobNextBatch`?

tests/csapi/thread_notifications_test.go
DMRobertson: It might also help to refer to the names `A, B, C, D` from the diagram?

tests/csapi/thread_notifications_test.go
```diff
@@ -0,0 +1,196 @@
...
+//
+// 1. Send a series of messages, some of which are in threads.
+// 2. Send combinations of threaded and unthreaded receipts.
+// 3. Ensure the notification counts are updated appropriately.
+//
+// This sends four messages as alice creating a DAG somewhat like:
+//
+// A<--B<--C
+// ^
+// +---D
```
DMRobertson, line 38: "DAG somewhat like" set off my pedantry alarm! I think this is a bit naughty since complement will generate $D$ using $\{C\}$ as its set of prev events.

Maybe let's use the word timeline instead of DAG?

```suggestion
// This sends four messages as alice creating a timeline somewhat like:
//
// A<--B<--C    [thread]
// ^
// +---D        [main timeline]
```

tests/csapi/thread_notifications_test.go
```diff
@@ -0,0 +1,196 @@
...
+		client.SyncTimelineHas(roomID, func(r gjson.Result) bool {
+			return r.Get("event_id").Str == secondEventID
+		}),
+		syncHasUnreadNotifs(roomID, func(r gjson.Result, t gjson.Result) bool {
+			return r.Get("highlight_count").Num == 2 && r.Get("notification_count").Num == 4 && !t.Exists()
+		}),
+	)
+	bob.MustSyncUntil(
+		t,
+		client.SyncReq{Since: token, Filter: threadFilter}, client.SyncTimelineHas(roomID, func(r gjson.Result) bool {
```
DMRobertson, line 116: Suggest starting `SyncTimelineHas` on a new line. (Give me `black` for golang. Surprised gofmt doesn't just enforce this for you---I thought go was all about opinionated tooling?)

tests/csapi/thread_notifications_test.go
```diff
@@ -0,0 +1,196 @@
...
+			return r.Get("event_id").Str == secondEventID
+		}),
+		syncHasUnreadNotifs(roomID, func(r gjson.Result, t gjson.Result) bool {
+			threadNotifications := t.Get(client.GjsonEscape(firstEventID))
+			return r.Get("highlight_count").Num == 1 && r.Get("notification_count").Num == 2 &&
```
DMRobertson, line 121: Slightly surprised that the notification count in the thread is 2 here (B and C) rather than 3 (A, B and C). But presumably that's intentional?

DMRobertson (approved):
Eric's already taken a look at this (thanks @MadLittleMods), but this is in the review queue so I also took a look.

I'd be happy to see this land as-is! I've left some thoughts and suggestions, but they're nitpicking and left to your discretion.

tests/csapi/thread_notifications_test.go
clokep: I think we only use build tags for unstable features?

tests/csapi/thread_notifications_test.go
clokep: I thought `gofmt` was more opinionated than it was...

tests/csapi/thread_notifications_test.go
clokep: `A` is not part of the thread itself. (This is specced in the MSC.)

It would be double counted otherwise.

PR #508, opened 2022-10-14
Sytest: Add `50federation` / `02server-names` & `10query-profile` remaining tests

ShadowJonathan:
This adds two sytests:
- `./tests/50federation/02server-names.pl:test "Non-numeric ports in server names are rejected",`
- `./tests/50federation/10query-profile.pl:test "Inbound federation can query profile data",`

This also adds one generic function, `(*federation.Server).DoFederationRequest`, as a counterpart to `SendFederationRequest`, though with the only difference being that it returns the response instead of trying to JSON-deserialise it.

Note: This function is largely a copy of `SendFederationRequest`, I tried to figure out how I can dedup them into one function, but seeing as they both use different functions from the HTTP Client object, it's probably not possible without having it pass a `func(*gomatrixserverlib.Client)` or likewise to a third internal function.

### DIFF
--- a/internal/federation/server.go
+++ b/internal/federation/server.go
@@ -243,9 +243,40 @@ func (s *Server) SendFederationRequest(
 	if httpError, ok := err.(gomatrix.HTTPError); ok {
 		t.Logf("[SSAPI] %s %s%s => error(%d): %s (%s)", req.Method(), req.Destination(), req.RequestURI(), httpError.Code, err, time.Since(start))
 	} else if err == nil {
-		t.Logf("[SSAPI] %s %s%s => 2xx (%s)", req.Method(), req.Destination(), req.RequestURI(), time.Since(start))
+		t.Logf("[SSAPI] %s %s%s => (%s)", req.Method(), req.Destination(), req.RequestURI(), time.Since(start))
 	}
 	return err
+}
+
+// DoFederationRequest signs and sends an arbitrary federation request from this server, and returns the response.
+//
+// The requests will be routed according to the deployment map in `deployment`.
+func (s *Server) DoFederationRequest(
+	ctx context.Context,
+	t *testing.T,
+	deployment *docker.Deployment,
+	req gomatrixserverlib.FederationRequest) (*http.Response, error) {
+	if err := req.Sign(gomatrixserverlib.ServerName(s.serverName), s.KeyID, s.Priv); err != nil {
+		return nil, err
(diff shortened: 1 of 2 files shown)


internal/federation/server.go
```diff
@@ -243,11 +243,42 @@ func (s *Server) SendFederationRequest(
 	if httpError, ok := err.(gomatrix.HTTPError); ok {
 		t.Logf("[SSAPI] %s %s%s => error(%d): %s (%s)", req.Method(), req.Destination(), req.RequestURI(), httpError.Code, err, time.Since(start))
 	} else if err == nil {
-		t.Logf("[SSAPI] %s %s%s => 2xx (%s)", req.Method(), req.Destination(), req.RequestURI(), time.Since(start))
```
ShadowJonathan, line 246: While working with `DoFederationRequest`, I noticed this will unconditionally insert this `2xx` while not checking if it's the actual status code, as the test writer might expect an error, adjust `resBody` to the error body, receive an expected `4xx`, and this'll still say `2xx`.

There's no way to fetch the actual status code at this point, as the response body is already cleaned up in the `DoRequestAndParseResponse` function.

kegsay: Are you sure? The docs in GMSL state:
```go
// DoRequestAndParseResponse calls DoHTTPRequest and then decodes the response.
//
// If the HTTP response is not a 200, an attempt is made to parse the response
// body into a gomatrix.RespError. In any case, a non-200 response will result
// in a gomatrix.HTTPError.
func (fc *Client) DoRequestAndParseResponse(
```

kegsay (approved):
I think that the 2xx comment is still valid, but I trust you wouldn't modify it for no good reason. It would be interesting to know the exact repro conditions here.

> This also adds one generic function, (*federation.Server).DoFederationRequest, as a counterpart to SendFederationRequest, though with the only difference being that it returns the response instead of trying to JSON-deserialise it.

The addition of `DoFederationRequest` is fine, but the code impl duplication is a shame. I can see why you've ended up doing this though, as the GMSL API somewhat forces you into this shape as `DoHTTPRequest` and `DoRequestAndParseResponse` do not compose well together.

internal/federation/server.go
ShadowJonathan: Oh! In that case I'll revert this, thanks.

kegsay (approved):

PR #511, opened 2022-10-15
Sytest: Add `30rooms/22profile` test

ShadowJonathan:
This adds one sytest:
- `./tests/30rooms/22profile.pl:   test "$datum updates affect room member events",`

This also alters `SyncJoinedTo` in a backwards-compatible way to receive additional check functions, to check on the event of the join, this could be useful for things like restricted room joins, or - in this case - checking changed membership profile data.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -741,29 +741,41 @@ func SyncInvitedTo(userID, roomID string) SyncCheckOpt {
 	}
 }
 
-// Check that `userID` gets joined to `roomID` by inspecting the join timeline for a membership event
-func SyncJoinedTo(userID, roomID string) SyncCheckOpt {
+// Check that `userID` gets joined to `roomID` by inspecting the join timeline for a membership event.
+//
+// Additional checks can be passed to narrow down the check, all must pass.
+func SyncJoinedTo(userID, roomID string, checks ...func(gjson.Result) bool) SyncCheckOpt {
 	checkJoined := func(ev gjson.Result) bool {
-		return ev.Get("type").Str == "m.room.member" && ev.Get("state_key").Str == userID && ev.Get("content.membership").Str == "join"
+		if ev.Get("type").Str == "m.room.member" && ev.Get("state_key").Str == userID && ev.Get("content.membership").Str == "join" {
+			for _, check := range checks {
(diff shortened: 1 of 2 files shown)


internal/client/client.go
```diff
@@ -741,29 +741,41 @@ func SyncInvitedTo(userID, roomID string) SyncCheckOpt {
...
-		err = loopArray(
+		secondErr := loopArray(
 			topLevelSyncJSON, "rooms.join."+GjsonEscape(roomID)+".state.events", checkJoined,
 		)
-		if err == nil {
+		if secondErr == nil {
 			return nil
 		}
-		return fmt.Errorf("SyncJoinedTo(%s): %s", roomID, err)
+		return fmt.Errorf("SyncJoinedTo(%s): %s & %s", roomID, firstErr, secondErr)
```
ShadowJonathan, line 778: While taking a look at this function, I noticed that it effectively ignored the first error when reporting it, so altering it here so that both errors are reported would help debugging test cases, such as empty arrays.

tests/csapi/room_profile_test.go
```diff
@@ -0,0 +1,47 @@
...
+func TestDisplayNameUpdate(t *testing.T) {
+	testProfileFieldUpdate(t, "displayname")
+}
+
+// sytest: $datum updates affect room member events
+func testProfileFieldUpdate(t *testing.T, field string) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+
+	const bogusData = "LemurLover"
```
kegsay, line 25: For things like `avatar_url` we expect this to be an `mxc://` URI. It would be reasonable to think that a HS impl may reject non-mxc URIs, so we can't put bogus data for this field. Yank the data out to the function args please `testProfileFieldUpdate(t *testing.T, field, val string)`

kegsay: Actually the sytest does this incorrectly so meh, fine.

kegsay (approved):

ShadowJonathan:
Re MXC validation: i asked about this in the homeservers room but i got no response, seeing how the spec is also silent on it, and the sytest negatively asserts it with this, i added it like that

Issue #515, opened 2022-10-16
`TestMembersLocal` -> "member presence" test relies on parallel side-effects


ShadowJonathan:
I found this while investigating why the `runtime.Dendrite` skipif was present;

The [TestMembersLocal](https://github.com/matrix-org/complement/blob/6c31e8dd4e04fc12e9b3a7aa5553d80b42092cf8/tests/csapi/rooms_members_local_test.go#L14) test, specifically, the [`"Existing members see new members' presence"`](https://github.com/matrix-org/complement/blob/6c31e8dd4e04fc12e9b3a7aa5553d80b42092cf8/tests/csapi/rooms_members_local_test.go#L38) test, relies on [this](https://github.com/matrix-org/complement/blob/6c31e8dd4e04fc12e9b3a7aa5553d80b42092cf8/tests/csapi/rooms_members_local_test.go#L32) simultaneously-executed room join to pass.

When ran in isolation, this test fails, as that line is not executed to make this test pass, it times out on syncing.

PR #516, opened 2022-10-16
Fix local members presence test

ShadowJonathan:
This makes it able to be ran in isolation, and fixes it for dendrite test failures, as https://github.com/matrix-org/dendrite/issues/2803 is fixed.

Fixes https://github.com/matrix-org/complement/issues/515

### DIFF
--- a/tests/csapi/rooms_members_local_test.go
+++ b/tests/csapi/rooms_members_local_test.go
@@ -13,7 +13,24 @@ import (
 )
 
 func TestMembersLocal(t *testing.T) {
-	deployment := Deploy(t, b.BlueprintOneToOneRoom)
+	deployment := Deploy(t, b.MustValidate(b.Blueprint{
+		Name: "ab",
+		Homeservers: []b.Homeserver{
+			{
+				Name: "hs1",
+				Users: []b.User{
+					{
+						Localpart:   "@alice",
+						DisplayName: "Alice",
+					},
+					{
+						Localpart:   "@bob",
+						DisplayName: "Bob",
+					},
+				},
+			},
+		},
+	}))
 	defer deployment.Destroy(t)
 
 	alice := deployment.Client(t, "hs1", "@alice:hs1")
@@ -53,7 +70,6 @@ func TestMembersLocal(t *testing.T) {
 
 		// sytest: Existing members see new members' presence
 		t.Run("Existing members see new members' presence", func(t *testing.T) {
-			runtime.SkipIf(t, runtime.Dendrite) // Still failing
 			t.Parallel()
 			alice.MustSyncUntil(t, client.SyncReq{Since: sinceToken},
 				client.SyncJoinedTo(bob.UserID, roomID),
--- a/tests/csapi/rooms_members_local_test.go
+++ b/tests/csapi/rooms_members_local_test.go
@@ -6,6 +6,8 @@ import (
 
 	"github.com/matrix-org/complement/internal/b"
 	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
 	"github.com/matrix-org/complement/runtime"
 	"github.com/tidwall/gjson"
 )
@@ -17,6 +19,18 @@ func TestMembersLocal(t *testing.T) {
 	alice := deployment.Client(t, "hs1", "@alice:hs1")
 	bob := deployment.Client(t, "hs1", "@bob:hs1")
 	roomID := alice.CreateRoom(t, map[string]interface{}{"preset": "public_chat"})
+
+	res := bob.DoFunc(
+		t,
+		"PUT",
+		[]string{"_matrix", "client", "v3", "presence", bob.UserID, "status"},
+		client.WithJSONBody(t, map[string]interface{}{
+			"presence": "online",
+		}),
+	)
+	must.MatchResponse(t, res, match.HTTPResponse{
+		StatusCode: 200,
+	})
 
 	_, sinceToken := alice.MustSync(t, client.SyncReq{TimeoutMillis: "0"})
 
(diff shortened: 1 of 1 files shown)


H-Shay:
looks like there are some test failures that need to be fixed?

ShadowJonathan:
Fixed, it was a build issue with imports

tests/csapi/rooms_members_local_test.go
```diff
@@ -4,20 +4,55 @@ import (
...
 	"github.com/matrix-org/complement/internal/client"
-	"github.com/matrix-org/complement/runtime"
-	"github.com/tidwall/gjson"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
 )
 
 func TestMembersLocal(t *testing.T) {
-	deployment := Deploy(t, b.BlueprintOneToOneRoom)
+	deployment := Deploy(t, b.MustValidate(b.Blueprint{
```
S7evinK, line 16: Can we please use `BlueprintOneToOneRoom`?

tests/csapi/rooms_members_local_test.go
```diff
@@ -4,20 +4,55 @@ import (
...
+		t,
+		"PUT",
+		[]string{"_matrix", "client", "v3", "presence", bob.UserID, "status"},
+		client.WithJSONBody(t, map[string]interface{}{
+			"presence": "online",
+		}),
+	)
+	must.MatchResponse(t, res, match.HTTPResponse{
+		StatusCode: 200,
+	})
```
S7evinK, line 50: This can be replaced by a simple `bob.MustSync(t, client.SyncReq{})` (optionally set the presence there)

S7evinK (changes requested):
Minor things. :)
Otherwise looks good.

tests/csapi/rooms_members_local_test.go
ShadowJonathan: No, because that messes with the presence test, as then Alice already shares a room with Bob, and it's not conclusively testing if a new sync will include *newly seen* members' presence.

If we use one to one, the latter condition isn't true anymore, as Alice has already "seen" Bob before, in another room.

tests/csapi/rooms_members_local_test.go
ShadowJonathan: The original sytest explicitly uses a presence-setting url, but I can change it to that.

tests/csapi/rooms_members_local_test.go
ShadowJonathan: I'll change it to BlueprintAlice and register bob afterwards

ShadowJonathan:
I'm going to re-add the skipif, to unblock this PR, and if https://github.com/matrix-org/dendrite/issues/2803 is removed, i'll remove the skipif in another PR.

S7evinK:
It should be working, once https://github.com/matrix-org/dendrite/pull/2854 is merged

ShadowJonathan:
It seems https://github.com/matrix-org/complement/pull/563 was made before this PR has had any chance to progress.

Seeing as how the changes introduced there are antithetical to the actual tests being done (i.e. bob joining a room, and alice seeing bob for the first time, but this being subverted with the addition of an extra pre-existing one-on-one room), I will be reverting that.

tests/csapi/rooms_members_local_test.go
```diff
@@ -9,11 +9,11 @@ import (
 )
 
 func TestMembersLocal(t *testing.T) {
-	deployment := Deploy(t, b.BlueprintOneToOneRoom)
+	deployment := Deploy(t, b.BlueprintAlice)
 	defer deployment.Destroy(t)
 
 	alice := deployment.Client(t, "hs1", "@alice:hs1")
-	bob := deployment.Client(t, "hs1", "@bob:hs1")
+	bob := deployment.RegisterUser(t, "hs1", "bob", "bobspassword", false)
```
MadLittleMods, line 16: Per your comment at https://github.com/matrix-org/complement/pull/563#issuecomment-1345195887,

> `BlueprintOneToOneRoom`, which was already determined in #516 to provide faulty results, as Bob and Alice would already see eachother's presence changes through that pre-existing one-on-one room.


We could add a comment here with this same context since it's important to have a fresh user that doesn't have any connection to Alice already to actually test whether presence changes come through.

tests/csapi/rooms_members_local_test.go
```diff
@@ -23,7 +23,7 @@ func TestMembersLocal(t *testing.T) {
 		}),
 	)
 
-	_, incrementalSyncToken := alice.MustSync(t, client.SyncReq{})
+	_, incrementalSyncToken := alice.MustSync(t, client.SyncReq{TimeoutMillis: "0"})
```
MadLittleMods, line 26: The spec says default `timeout` is `0`. Why do we specify it here explicitly?

Spec: https://spec.matrix.org/v1.6/client-server-api/#get_matrixclientv3sync

MadLittleMods, line 26: Perhaps we could name it like this to make it more clear what point this random sync token is from,

```suggestion
	_, incrementalSyncTokenBeforeBobJoinsRoom := alice.MustSync(t, client.SyncReq{TimeoutMillis: "0"})
```

ShadowJonathan: I'm being pedantic, but I also vaguely remember somewhere, somewhen, an implementation using 10/30 seconds as a default like this. It makes sense, else a query-parameter-less sync request would tight-loop.

Here I'm being explicit for the sake of clarity (and pedant), is that alright?

ShadowJonathan:
(CI needs to be re-booped because I only added the signature now, though since it is in code, i'm not sure if it'll go through)

tests/csapi/rooms_members_local_test.go
kegsay: Yes that's fine. Please prefer being explicit.

kegsay:
Rerunning CI then merging.

tests/csapi/rooms_members_local_test.go
MadLittleMods: (just clarifying) The only reason is so that the test runs fast regardless of implementation?

Issue #517, opened 2022-10-16
Two key-upload-related tests being blacklisted by synapse


ShadowJonathan:
When I was looking over `SkipIf`s in the codebase, I found out that these following tests are explicitly blacklisted on synapse.

https://github.com/matrix-org/complement/blob/d5025cd75007609d52a77535388be9317963aedd/tests/csapi/upload_keys_test.go#L77

https://github.com/matrix-org/complement/blob/d5025cd75007609d52a77535388be9317963aedd/tests/csapi/upload_keys_test.go#L56

Seeing as how these tests are related to data validation on a server's side, I do not understand why Synapse wants to skip them, thus this issue as a nag; Why does Synapse skip these tests?

I want to update the comment with a reason given, or a reference to an issue/discussion about exactly that, or to an issue working towards clearing this skip.

PR #499, opened 2022-10-12
Allow relations tests for Dendrite

neilalexander:
To go with matrix-org/dendrite#2790.

PR #519, opened 2022-10-16
Do once-over on all `SkipIf`s in codebase, adding issue numbers, un-skipping passing tests

ShadowJonathan:
Seeing as almost all SkipIfs in complement are due to the fact the originating homeservers do not implement the functionality, I did a once-over to check all of them, and paired them with those issues, and standardized the format of `SkipIf` comments along the way.

Like this, the intended way how to skip a test on a homeserver then always comes paired with an issue to watch or resolve to remove that skip, else these skips would bitrot and the related tests aren't executed, potentially hiding non-compliant behaviour.

Furthermore, some tests *did* pass inbetween then and now, those skips have been removed.

### DIFF
--- a/tests/csapi/apidoc_room_state_test.go
+++ b/tests/csapi/apidoc_room_state_test.go
@@ -330,7 +330,6 @@ func TestRoomState(t *testing.T) {
 			})
 		})
 		t.Run("GET /rooms/:room_id/joined_members is forbidden after leaving room", func(t *testing.T) {
-			runtime.SkipIf(t, runtime.Dendrite) // FIXME: https://github.com/matrix-org/dendrite/issues/2802
 			t.Parallel()
 			roomID := authedClient.CreateRoom(t, map[string]interface{}{})
 			authedClient.LeaveRoom(t, roomID)
--- a/tests/csapi/apidoc_room_state_test.go
+++ b/tests/csapi/apidoc_room_state_test.go
@@ -330,7 +330,7 @@ func TestRoomState(t *testing.T) {
 			})
 		})
 		t.Run("GET /rooms/:room_id/joined_members is forbidden after leaving room", func(t *testing.T) {
-			runtime.SkipIf(t, runtime.Dendrite) // https://github.com/matrix-org/complement/pull/424
+			runtime.SkipIf(t, runtime.Dendrite) // FIXME: https://github.com/matrix-org/dendrite/issues/2802
 			t.Parallel()
 			roomID := authedClient.CreateRoom(t, map[string]interface{}{})
 			authedClient.LeaveRoom(t, roomID)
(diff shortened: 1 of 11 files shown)


tests/csapi/room_relations_test.go
```diff
@@ -12,11 +12,9 @@ import (
 	"github.com/matrix-org/complement/internal/client"
 	"github.com/matrix-org/complement/internal/match"
 	"github.com/matrix-org/complement/internal/must"
-	"github.com/matrix-org/complement/runtime"
 )
 
 func TestRelations(t *testing.T) {
-	runtime.SkipIf(t, runtime.Dendrite) // not supported
```
clokep, line 19: This seems to be #499?

tests/csapi/upload_keys_test.go
```diff
@@ -54,7 +54,7 @@ func TestUploadKey(t *testing.T) {
 
 		// sytest: Rejects invalid device keys
 		t.Run("Rejects invalid device keys", func(t *testing.T) {
-			runtime.SkipIf(t, runtime.Dendrite, runtime.Synapse) // Dendrite doesn't pass, Synapse has it blacklisted
+			runtime.SkipIf(t, runtime.Dendrite, runtime.Synapse) // Blacklisted on Synapse, Dendrite FIXME: https://github.com/matrix-org/dendrite/issues/2804
```
clokep, line 57: Is there a corresponding Synapse issue?

tests/csapi/room_relations_test.go
ShadowJonathan: I did not spot that, should I remove these changes so they don't conflict?

tests/csapi/upload_keys_test.go
ShadowJonathan: There is one here; https://github.com/matrix-org/complement/issues/517

I looked for one on synapse, but didn't open one there as I wasn't entirely sure what the reasoning was why these tests were blacklisted in the first place, so I opened one here to start that inquiry.

tests/csapi/apidoc_room_state_test.go
```diff
@@ -330,7 +330,7 @@ func TestRoomState(t *testing.T) {
 			})
 		})
 		t.Run("GET /rooms/:room_id/joined_members is forbidden after leaving room", func(t *testing.T) {
-			runtime.SkipIf(t, runtime.Dendrite) // https://github.com/matrix-org/complement/pull/424
+			runtime.SkipIf(t, runtime.Dendrite) // FIXME: https://github.com/matrix-org/dendrite/issues/2802
```
ShadowJonathan, line 333:
```suggestion
```

https://github.com/matrix-org/dendrite/issues/2802 is fixed

tests/csapi/room_relations_test.go
clokep: I guess leave it, they'll conflict and they can fix it there.

clokep (approved):

PR #520, opened 2022-10-16
Add "Test URL preview" sytest and basic webserver creation

ShadowJonathan:
This follows up on #279, specifically [this](https://github.com/matrix-org/complement/pull/279#discussion_r833158719) comment.

This re-adds the URL preview test, but also adds a basic framework in `internal/web` for those webservers, with example use being displayed in the re-added test.

This PR is related to https://github.com/matrix-org/synapse/pull/14198, which ~~needs to land first before the test can be un-skipped~~ had landed.

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -13,12 +13,13 @@ import (
 )
 
 type Server struct {
-	Port     int
+	Url      string
+	port     int
 	server   *http.Server
 	listener net.Listener
 }
 
-func NewServer(t *testing.T, configFunc func(router *mux.Router)) *Server {
+func NewServer(t *testing.T, comp *config.Complement, configFunc func(router *mux.Router)) *Server {
 	t.Helper()
 
 	listener, err := net.Listen("tcp", ":0")
@@ -37,7 +38,8 @@ func NewServer(t *testing.T, configFunc func(router *mux.Router)) *Server {
 	go server.Serve(listener)
 
 	return &Server{
-		Port:     port,
+		Url:      fmt.Sprintf("http://%s:%d", comp.HostnameRunningComplement, port),
+		port:     port,
 		server:   server,
 		listener: listener,
 	}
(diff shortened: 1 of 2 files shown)


internal/web/server.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+}
+
+func (s *Server) Close() {
+	s.server.Shutdown(context.Background())
+	s.listener.Close()
+}
+
+// Url returns the formatted URL that homeservers can access the server with, without leading slash
+func (s *Server) Url(comp *config.Complement) string {
```
kegsay, line 52:
```suggestion
func (s *Server) URL(comp *config.Complement) string {
```

kegsay: It would be nice to have this as a calculated property when the server is made, given that is where the port is allocated.

tests/csapi/url_preview_test.go
```diff
@@ -0,0 +1,96 @@
...
+			w.Header().Set("Content-Type", "text/html")
+			w.WriteHeader(200)
+			w.Write([]byte(oGraphHtml))
+		}).Methods("GET")
+	})
+	defer webServer.Close()
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+
+	res := alice.MustDoFunc(t, "GET", []string{"_matrix", "media", "r0", "preview_url"},
```
kegsay, line 67:
```suggestion
	res := alice.MustDoFunc(t, "GET", []string{"_matrix", "media", "v3", "preview_url"},
```
surely?

kegsay (approved):
Otherwise LGTM, this is a good API shape, thanks!

tests/csapi/url_preview_test.go
ShadowJonathan: Sorry, copy pasted from an older PR

kegsay (approved):

PR #521, opened 2022-10-17
Deflake `GET /rooms/:room_id/aliases lists aliases`

DMRobertson:
This essentially ports a retry from sytest that wasn't included in when this was ported to complement: see https://github.com/matrix-org/synapse/issues/14183#issuecomment-1280910298 for the history.

If I had a magic wand, Synapse would provide read-after-write consistency here and the retries would be unnecessary. But it doesn't so I've added them.

Resolves https://github.com/matrix-org/synapse/issues/14183.

### DIFF
--- a/tests/csapi/apidoc_room_alias_test.go
+++ b/tests/csapi/apidoc_room_alias_test.go
@@ -3,6 +3,7 @@ package csapi_tests
 import (
 	"net/http"
 	"testing"
+	"time"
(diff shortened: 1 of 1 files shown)


H-Shay (approved):
This makes sense to me, my only question would be whether it makes sense to use a `waiter` to wait a second before trying the request vs retrying?

PR #523, opened 2022-10-18
Update sytest.list

### LINKED ISSUE #270: Update sytest.list
I'd like for the sytest list to be updated, and be somewhat reproducible, with something like the following header;

```
# Extracted from matrix-org/sytest#198c87
# Extracted with `grep -E "^\s*(test|multi_test) \"" -R . | grep ".pl:" | sort`
```

However, i'm not simply doing this in a PR yet because of the following reasons;
- This'd complicate the history of the file
- Tests have moved (and possibly renamed), this could mess up some of the sytest coverage, and when re-covering it, tests could be duplicated
- Some tests were commented out, while those have been removed, i'd like for the excluded/blacklisted sytests to be tracked in a different file, so this file can be generated more easily

ShadowJonathan:
Closes #270

No new tests added, however some sytests were removed (https://github.com/matrix-org/sytest/pull/1291 and https://github.com/matrix-org/sytest/pull/1297) that were already part of complement.

### DIFF
--- a/sytest.list
+++ b/sytest.list
@@ -1,4 +1,4 @@
-# Extracted from matrix-org/sytest#9ea3f15b
+# Extracted from matrix-org/sytest#da1c9e22
 # With `grep -E "^\s*(test|multi_test) \"" -R . | grep ".pl:" | docker run --rm -i -e LC_ALL=C ubuntu:18.04 sort`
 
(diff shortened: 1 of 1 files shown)


H-Shay (approved):

H-Shay:
Merged #523.

PR #524, opened 2022-10-19
Add tests for /get_missing_events

S7evinK:
Adds `Inbound federation can return missing events for $vis visibility` from `tests/50federation/33room-get-missing-events.pl`

### DIFF
--- a/tests/federation_room_get_missing_events_test.go
+++ b/tests/federation_room_get_missing_events_test.go
@@ -21,7 +21,6 @@ import (
 
 // TODO:
 // Outbound federation can request missing events
-// Inbound federation can return missing events for $vis visibility
 // outliers whose auth_events are in a different room are correctly rejected
 
 // /get_missing_events is used to fill in gaps in the room DAG when a server is pushed (via /send)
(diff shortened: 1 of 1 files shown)


tests/federation_room_get_missing_events_test.go
```diff
@@ -314,3 +313,135 @@ func TestOutboundFederationIgnoresMissingEventWithBadJSONForRoomVersion6(t *test
...
+			// * the message
+			events := result.Events.UntrustedEvents(roomVersion)
+			i := 0
+			for _, ev := range events {
+				must.EqualStr(t, ev.RoomID(), roomID, "unexpected roomID")
+				switch i {
+				case 0:
+					must.EqualStr(t, ev.Type(), "m.room.member", "not a membership event")
+					must.EqualStr(t, *ev.StateKey(), alice.UserID, "unexpected creator")
+					i++
```
DMRobertson, line 410: Would it be easier to loop over `i, ev := range events` here and additionally check that `len(events)` is the correct size?

DMRobertson (approved):
Seems like a faithful port of the sytest to me. Should we remove it from sytest?

I did have a style suggestion, but happy for this to go-in as is!

tests/federation_room_get_missing_events_test.go
S7evinK: I've updated the test. Now we're also checking that we received the message, if the visibility is `shared`

DMRobertson (approved):
LGTM, thanks!

PR #525, opened 2022-10-20
Add presence test from people in 2 different rooms in incremental sync

MatMaul:
This will fail for now, and should be fixed by https://github.com/matrix-org/synapse/pull/14243.

### DIFF
--- a/tests/csapi/sync_test.go
+++ b/tests/csapi/sync_test.go
@@ -382,6 +382,50 @@ func TestSync(t *testing.T) {
 	})
 }
 
+// Test presence from people in 2 different rooms in incremental sync
+func TestPresenceSyncDifferentRooms(t *testing.T) {
+	runtime.SkipIf(t, runtime.Dendrite) // does not yet pass
+
+	deployment := Deploy(t, b.BlueprintOneToOneRoom)
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+	bob := deployment.Client(t, "hs1", "@bob:hs1")
+
+	deployment.RegisterUser(t, "hs1", "charlie", "charlie", false)
+	charlie := deployment.Client(t, "hs1", "@charlie:hs1")
+
+	bobRoomID := alice.CreateRoom(t, struct{}{})
+	charlieRoomID := alice.CreateRoom(t, struct{}{})
+	nextBatch := alice.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(alice.UserID, bobRoomID), client.SyncJoinedTo(alice.UserID, charlieRoomID))
+
+	alice.InviteRoom(t, bobRoomID, bob.UserID)
+	alice.InviteRoom(t, charlieRoomID, charlie.UserID)
+	bob.JoinRoom(t, bobRoomID, nil)
+	charlie.JoinRoom(t, charlieRoomID, nil)
+
+	nextBatch = alice.MustSyncUntil(t,
+		client.SyncReq{Since: nextBatch},
+		client.SyncJoinedTo(bob.UserID, bobRoomID),
+		client.SyncJoinedTo(charlie.UserID, charlieRoomID),
+	)
+
+	reqBody := client.WithJSONBody(t, map[string]interface{}{
+		"presence": "online",
+	})
+	bob.DoFunc(t, "PUT", []string{"_matrix", "client", "v3", "presence", "@bob:hs1", "status"}, reqBody)
+	charlie.DoFunc(t, "PUT", []string{"_matrix", "client", "v3", "presence", "@charlie:hs1", "status"}, reqBody)
+
+	nextBatch = alice.MustSyncUntil(t, client.SyncReq{Since: nextBatch}, func(clientUserID string, sync gjson.Result) error {
+		presence := sync.Get("presence")
+		if len(presence.Get("events").Array()) == 0 {
+			return fmt.Errorf("presence.events is empty: %+v", presence)
+		}
+		usersInPresenceEvents(t, presence, []string{bob.UserID, charlie.UserID})
+		return nil
+	})
+}
+
 func sendMessages(t *testing.T, client *client.CSAPI, roomID string, prefix string, count int) {
 	t.Helper()
 	for i := 0; i < count; i++ {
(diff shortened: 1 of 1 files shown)


tests/csapi/sync_test.go
```diff
@@ -382,6 +382,50 @@ func TestSync(t *testing.T) {
 	})
 }
 
+// Test presence from people in 2 different rooms in incremental sync
+func TestPresenceSyncDifferentRooms(t *testing.T) {
+	runtime.SkipIf(t, runtime.Dendrite) // does not yet pass
```
ShadowJonathan, line 387: I would appreciate it if you could change this to a concrete `FIXME:` comment, as I'm currently changing the whole codebase over as of https://github.com/matrix-org/complement/pull/519.

I think this test would be affected by https://github.com/matrix-org/dendrite/issues/2803, but please do due diligence next time figuring out what makes this test fail, or at least make an issue here noting it.

MatMaul: It's not working, I've added a FIXME referencing your bug since chances that this is related are high indeed.

tests/csapi/sync_test.go
```diff
@@ -382,6 +382,50 @@ func TestSync(t *testing.T) {
...
+	})
+	bob.DoFunc(t, "PUT", []string{"_matrix", "client", "v3", "presence", "@bob:hs1", "status"}, reqBody)
+	charlie.DoFunc(t, "PUT", []string{"_matrix", "client", "v3", "presence", "@charlie:hs1", "status"}, reqBody)
+
+	nextBatch = alice.MustSyncUntil(t, client.SyncReq{Since: nextBatch}, func(clientUserID string, sync gjson.Result) error {
+		presence := sync.Get("presence")
+		if len(presence.Get("events").Array()) == 0 {
+			return fmt.Errorf("presence.events is empty: %+v", presence)
+		}
+		usersInPresenceEvents(t, presence, []string{bob.UserID, charlie.UserID})
```
ShadowJonathan, line 424: Keep in mind that it's entirely possible (and valid) for the server to send the presence events down different syncs, and/or on a later sync than the first one.

This function bails out early with `t.Fatalf`, possibly creating more false negatives than real test results, and is possibly flakey.

What you want is "checking off" the users that have been seen, and only return `nil` when all users have been seen on "that" run, then it can timeout if it doesn't see those users.

ShadowJonathan (changes requested):

tests/csapi/sync_test.go
ShadowJonathan: I think adding this FIXME is alright for now, when that issue is fixed, i'm sure to find this test to un-skip.

tests/csapi/sync_test.go
MatMaul: Indeed, I've just fix that, thanks.

ShadowJonathan (changes requested):

tests/csapi/sync_test.go
```diff
@@ -382,6 +382,65 @@ func TestSync(t *testing.T) {
...
+func TestPresenceSyncDifferentRooms(t *testing.T) {
+	runtime.SkipIf(t, runtime.Dendrite) // FIXME: https://github.com/matrix-org/dendrite/issues/2803
+
+	deployment := Deploy(t, b.BlueprintOneToOneRoom)
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+	bob := deployment.Client(t, "hs1", "@bob:hs1")
+
+	deployment.RegisterUser(t, "hs1", "charlie", "charlie", false)
```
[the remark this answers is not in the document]
ShadowJonathan: And, after running the latest commit (https://github.com/matrix-org/complement/pull/525/commits/459130dea2f5a09c57df3bcdc8ed2af0d6cca0f4) locally with this change, it passes on dendrite :)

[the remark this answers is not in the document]
MatMaul: oh well cool :) thanks for the test, I'll remove the skip.

tests/csapi/sync_test.go
```diff
@@ -382,6 +382,63 @@ func TestSync(t *testing.T) {
...
+	deployment := Deploy(t, b.BlueprintOneToOneRoom)
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+	bob := deployment.Client(t, "hs1", "@bob:hs1")
+
+	deployment.RegisterUser(t, "hs1", "charlie", "charliepassword", false)
+	charlie := deployment.Client(t, "hs1", "@charlie:hs1")
+
+	bobRoomID := alice.CreateRoom(t, struct{}{})
```
DMRobertson, line 396:
```suggestion
	// Alice creates two rooms: one with her and Bob, and a second with her and Charlie.
	bobRoomID := alice.CreateRoom(t, struct{}{})
```

tests/csapi/sync_test.go
```diff
@@ -382,6 +382,63 @@ func TestSync(t *testing.T) {
...
+	bob.JoinRoom(t, bobRoomID, nil)
+	charlie.JoinRoom(t, charlieRoomID, nil)
+
+	nextBatch = alice.MustSyncUntil(t,
+		client.SyncReq{Since: nextBatch},
+		client.SyncJoinedTo(bob.UserID, bobRoomID),
+		client.SyncJoinedTo(charlie.UserID, charlieRoomID),
+	)
+
+	reqBody := client.WithJSONBody(t, map[string]interface{}{
```
DMRobertson, line 411:
```suggestion
	// Bob and Charlie mark themselves as online.
	reqBody := client.WithJSONBody(t, map[string]interface{}{
```

tests/csapi/sync_test.go
```diff
@@ -382,6 +382,63 @@ func TestSync(t *testing.T) {
...
+		"presence": "online",
+	})
+	bob.DoFunc(t, "PUT", []string{"_matrix", "client", "v3", "presence", "@bob:hs1", "status"}, reqBody)
+	charlie.DoFunc(t, "PUT", []string{"_matrix", "client", "v3", "presence", "@charlie:hs1", "status"}, reqBody)
+
+	presenceBob, presenceCharlie := false, false
```
DMRobertson, line 417: We should spell out what we're doing here.
I think I'd call these `seenBobOnline` and `seenCharlieOnline`.

DMRobertson (approved):

PR #527, opened 2022-10-20
Sytest: Add `46direct/01directmessage` tests

ShadowJonathan:
This adds three sytests;
- `./tests/46direct/01directmessage.pl:test "Can recv a device message using /sync",`
- `./tests/46direct/01directmessage.pl:test "Can send a message directly to a device using PUT /sendToDevice",`
- `./tests/46direct/01directmessage.pl:test "Can send a to-device message to two users which both receive it using /sync",`

This also adds the following to `internal/`:
- `SendToDeviceMessages`, to more easily send to-device messages, in the correct format.
- `SyncToDeviceHas` `SyncCheckOpt`, to check for to-device messages in sync responses.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -850,3 +850,48 @@ func SplitMxc(mxcUri string) (string, string) {
 
 	return origin, mediaId
 }
+
+// SendToDeviceMessages sends to-device messages over /sendToDevice/.
+//
+// The messages parameter is nested as follows:
+// user_id -> device_id -> content (map[string]interface{})
+func (c *CSAPI) SendToDeviceMessages(t *testing.T, evType string, messages map[string]map[string]map[string]interface{}) {
+	t.Helper()
+	c.txnID++
+	c.MustDoFunc(
+		t,
+		"PUT",
+		[]string{"_matrix", "client", "v3", "sendToDevice", evType, strconv.Itoa(c.txnID)},
+		WithJSONBody(
+			t,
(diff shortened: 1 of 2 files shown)


internal/client/client.go
```diff
@@ -850,3 +850,48 @@ func SplitMxc(mxcUri string) (string, string) {
 
 	return origin, mediaId
 }
+
+// SendToDeviceMessages sends to-device messages over /sendToDevice/.
+//
+// The messages parameter is nested as follows:
+// user_id -> device_id -> content (map[string]interface{})
+func (c *CSAPI) SendToDeviceMessages(t *testing.T, evType string, messages map[string]map[string]map[string]interface{}) {
```
kegsay, line 858: Given this is a helper function, it would be nice to have `SendToDeviceMessage` (singular not plural) which has a nicer API shape:
```go
func (c *CSAPI) SendToDeviceMessage(t *testing.T, evType, userID, deviceID string, content interface{}) {
```

This can then call through to this function. For most people, they just want to send a to-device message to one device or `*` devices for a single user, which the simpler format makes clearer.

kegsay (approved):
Adding the nicer API function can come later and needn't block this PR.

internal/client/client.go
ShadowJonathan: I originally had this, but I over-engineered it and added 2 additional "nice" functions for just one test, so I thought that was overkill.

PR #533, opened 2022-10-21
Remove `MustDo`

ShadowJonathan:
...and update all usages to `MustDoFunc`, in addition to some miscellaneous code polish surrounding the previous usages (f.e. swapping `struct{}` to `map`)

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -115,7 +115,7 @@ func (c *CSAPI) UploadContent(t *testing.T, fileBody []byte, fileName string, co
 func (c *CSAPI) DownloadContent(t *testing.T, mxcUri string) ([]byte, string) {
 	t.Helper()
 	origin, mediaId := SplitMxc(mxcUri)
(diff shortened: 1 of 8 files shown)


internal/client/client.go
```diff
@@ -472,7 +457,7 @@ func (c *CSAPI) MustDoFunc(t *testing.T, method string, paths []string, opts ...
 	if res.StatusCode < 200 || res.StatusCode >= 300 {
 		defer res.Body.Close()
 		body, _ := ioutil.ReadAll(res.Body)
-		t.Fatalf("CSAPI.MustDoFunc response return non-2xx code: %s - body: %s", res.Status, string(body))
+		t.Fatalf("CSAPI.MustDoFunc %s %s returned non-2xx code: %s - body: %s", method, res.Request.URL.String(), res.Status, string(body))
```
ShadowJonathan, line 460: `MustDo` was a bit more verbose this way, and I remember seeing it useful in the logs the few times that the codebase used `MustDo` over `MustDoFunc`, it's much clearer what exactly complement is doing that way.

kegsay (approved):

PR #534, opened 2022-10-21
Sytest: Add `31sync/09archived` tests

ShadowJonathan:
This adds 8 sytests:
- `./tests/31sync/09archived.pl:test "Archived rooms only contain history from before the user left",`
- `./tests/31sync/09archived.pl:test "Left rooms appear in the leave section of full state sync",`
- `./tests/31sync/09archived.pl:test "Left rooms appear in the leave section of sync",`
- `./tests/31sync/09archived.pl:test "Newly left rooms appear in the leave section of gapped sync",`
- `./tests/31sync/09archived.pl:test "Newly left rooms appear in the leave section of incremental sync",`
- `./tests/31sync/09archived.pl:test "Previously left rooms don't appear in the leave section of sync",`
- `./tests/31sync/09archived.pl:test "We should see our own leave event when rejecting an invite, ".`
- `./tests/31sync/09archived.pl:test "We should see our own leave event, even if history_visibility is " .`

This also brushes up on an csapi-internal function `createFilter`, to have it create less boilerplate.

Tests are skipped on dendrite because of https://github.com/matrix-org/dendrite/issues/1323

### DIFF
--- a/tests/csapi/sync_archive_test.go
+++ b/tests/csapi/sync_archive_test.go
@@ -1,0 +1,436 @@
+package csapi_tests
+
+import (
+	"testing"
+
+	"github.com/tidwall/gjson"
+
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/runtime"
+)
+
+// sytest: Left rooms appear in the leave section of sync
+func TestSyncLeaveSection(t *testing.T) {
+	runtime.SkipIf(t, runtime.Dendrite) // FIXME: https://github.com/matrix-org/dendrite/issues/1323
+
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+
+	includeLeaveFilter := createFilter(t, alice, map[string]interface{}{
+		"room": map[string]interface{}{
+			"include_leave": true,
+		},
+	})
+
+	roomID := alice.CreateRoom(t, map[string]interface{}{})
+
+	alice.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(alice.UserID, roomID))
+
+	s := func(req client.SyncReq) string {
+		req.TimeoutMillis = "0"
+		_, token := alice.MustSync(t, req)
+		return token
+	}
+
+	// Get independently-fetched sync tokens for every subsequent test,
+	// to isolate interference/side effects otherwise possibly caused by
+	// multiple  sync calls with the same token using different filters or
+	// settings.
+	fullStateSince := s(client.SyncReq{
+		Filter: includeLeaveFilter,
+	})
+	incrementalSince := s(client.SyncReq{})
+
+	alice.LeaveRoom(t, roomID)
+
+	// sytest: Left rooms appear in the leave section of sync
+	t.Run("Left rooms appear in the leave section of sync", func(t *testing.T) {
+		// SyncLeftFrom does "active" probing of rooms.leave if userID == clientUserID
+		alice.MustSyncUntil(t, client.SyncReq{
+			Filter: includeLeaveFilter,
+		}, client.SyncLeftFrom(alice.UserID, roomID))
+	})
+
+	// sytest: Left rooms appear in the leave section of full state sync
+	t.Run("Left rooms appear in the leave section of full state sync", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


tests/csapi/sync_archive_test.go
```diff
@@ -0,0 +1,436 @@
...
+		Type:     "m.room.member",
+		StateKey: b.Ptr(bob.UserID),
+		Content: map[string]interface{}{
+			"membership": "join",
+			"filler":     "Welcome back, my friends, to the show that never ends...",
+		},
+	})
+
+	for i := 0; i < 20; i++ {
+		alice.SendEventSynced(t, roomToSpam, b.Event{
```
squahtx, line 283: nit: in the sytest, it's bob who sends the spam. It looks like it doesn't really matter though.
```suggestion
		bob.SendEventSynced(t, roomToSpam, b.Event{
```

tests/csapi/sync_archive_test.go
```diff
@@ -0,0 +1,436 @@
...
+	bob.SendEventSynced(t, roomToLeave, b.Event{
+		Type:     "m.room.member",
+		StateKey: b.Ptr(bob.UserID),
+		Content: map[string]interface{}{
+			"membership": "join",
+			"filler":     "Welcome back, my friends, to the show that never ends...",
+		},
+	})
+
+	for i := 0; i < 20; i++ {
```
squahtx, line 282: I'd like to preserve the comment from sytest.
```suggestion
	// Pad out the timeline with filler messages to create a "gap" between
	// this sync and the next.
	for i := 0; i < 20; i++ {
```

tests/csapi/sync_archive_test.go
```diff
@@ -0,0 +1,436 @@
...
+			t.Fatalf("rooms.leave is not an object in syncRes")
+		}
+
+		if len(roomLeaveRes.Map()) != 0 {
+			t.Fatalf("Expected no rooms in 'leave' state")
+		}
+	}
+}
+
+// sytest: We should see our own leave event, even if history_visibility is
```
squahtx, line 306: typo:
```suggestion
// sytest: We should see our own leave event, even if history_visibility is restricted (SYN-662)
```

tests/csapi/sync_archive_test.go
```diff
@@ -0,0 +1,436 @@
...
+	if event.Get("type").Str != "m.room.member" {
+		t.Fatalf("expected timeline event was not m.room.member")
+	} else if event.Get("sender").Str != alice.UserID {
+		t.Fatalf("sender of membership event was not alice")
+	} else if event.Get("content.membership").Str != "leave" {
+		t.Fatalf("membership event was not leave")
+	}
+}
+
+// sytest: We should see our own leave event when rejecting an invite,
```
squahtx, line 391: typo:
```suggestion
// sytest: We should see our own leave event when rejecting an invite, even if history_visibility is restricted (riot-web/3462)
```

squahtx (commented):
Looks roughly equivalent to the original sytests.

Some minor nits.

tests/csapi/sync_archive_test.go
```diff
@@ -0,0 +1,436 @@
...
+	s := func(req client.SyncReq) string {
+		req.TimeoutMillis = "0"
+		_, token := alice.MustSync(t, req)
+		return token
+	}
+
+	// Get independently-fetched sync tokens for every subsequent test,
+	// to isolate interference/side effects otherwise possibly caused by
+	// multiple  sync calls with the same token using different filters or
+	// settings.
```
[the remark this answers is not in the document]
ShadowJonathan: It is, will add 👍

tests/csapi/sync_archive_test.go
ShadowJonathan: Not a typo, it's intentional, else `sytest-coverage` will not catch this test.

It is based off of sytest.list, which is based off of a grep from the `sytest` repo, which contains a newline exactly here.

This means that the line needs to be the *exact* one (in quotes) from sytest.list, else it will not be recognised.

Adding this bit on a newline here could work, though.

tests/csapi/sync_archive_test.go
ShadowJonathan: Ditto to my previous comment, not a typo

tests/csapi/sync_archive_test.go
squahtx: That's unfortunate. Let's add the rest on a new line then?

squahtx (approved):
Thanks for porting the tests!

PR #542, opened 2022-10-29
Add `MatchJSON`, `MatchJSONBytes` and `MatchGJSON`

### LINKED ISSUE #528: Adding `must.MatchGJSON` and `must.MatchJSON`
Some APIs return `gjson.Result`, which is then meant to be consumed with `Get`-like apis, to extract data.

However, sometimes it's useful to use such a result in a `must.MatchResponse`-like fashion, to perform multiple standard operations on the `gjson.Result` itself, a-la `[]JSON` in `match.HTTPResponse`

So, I think an api like the following would help a lot of boilerplate go away;
```golang
must.MatchGJSON(t, result, []match.JSON{
	match.JSONKeyEqual("device_id", deviceID),
	match.JSONKeyEqual("display_name", "device display"),
})
```

Because an equivalent check without the above could/would look like the following;

```golang
deviceIdRes := result.Get("device_id")
if !deviceIdRes.Exists() 
   || deviceIdRes.Type != gjson.String 
   || deviceIdRes.Str != deviceID {
  return false
}

displayNameRes := result.Get("display_name")
if !displayNameRes.Exists() 
   || displayNameRes.Type != gjson.String 
   || displayNameRes.Str != "device display" {
  return false
}
```

This'd be in addition to a `must.MatchJSON`, which performs the same operations as above, but on a `string` (or alternatively `[]byte`) containing raw json.

ShadowJonathan:
Resolves #528

This adds three convenience methods, which call eachother for deduplication.

The effective round-trip from gjson.Result.Str to byteslice and back is unfortunate, but it's how `match.JSON` is defined.

### DIFF
--- a/internal/must/must.go
+++ b/internal/must/must.go
@@ -132,6 +132,35 @@ func MatchFederationRequest(t *testing.T, fedReq *gomatrixserverlib.FederationRe
 	}
 }
 
+// MatchGJSON performs JSON assertions on a gjson.Result object.
+func MatchGJSON(t *testing.T, jsonResult gjson.Result, matchers ...match.JSON) {
+	t.Helper()
+
+	MatchJSON(t, jsonResult.Str, matchers...)
+}
+
+// MatchJSON performs JSON assertions on a raw JSON string.
+func MatchJSON(t *testing.T, json string, matchers ...match.JSON) {
+	t.Helper()
+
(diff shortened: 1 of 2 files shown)


kegsay (approved):
Nice improvement; it would be good to see other places where we can now call match functions.

Issue #529, opened 2022-10-21
Add `(*docker.Deployment).NewUser`


ShadowJonathan:
...as an easier shorthand to `RegisterUser`.

I find the `RegisterUser` become boilerplate-y, and I've seen a few times where I (and others) trip up over making a long enough password to satisfy most homeservers.

So, I suggest making a new function `NewUser(t *testing.T, localpart, hs string) *client.CSAPI`, which takes care of all of that.

PR #543, opened 2022-10-29
Add `NewUser` convenience method

ShadowJonathan:
Resolves https://github.com/matrix-org/complement/issues/529

This function is to make some bits of the codebase less boilerplate-y, and have test writers not shoot themselves in the foot over and over with the password requirements on some servers.


internal/docker/deployment.go
```diff
@@ -112,10 +132,10 @@ func (d *Deployment) RegisterUser(t *testing.T, hsName, localpart, password stri
 }
 
 // Restart a deployment.
-func (dep *Deployment) Restart(t *testing.T) error {
+func (d *Deployment) Restart(t *testing.T) error {
 	t.Helper()
-	for _, hsDep := range dep.HS {
-		err := dep.Deployer.Restart(hsDep, dep.Config)
+	for _, hsDep := range d.HS {
+		err := d.Deployer.Restart(hsDep, d.Config)
```
ShadowJonathan, line 138: This change was brought to you by goland bugging about different receiver names

internal/docker/deployment.go
```diff
@@ -80,6 +82,24 @@ func (d *Deployment) Client(t *testing.T, hsName, userID string) *client.CSAPI {
...
+	rand.Read(b)
+	return fmt.Sprintf("%x", b)[:length]
+}
+
+// NewUser creates a new user as a convenience method to RegisterUser.
+//
+//It registers the user with a random password, and without admin privileges.
+func (d *Deployment) NewUser(t *testing.T, localpart, hs string) *client.CSAPI {
+	// Dendrite requires at least 8 characters for a password, we choose 12 here in case they increase that.
+	password := randomString(12)
```
kegsay, line 98: Please don't do this. We don't need the security for these accounts. Make them deterministic to make it easier for developers to introspect servers. For example, I've seen cases where developers keep the HS containers running then point a local element web at it in order to check what is on the account - they need to login with a password and can do so because it is deterministic. Making this random makes this impossible.

kegsay (changes requested):
Use the same deterministic code as https://github.com/matrix-org/complement/blob/main/internal/instruction/runner.go#L545 please.

internal/docker/deployment.go
ShadowJonathan: > Use the same deterministic code as https://github.com/matrix-org/complement/blob/main/internal/instruction/runner.go#L545 please.

Thanks for mentioning that, I'll obviously do that, though I wasn't aware of that bit of code before 👍

internal/docker/deployment.go
```diff
@@ -80,6 +80,13 @@ func (d *Deployment) Client(t *testing.T, hsName, userID string) *client.CSAPI {
 	return client
 }
 
+// NewUser creates a new user as a convenience method to RegisterUser.
+//
+//It registers the user with a random password, and without admin privileges.
```
kegsay, line 85:
```suggestion
//It registers the user with a deterministic password, and without admin privileges.
```

kegsay (approved):

ShadowJonathan:
It seems the CI failure is a flake, could you re-run?

PR #545, opened 2022-10-29
Sytest: Add `10apidoc/36room-levels` tests

ShadowJonathan:
This adds three sytests:
- `./tests/10apidoc/36room-levels.pl:test "GET /rooms/:room_id/state/m.room.power_levels can fetch levels",`
- `./tests/10apidoc/36room-levels.pl:test "PUT /rooms/:room_id/state/m.room.power_levels can set levels",`
- `./tests/10apidoc/36room-levels.pl:test "PUT power_levels should not explode if the old power levels were empty",`

This skips a test called "Both GET and PUT work" because its sole existence in sytest appears to be to prove the ability to set and get powerlevels, for other test's requirements.

### DIFF
--- a/tests/csapi/power_levels_test.go
+++ b/tests/csapi/power_levels_test.go
@@ -76,10 +76,21 @@ func TestPowerLevels(t *testing.T) {
 				match.JSONKeyTypeEqual("events_default", gjson.Number),
 				match.JSONKeyTypeEqual("users_default", gjson.Number),
 
-				match.JSONKeyTypeEqual("events", gjson.JSON),
+				match.JSONMapEach("events", func(k, v gjson.Result) error {
+					if v.Type != gjson.Number {
+						return fmt.Errorf("key %s is not a number", k.Str)
+					} else {
+						return nil
+					}
+				}),
 
-				match.JSONKeyTypeEqual("users", gjson.JSON),
-				match.JSONKeyTypeEqual("users."+client.GjsonEscape(alice.UserID), gjson.Number),
+				match.JSONMapEach("users", func(k, v gjson.Result) error {
+					if v.Type != gjson.Number {
+						return fmt.Errorf("key %s is not a number", k.Str)
+					} else {
+						return nil
+					}
+				}),
 
 				func(body []byte) error {
 					userDefault := int(gjson.GetBytes(body, "users_default").Num)
@@ -101,7 +112,8 @@ func TestPowerLevels(t *testing.T) {
 		PLContent := map[string]interface{}{
 			"invite": 100.0,
 			"users": map[string]interface{}{
-				alice.UserID: 100.0,
+				alice.UserID:                    100.0,
+				"@random-other-user:their.home": 20.0,
 			},
 		}
 
--- a/tests/csapi/power_levels_test.go
+++ b/tests/csapi/power_levels_test.go
@@ -4,6 +4,9 @@ import (
 	"testing"
 
 	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
 )
 
 // This test ensures that an authorised (PL 100) user is able to modify the users_default value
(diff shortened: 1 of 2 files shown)


tests/csapi/power_levels_test.go
```diff
@@ -40,3 +43,83 @@ func TestDemotingUsersViaUsersDefault(t *testing.T) {
...
+func TestPowerLevels(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+
+	roomID := alice.CreateRoom(t, map[string]interface{}{})
+
+	// sytest: GET /rooms/:room_id/state/m.room.power_levels can fetch levels
+	t.Run("GET /rooms/:room_id/state/m.room.power_levels can fetch levels", func(t *testing.T) {
```
reivilibre, line 56: This test appears to be missing `assert_json_keys( $body, qw( ban kick redact state_default events_default users_default ));` and a few other checks
https://github.com/matrix-org/sytest/blob/fd4f4132913e9575471af515f67c633d2e3123d3/tests/10apidoc/36room-levels.pl#L17-L34

tests/csapi/power_levels_test.go
```diff
@@ -40,3 +43,83 @@ func TestDemotingUsersViaUsersDefault(t *testing.T) {
...
+		must.MatchResponse(t, res, match.HTTPResponse{
+			StatusCode: 200,
+			JSON: []match.JSON{
+				match.JSONKeyPresent("users"),
+			},
+		})
+	})
+
+	// sytest: PUT /rooms/:room_id/state/m.room.power_levels can set levels
+	t.Run("PUT /rooms/:room_id/state/m.room.power_levels can set levels", func(t *testing.T) {
```
reivilibre, line 69: This should be setting the PL for another user. It's very likely alice is already PL 100.
Further you should check that the event comes down sync with the PL you expect?

Further, is this even calling the endpoint mentioned in the test name? (I'm not sure whether it's using a more generic endpoint for sending events — possibly it's fine, I just don't find it as obvious as SyTest :eyes: .)

https://github.com/matrix-org/sytest/blob/fd4f4132913e9575471af515f67c633d2e3123d3/tests/10apidoc/36room-levels.pl#L38-L72

reivilibre: > Further, is this even calling the endpoint mentioned in the test name? (I'm not sure whether it's using a more generic endpoint for sending events — possibly it's fine, I just don't find it as obvious as SyTest eyes .)

On this: I'd be tempted to call the endpoint manually (rather than relying on `SendEventSynced` doing it for you), personally, but maybe worth checking with a Complement expert.

ShadowJonathan: > On this: I'd be tempted to call the endpoint manually

It calls `/state/m.room.power_levels`, which `SendEventSynced` calls (when it sees its a state event), so its the same endpoint.

Otherwise good points for improvement 👍

tests/csapi/power_levels_test.go
```diff
@@ -40,3 +46,125 @@ func TestDemotingUsersViaUsersDefault(t *testing.T) {
...
+		// note: these need to be floats to allow a roundtrip comparison
+		PLContent := map[string]interface{}{
+			"invite": 100.0,
+			"users": map[string]interface{}{
+				alice.UserID: 100.0,
```
reivilibre, line 104: I think this was missed from last time: I'd advise adding another user ID here (sytest used `@random-other-user:their.home` at power level 20) — that way, we can test that the server allows you to set *someone else's* power level (!)

reivilibre (changes requested):
minor points, other than that I think we should have the SyTest PR to remove the SyTest before merging this

tests/csapi/power_levels_test.go
```diff
@@ -40,3 +46,125 @@ func TestDemotingUsersViaUsersDefault(t *testing.T) {
...
+	// sytest: GET /rooms/:room_id/state/m.room.power_levels can fetch levels
+	t.Run("GET /rooms/:room_id/state/m.room.power_levels can fetch levels", func(t *testing.T) {
+		// Test if the old state still exists
+		res := alice.MustDoFunc(t, "GET", []string{"_matrix", "client", "v3", "rooms", roomID, "state", "m.room.power_levels"})
+
+		// note: before v10 we technically cannot assume that powerlevel integers are json numbers,
+		//  as they can be both strings and numbers.
+		// However, for this test, we control the test environment,
+		//  and we will assume the server is sane and give us powerlevels as numbers,
+		//  and if it doesn't, that's an offense worthy of a frown.
```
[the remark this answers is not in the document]
reivilibre: (or put another way: Complement should not be afraid to test 'good practices' that go a little bit above and beyond the bare minimum in the spec, as long as there's no good reason why a homeserver implementer would do the opposite. In this case, I can't think of any valid reason why a homeserver should produce string power levels on its own.)

tests/csapi/power_levels_test.go
```diff
@@ -40,3 +46,125 @@ func TestDemotingUsersViaUsersDefault(t *testing.T) {
...
+			JSON: []match.JSON{
+				match.JSONKeyTypeEqual("ban", gjson.Number),
+				match.JSONKeyTypeEqual("kick", gjson.Number),
+				match.JSONKeyTypeEqual("redact", gjson.Number),
+				match.JSONKeyTypeEqual("state_default", gjson.Number),
+				match.JSONKeyTypeEqual("events_default", gjson.Number),
+				match.JSONKeyTypeEqual("users_default", gjson.Number),
+
+				match.JSONKeyTypeEqual("events", gjson.JSON),
+
```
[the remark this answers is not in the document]
ShadowJonathan: Oh sorry, I must've glanced over that, will add 👍

tests/csapi/power_levels_test.go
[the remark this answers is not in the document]
reivilibre: Considering there'd be no reason for it to do so and it's only 'allowed' for compatibility, I really can't see any way to defend allowing a server doing this by default in a clean test environment such as Complement.

PR #550, opened 2022-11-04
MatchGJSON: `jsonResult.Raw` instead of `jsonResult.Str`

ShadowJonathan:
Encountered while working on another test, seems like I slipped in a bug with this


squahtx:
Note that this is the same change as #556.

anoadragon453:
https://github.com/matrix-org/complement/pull/556 has some more context if it's helpful for review.

anoadragon453 (approved):

kegsay (approved):

PR #551, opened 2022-11-04
Add regression test for room summaries

S7evinK:
A simple test to verify room summaries are correctly send down sync.


neilalexander (approved):

ShadowJonathan:
Please add a FIXME with a synapse issue number if this does not run on synapse, looking at the spec, synapse simply does not follow it, and thus it needs one.

PR #553, opened 2022-11-07
Create & use filter with enabled LL when testing room summaries

S7evinK:
Synapse seems to need a filter with enabled `lazy_load_members` to add room summaries. This now creates that filter and uses it in `/sync` requests.

### DIFF
--- a/tests/csapi/sync_test.go
+++ b/tests/csapi/sync_test.go
@@ -436,18 +436,34 @@ func TestPresenceSyncDifferentRooms(t *testing.T) {
 }
 
 func TestRoomSummary(t *testing.T) {
-	runtime.SkipIf(t, runtime.Synapse) // Currently more of a Dendrite test, so skip on Synapse
 	deployment := Deploy(t, b.BlueprintOneToOneRoom)
 	defer deployment.Destroy(t)
 	alice := deployment.Client(t, "hs1", "@alice:hs1")
 	bob := deployment.Client(t, "hs1", "@bob:hs1")
 
-	_, aliceSince := alice.MustSync(t, client.SyncReq{TimeoutMillis: "0"})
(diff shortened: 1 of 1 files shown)


reivilibre:
> Synapse seems to need a filter with enabled lazy_load_members to add room summaries.

maybe a more pressing question: but should it?

It feels like we should ensure the spec is clear about what the requirements are here — it smells wrong if Dendrite and Synapse don't do the same thing

S7evinK:
As @ShadowJonathan  mentioned in https://github.com/matrix-org/matrix-spec/issues/1325, the spec very vague about this field.

Dendrite adds room summaries if there are membership changes, not only if a LL filter is supplied.

Synapse has [this](https://github.com/matrix-org/synapse/blob/a5fcdea090c2396c30dd07c357ce4d9c90004c34/synapse/handlers/sync.py#L2365-L2388), so is only adding the field if it's an not "out of band" AND LL is enable and there are membership changes. (Q: What is "out of band" in this case?)

ShadowJonathan:
For the record, I would really not add anything in complement that cannot be backed up by wording in the spec, and I hope @kegsay can agree with that.

reivilibre:
yeah, if the spec is too vague to be useful then, in my opinion, it'd be nice to feed some rigidity back into the spec.
With that said, I appreciate the spec process can be extremely slow. Maybe worth raising in #matrix-spec anyway? Possibly it could be settled with a 'clarification' by documenting existing expectations. At least having some sort of PR open (or failing that, an issue link that can be added to the test) for the spec would be a good start to support this change to the test.

squahtx:
Going to remove this from the Synapse review queue until we have more clarity on the spec side. At the very least, some sort of spec PR or issue to link against would be good, as suggested by @reivilibre.

kegsay:
Closing as stale.

PR #554, opened 2022-11-10
Sytest: Add `30rooms/60version_upgrade` tests

ShadowJonathan:
This adds 6 sytests:
- `./tests/30rooms/60version_upgrade.pl:   test "$user_type user has push rules copied to upgraded room",`
- `./tests/30rooms/60version_upgrade.pl:test "/upgrade creates a new room",`
- `./tests/30rooms/60version_upgrade.pl:test "/upgrade moves aliases to the new room",`
- `./tests/30rooms/60version_upgrade.pl:test "/upgrade is rejected if the user can't send state events",`
- `./tests/30rooms/60version_upgrade.pl:test "/upgrade of a bogus room fails gracefully",`
- `./tests/30rooms/60version_upgrade.pl:test "/upgrade to an unknown version is rejected",`

However, this also ignores the remaining 13 sytests, due to non-formalisation of what they're testing for in the spec, see https://github.com/matrix-org/matrix-spec/issues/1324 and https://github.com/matrix-org/matrix-spec/issues/456 for more details on that.

This also adds some extra logging in `MustSyncUntil` to make it clear which checker exactly it is printing errors for, this is useful when the errors inside the checker itself are ambiguous or not helpful with identifying the checker.


ShadowJonathan:
CI is failing because of https://github.com/matrix-org/complement/pull/550

ShadowJonathan:
@clokep those are unrelated flakes, could you kick off a rerun?

DMRobertson:
I did another rerun and this is green.

tests/csapi/room_upgrade_test.go
```diff
@@ -0,0 +1,309 @@
...
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+
+	roomID := alice.CreateRoom(t, map[string]interface{}{
+		"preset": "public_chat",
+	})
+
+	_, tombstoneSinceToken := alice.MustSync(t, client.SyncReq{TimeoutMillis: "0"})
+	_, sinceToken := alice.MustSync(t, client.SyncReq{TimeoutMillis: "0"})
```
DMRobertson, line 27: A bit odd to sync twice rather than copy the token into a new var

DMRobertson: Might be clearer if this was a `MustSyncUntil` which waits for Alice to see her membership of the new room?

tests/csapi/room_upgrade_test.go
```diff
@@ -0,0 +1,309 @@
...
+	replacementRoom := upgradeJson.Get("replacement_room").Str
+
+	if roomID == replacementRoom {
+		t.Fatalf("Rooms are the same (%s)", roomID)
+	}
+
+	// Check for and fetch the tombstone event ID
+	tombstoneEventId := ""
+	alice.MustSyncUntil(t, client.SyncReq{Since: tombstoneSinceToken}, client.SyncTimelineHas(roomID, func(ev gjson.Result) bool {
+		if ev.Get("type").Str == "m.room.tombstone" && ev.Get("state_key").Exists() {
```
DMRobertson, line 54: Can we enforce that tombstone events should have a state key equal to the tmpty string, please?

tests/csapi/room_upgrade_test.go
```diff
@@ -0,0 +1,309 @@
...
+
+		return false
+	}))
+
+	// Convenience method to check for state event
+	stateExists := func(evType string) func(gjson.Result) bool {
+		return func(ev gjson.Result) bool {
+			return ev.Get("type").Str == evType && ev.Get("state_key").Exists()
+		}
+	}
```
DMRobertson, line 76: Again, can we be precise about the values of state keys here (not just their existence).

tests/csapi/room_upgrade_test.go
```diff
@@ -0,0 +1,309 @@
...
+						t.Errorf("predecessor room id did not match: got %s, want %s", predecessor.Get("room_id").Str, roomID)
+					}
+
+					if predecessor.Get("event_id").Str != tombstoneEventId {
+						t.Errorf("predecessor tombstone event id did not match: got %s, want %s", predecessor.Get("event_id").Str, tombstoneEventId)
+					}
+
+					return true
+				}
+			}
```
DMRobertson, line 102: The old sytest used to assert that the upgraded room has the requested room version. I don't think this is done here(?)

tests/csapi/room_upgrade_test.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	// Note: we cannot check for the following state here, as it's only prescribed as an "implementation detail" by the spec:
+	// - m.room.guest_access
+	// - m.room.history_visibility
+	// - m.room.join_rules
+	alice.MustSyncUntil(t, client.SyncReq{Since: sinceToken}, client.SyncJoinedTo(alice.UserID, replacementRoom),
+		client.SyncTimelineHas(replacementRoom, stateExists("m.room.create")),
+		client.SyncTimelineHas(replacementRoom, stateExists("m.room.member")),
+		client.SyncTimelineHas(replacementRoom, stateExists("m.room.power_levels")),
```
DMRobertson, line 85: The old sytest used to assert that the create event was the first thing to appear in the timeline. Don't mind dropping that from this test, but just wanted to highlight that it's missing.

DMRobertson (changes requested):
I've only looked at the first sytest. Some thoughts in passing. Will see if I can get to the others next week.

tests/federation_room_upgrade_test.go
```diff
@@ -0,0 +1,123 @@
...
+	bob := deployment.Client(t, "hs1", "@bob:hs1")
+	charlie := deployment.Client(t, "hs2", "@charlie:hs2")
+
+	doTestWith := func(t *testing.T, user *client.CSAPI) {
+		roomID := alice.CreateRoom(t, map[string]interface{}{
+			"preset": "public_chat",
+		})
+
+		user.JoinRoom(t, roomID, []string{"hs1"})
+		alice.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(user.UserID, roomID))
```
DMRobertson, line 34: Why is _alice_ syncing? Presumably this should be `user.MustSyncUntil`?

tests/federation_room_upgrade_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+		user.MustDoFunc(t, "PUT", []string{"_matrix", "client", "v3", "pushrules", "global", "room", roomID},
+			client.WithJSONBody(t, map[string]interface{}{
+				"actions": []string{"notify"},
+			}),
+		)
+
+		_, sinceToken := alice.MustSync(t, client.SyncReq{TimeoutMillis: "0"})
```
DMRobertson, line 42: Why is alice syncing here? Please make the condition explicit

tests/federation_room_upgrade_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+		// Note: we cannot check for the following state here, as it's only prescribed as an "implementation detail" by the spec:
+		// - m.room.guest_access
+		// - m.room.history_visibility
+		// - m.room.join_rules
+		alice.MustSyncUntil(t, client.SyncReq{Since: sinceToken}, client.SyncJoinedTo(alice.UserID, replacementRoom),
+			client.SyncTimelineHas(replacementRoom, stateExists("m.room.create")),
+			client.SyncTimelineHas(replacementRoom, stateExists("m.room.member")),
+			client.SyncTimelineHas(replacementRoom, stateExists("m.room.power_levels")),
+		)
```
DMRobertson, line 78: This isn't relevant to copying push rules. Haven't we checked this in some other test?

tests/federation_room_upgrade_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+		user.JoinRoom(t, replacementRoom, []string{"hs1"})
+		alice.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(user.UserID, replacementRoom))
```
DMRobertson, line 81: Presumably this should be the `user` syncing, since `user` is about to make a bunch of push rules lookups below

tests/federation_room_upgrade_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+		success := false
+
+		// We try to fetch the new alias 10 times, each with 1s backoff
```
DMRobertson, line 85: Not fetching an alias here

DMRobertson (changes requested):

tests/csapi/room_upgrade_test.go
ShadowJonathan: Using the same since-token creates - for my feeling - Undefined Behaviour, and these exist to detect two different things in disjoint "timelines";
- One is looking for the tombstone event, which can happen before or after creating the new room.
- One is looking for the new room, which can happen before or after the tombstone event is emitted.

This might look like a clear case of using the same since-token twice, but I've encountered subtle weird behaviour in the past with using the same sync token twice, and as this kind of behaviour is not explicitly defined in the spec (it only talks about and assumes that one uses a since token exactly once), i'm being conservative here, creating two different since tokens to use for different "spans of time" to test.

tests/csapi/room_upgrade_test.go
ShadowJonathan: There's not exactly an easy way to assert that, and honestly I'd use a complement-library-defined function for it instead of assembling one here.

I'll drop a todo 👍

tests/csapi/room_upgrade_test.go
```diff
@@ -0,0 +1,309 @@
...
+	alice.SendEventSynced(t, roomID, b.Event{
+		Type:     "m.room.canonical_alias",
+		StateKey: b.Ptr(""),
+		Content: map[string]interface{}{
+			"alias":       localRoomAliasMain,
+			"alt_aliases": []interface{}{localRoomAliasAlt},
+		},
+	})
+
+	_, sinceToken := alice.MustSync(t, client.SyncReq{TimeoutMillis: "0"})
```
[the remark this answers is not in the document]
ShadowJonathan: We are syncing explicitly over the span of time of the room upgrade, I find this cleaner to test for.

tests/csapi/room_upgrade_test.go
```diff
@@ -0,0 +1,309 @@
...
+
+	resp := alice.MustDoFunc(
+		t,
+		"POST",
+		[]string{"_matrix", "client", "v3", "rooms", roomID, "upgrade"},
+		client.WithJSONBody(t, map[string]string{
+			"new_version": "6",
```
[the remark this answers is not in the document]
ShadowJonathan: No particular reason, but I think that we should really start to use a more modern room number than 2.

tests/csapi/room_upgrade_test.go
```diff
@@ -0,0 +1,309 @@
...
+				}
+			}
+
+			if i == 9 {
+				t.Error("room_id did not match replacement room after 10 tries")
+			} else {
+				t.Logf("try %d failed, sleeping 1 second...", i+1)
+				time.Sleep(1 * time.Second)
+			}
+		}
```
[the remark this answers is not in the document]
ShadowJonathan: client.WithRetryUntil, but yes, I wasn't aware of this, will rewrite 👍

tests/federation_room_upgrade_test.go
ShadowJonathan: If `user` asserts that it has joined the room, the join might not have propagated to `alice`'s homeserver, so when `alice` asserts it, it must've registered on both `user` and `alice`'s homeserver, which means its safe to assume this going forward, whereas its not when only `user` sees it.

It might look small, but it could result in a flake if `alice`'s homeserver is just slow enough.

tests/federation_room_upgrade_test.go
ShadowJonathan: This is because alice upgrades the room, and will be the first to see the new events and its results. IIRC this is also what the sytest did, incrementally syncing from a room's upgrader to observe its results there.

tests/federation_room_upgrade_test.go
ShadowJonathan: No, that's why we do it here now.

FWIW the original sytest also asserts this by their common room upgrade function, so its implicit that this should be tested here as well.

tests/federation_room_upgrade_test.go
ShadowJonathan: Technically correct, though `alice` syncing would assert that the join came through on her server as well.

tests/federation_room_upgrade_test.go
ShadowJonathan: Copy-paste leftover, oops

tests/csapi/room_upgrade_test.go
DMRobertson: Okay. Can you please spell this out with a comment: something to the effect of "the tombstone and new room can appear down /sync in either order, so we check for these using separate /sync tracks". Can we also have a more descriptive name for the second `sinceToken` which describes what it is being used to wait for, please.

tests/csapi/room_upgrade_test.go
DMRobertson: I also note in passing that the spec is unusually precise about the expected order of events that /createRoom should create: https://spec.matrix.org/v1.5/client-server-api/#post_matrixclientv3createroom . So it is surprising that we don't have the same specificity when it comes to the initial batch of events in an upgraded room. https://spec.matrix.org/v1.5/client-server-api/#server-behaviour-16

tests/federation_room_upgrade_test.go
DMRobertson: > when alice asserts it, it must've registered on both user and alice's homeserver,

while this is true, it's not obvious to the reader at first glance.

Can we either
- have a comment along the lines of your own above, or
- change (2) to have `user` wait for the join, and then  change alice's sync on +43 to wait for `user`'s join before upgrading the room.

tests/federation_room_upgrade_test.go
DMRobertson: My objection isn't that the user that's syncing is Alice; it's that I don't know what we are waiting for Alice to see down /sync before upgrading the room. That's why I suggest splitting out `user` and `alice`'s syncs in https://github.com/matrix-org/complement/pull/554/files#r1047211314 --- it makes it easier to see what each client is waiting to see before the scenario progresses.

sytest.ignored.list
```diff
@@ -16,6 +16,21 @@ Can backfill purged history
...
+/upgrade preserves direct room state
+/upgrade preserves room federation ability
+/upgrade preserves the power level of the upgrading user in old and new rooms
+/upgrade restricts power levels in the old room
+/upgrade restricts power levels in the old room when the old PLs are unusual
+/upgrade should preserve room visibility for $vis rooms
+
 # Tests deprecated endpoints
 Tags appear in the v1 /initialSync
 Tags appear in the v1 /events stream
```
MadLittleMods, line 36: @ShadowJonathan What's the status with this other unresolved review feedback?

kegsay:
Closing as stale.

PR #225, opened 2021-11-17
Test against the stable endpoint from MSC2946

clokep:
See matrix-org/synapse#11329 for the corresponding synapse PR. I've added this to the dendrite blacklist and also removed the old tests against `/spaces`.

Reviewable commit-by-commit.

PR #559, opened 2022-11-16
Use stable MSC3030 `/timestamp_to_event` endpoints

MadLittleMods:
Use stable MSC3030 `/timestamp_to_event` endpoints

Synapse changes: https://github.com/matrix-org/synapse/pull/14471

### DIFF
--- a/tests/msc2716_test.go
+++ b/tests/msc2716_test.go
@@ -1126,16 +1126,6 @@ func TestImportHistoricalMessages(t *testing.T) {
 	})
 }
 
-var txnCounter int = 0
-
-func getTxnID(prefix string) (txnID string) {
-	txnId := fmt.Sprintf("%s-%d", prefix, txnCounter)
-
-	txnCounter++
-
-	return txnId
-}
-
 func makeInterfaceSlice(slice []string) []interface{} {
 	interfaceSlice := make([]interface{}, len(slice))
 	for i := range slice {
--- a/tests/msc2716_test.go
+++ b/tests/msc2716_test.go
@@ -1164,42 +1164,6 @@ func includes(needle string, haystack []string) bool {
 	return false
 }
 
-func fetchUntilMessagesResponseHas(t *testing.T, c *client.CSAPI, roomID string, check func(gjson.Result) bool) {
-	t.Helper()
-	start := time.Now()
-	checkCounter := 0
-	for {
-		if time.Since(start) > c.SyncUntilTimeout {
-			t.Fatalf("fetchUntilMessagesResponseHas timed out. Called check function %d times", checkCounter)
-		}
-
-		messagesRes := c.MustDoFunc(t, "GET", []string{"_matrix", "client", "v3", "rooms", roomID, "messages"}, client.WithContentType("application/json"), client.WithQueries(url.Values{
-			"dir":   []string{"b"},
-			"limit": []string{"100"},
-		}))
-		messsageResBody := client.ParseJSON(t, messagesRes)
-		wantKey := "chunk"
-		keyRes := gjson.GetBytes(messsageResBody, wantKey)
-		if !keyRes.Exists() {
-			t.Fatalf("missing key '%s'", wantKey)
-		}
-		if !keyRes.IsArray() {
-			t.Fatalf("key '%s' is not an array (was %s)", wantKey, keyRes.Type)
-		}
-
-		events := keyRes.Array()
-		for _, ev := range events {
-			if check(ev) {
-				return
-			}
-		}
-
-		checkCounter++
-		// Add a slight delay so we don't hammmer the messages endpoint
-		time.Sleep(500 * time.Millisecond)
-	}
-}
-
 // Paginate the /messages endpoint until we find all of the expectedEventIds
 // (order does not matter). If any event in denyListEventIDs is found, an error
 // will be thrown.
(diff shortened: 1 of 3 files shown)


tests/msc3030_test.go
```diff
@@ -118,7 +118,7 @@ func TestJumpToDateEndpoint(t *testing.T) {
 			// Make the `/timestamp_to_event` request from Bob's perspective (non room member)
```
MadLittleMods, line 118: Now that MSC3030 is merged and stable, do we also need to remove the `+build msc3030` build tags and change this to Synapse only instead?

clokep: That's what I've been doing. 🤷

tests/room_timestamp_to_event_test.go
```diff
@@ -300,19 +300,20 @@ func mustCheckEventisReturnedForTime(t *testing.T, c *client.CSAPI, roomID strin
...
+	// Only allow a 200 response meaning we found an event or when no `expectedEventId`, a
+	// 404 meaning we didn't find anything. Other status codes will throw and assumed to
+	// be application errors.
 	actualEventId := ""
 	if timestampToEventRes.StatusCode == 200 {
 		actualEventId = client.GetJSONFieldStr(t, timestampToEventResBody, "event_id")
-	} else if timestampToEventRes.StatusCode != 404 {
-		t.Fatalf("mustCheckEventisReturnedForTime: /timestamp_to_event request failed with status=%d", timestampToEventRes.StatusCode)
+	} else if timestampToEventRes.StatusCode != 404 || (timestampToEventRes.StatusCode == 404 && expectedEventId != "") {
+		t.Fatalf("mustCheckEventisReturnedForTime: /timestamp_to_event request failed with status=%d body=%s", timestampToEventRes.StatusCode, string(timestampToEventResBody))
```
MadLittleMods, line 316: This is just a small fix to make easier to debug when the endpoint fails.

We now also throw when the endpoint throws a 404 but we were expecting an event.

tests/msc3030_test.go
MadLittleMods: 👍 Copied from https://github.com/matrix-org/complement/pull/225/files

Added the Dendrite blacklist tags:
https://github.com/matrix-org/complement/blob/9b2ba1a3f8286b8f7e215b95ce8a5e5315c38d42/tests/room_timestamp_to_event_test.go#L1-L2

tests/room_timestamp_to_event_test.go
```diff
@@ -1,9 +1,9 @@
...
-// +build msc3030
+//go:build !dendrite_blacklist
+// +build !dendrite_blacklist
 
-// This file contains tests for a jump to date API endpoint,
-// currently experimental feature defined by MSC3030, which you can read here:
-// https://github.com/matrix-org/matrix-doc/pull/3030
+// This file contains tests for the `/timestamp_to_event` client and federation API
+// endpoints (also known as *jump to date*). As defined by MSC3030, which you can read
+// here: https://github.com/matrix-org/matrix-doc/pull/3030
```
MadLittleMods, line 7: Where can `fetchUntilMessagesResponseHas` live to be shared with MSC2716 and `tests/room_timestamp_to_event_test.go` (MSC3030)?

The `msc2716` build tag isn't built for the Synapse workers environment (see https://github.com/matrix-org/synapse/pull/14074) so `fetchUntilMessagesResponseHas` isn't available for `tests/room_timestamp_to_event_test.go` to use

MadLittleMods: I guess we can move `fetchUntilMessagesResponseHas` to this file instead since it will always be available for `msc2716` to pull from ⏩

tests/msc2716_test.go
```diff
@@ -1164,42 +1164,6 @@ func includes(needle string, haystack []string) bool {
 	return false
 }
 
-func fetchUntilMessagesResponseHas(t *testing.T, c *client.CSAPI, roomID string, check func(gjson.Result) bool) {
```
MadLittleMods, line 1167: Moving this to `tests/room_timestamp_to_event_test.go` because we can't reference it from the `msc2716` build tag when it's not included.

We don't build with `msc2716` in the workers Complement environment (related https://github.com/matrix-org/synapse/pull/14074)

tests/room_timestamp_to_event_test.go
MadLittleMods: Conversation continued in https://github.com/matrix-org/complement/pull/559#discussion_r1024716899

tests/msc2716_test.go
```diff
@@ -1126,16 +1126,6 @@ func TestImportHistoricalMessages(t *testing.T) {
 	})
```
MadLittleMods, line 1126: Looks like the Dendrite CI job is [failing on `main`](https://github.com/matrix-org/complement/actions/runs/3482537405/jobs/5824967391) as well

tests/room_timestamp_to_event_test.go
```diff
@@ -300,19 +310,20 @@ func mustCheckEventisReturnedForTime(t *testing.T, c *client.CSAPI, roomID strin
...
 		"ts":  []string{timestampString},
 		"dir": []string{direction},
 	}))
 	timestampToEventResBody := client.ParseJSON(t, timestampToEventRes)
 
-	// Only allow a 200 response meaning we found an event or a 404 meaning we didn't.
-	// Other status codes will throw and assumed to be application errors.
+	// Only allow a 200 response meaning we found an event or when no `expectedEventId`, a
+	// 404 meaning we didn't find anything. Other status codes will throw and assumed to
+	// be application errors.
```
squahtx, line 321:
```suggestion
	// Only allow a 200 response meaning we found an event or when no `expectedEventId` is provided, a
	// 404 meaning we didn't find anything. Other status codes will throw and are assumed to
	// be application errors.
```

squahtx (approved):
Looks good.

PR #560, opened 2022-11-21
Homerunner: Permit running HS with remote networking, and provide docker build

michaelkaye:
The aim of this PR is to allow homerunner to start servers using docker-dind, and make them available over a non-local network. We can take advantage of this in a k8s environment by running servers on the PodIP, and so make them available outside of the pod.

This is wanted for  trafficlight, where a adapters wrap and automate real clients (eg, ios/android/web) and so running many of them can be heavyweight. Currently we can run everything locally on a single desktop, but managing simultaneous runs of multiple tests needs us to spread the load across hosts.

The change should be a noop for complement run tests, retaining the default behaviour there, and without a configuration option being set, is also a noop for homerunner users.

If we wanted, this could also allow running a homerunner as a hosted service outside of k8s, with the resultant servers available over the network.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,6 @@ complement
 .idea
 .bin
 .DS_Store
-homerunner
+# Homerunner can end up in various places
+/homerunner
+/cmd/homerunner/homerunner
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -3,4 +3,6 @@ complement
 .idea
 .bin
 .DS_Store
-homerunner
```
michaelkaye, line 6: (required as we were ignoring changes to the homerunner directory)

kegsay (approved):

PR #562, opened 2022-11-28
JoinRoom: send with empty JSON body for spec compliance

squahtx:
When joining a room, it is erroneous to send an empty POST body.
Send an empty JSON body instead.

---

Similar to #460.


squahtx:
The failing Dendrite tests are `TestEventRelationships` and `TestFederatedEventRelationships`, both of which also fail on main. They're unlikely to have anything to do with this PR.

PR #563, opened 2022-11-28
Fix setup for `TestMembersLocal` tests and split into sync types

squahtx:
Depending on timings, the "Existing members see new members' presence"
test in `TestMembersLocal` may see the expected join and presence in an
initial or incremental sync. Both types of sync elicit different
behaviour in homeserver implementations, such as Synapse. Split the test
into initial and incremental sync variants to eliminate the race
condition.

In addition, fix the initial sync case by explicitly setting Bob's
presence to online. The spec does not guarantee that Bob will have a
non-offline presence after merely joining a room, nor does the spec
guarantee that offline presences will show up in initial syncs.

Fixes https://github.com/matrix-org/synapse/issues/13199.

### DIFF
--- a/tests/csapi/rooms_members_local_test.go
+++ b/tests/csapi/rooms_members_local_test.go
@@ -24,6 +24,7 @@ func TestMembersLocal(t *testing.T) {
 	)
 
 	_, incrementalSyncToken := alice.MustSync(t, client.SyncReq{})
+	bob.JoinRoom(t, roomID, []string{})
 
 	t.Run("Parallel", func(t *testing.T) {
 		// sytest: New room members see their own join event
@@ -36,7 +37,6 @@ func TestMembersLocal(t *testing.T) {
 		// sytest: Existing members see new members' join events
 		t.Run("Existing members see new members' join events", func(t *testing.T) {
 			t.Parallel()
-			bob.JoinRoom(t, roomID, []string{})
 			// SyncJoinedTo already checks everything we need to know
 			alice.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(bob.UserID, roomID))
 		})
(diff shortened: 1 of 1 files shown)


DMRobertson (approved):
Many thanks for digging into this Sean, LGTM

squahtx:
I had to move Bob's join out of the test case, since it was getting queued after the two test cases that relied upon it.

DMRobertson:
> I had to move Bob's join out of the test case, since it was getting queued after the two test cases that relied upon it.

To check: was that just an oversight in porting the test to complement?

DMRobertson (approved):

squahtx:
> > I had to move Bob's join out of the test case, since it was getting queued after the two test cases that relied upon it.
> 
> To check: was that just an oversight in porting the test to complement?

Looks like it. In the sytest, Bob joins the room as part of the setup.
https://github.com/matrix-org/sytest/blob/225593e864be317c6efd2823a03a43f56294eebf/tests/30rooms/02members-local.pl#L21-L28

When I split the last test case into two, it changed the pattern in which the tests run and made it a problem.

squahtx:
Thanks for all the fast reviews today!

squahtx:
The tests in this PR aren't run on Dendrite, and the Dendrite CI failures seen are also happening on main.

ShadowJonathan:
This was already being addressed in https://github.com/matrix-org/complement/pull/516, I would appreciate it if that PR was given attention instead of concurrent changes being merged at the same time.

ShadowJonathan:
This PR is also faulty, it uses `BlueprintOneToOneRoom`, which was already determined in #516 to provide faulty results, as Bob and Alice would already see eachother's presence changes through that pre-existing one-on-one room. This was also already determined in #516, so I will be reverting this change in that PR.

PR #564, opened 2022-11-30
Check that message sent during fast join triggers a fed request

MatMaul:
This PR adds some code to check that a message sent during a room in partial state correctly triggers a federation request to be received by the other server.


squahtx (approved):
lgtm!

PR #567, opened 2022-12-02
Add test for joining a room already in partial state

MatMaul:
Also add test for changing display name in a partial state room, since
it looks very similar to a join event.

---

This is currently not working on Synapse.

Associated issue is https://github.com/matrix-org/synapse/issues/12801.
Tests https://github.com/matrix-org/synapse/pull/14606.


tests/federation_room_join_partial_state_test.go
```diff
@@ -3318,6 +3318,50 @@ func TestPartialStateJoin(t *testing.T) {
...
+
+		alice.Client.Timeout = 2 * time.Second
+		paths := []string{"_matrix", "client", "v3", "rooms", serverRoom.RoomID, "send", "m.room.message", "0"}
+		res := alice.MustDoFunc(t, "PUT", paths, client.WithJSONBody(t, map[string]interface{}{
+			"msgtype": "m.text",
+			"body":    "Hello world!",
+		}))
+		body := gjson.ParseBytes(client.ParseJSON(t, res))
+		eventID := body.Get("event_id").Str
+		t.Logf("Alice sent event event ID %s", eventID)
```
squahtx, line 3346: I think we might be able to remove this?

DMRobertson: Agreed---we don't use the eventID anywhere else; Alice waits to see Bob's join.

DMRobertson, line 3346:
```suggestion
```

PR #570, opened 2022-12-12
Allow the homeserver to finish partial-state joining before teardown

squahtx:
Since the partial state join tests reuse the same homeserver deployment,
it is important that each test leaves the homeserver in an acceptable
state for subsequent tests. In particular, we must allow in-progress
partial-state joins to complete before tearing down Complement servers,
otherwise the homeserver may mark Complement hostname:port combinations
as offline and refuse to contact them in subsequent tests.

---

These issues have 502 Bad Gateway failures:
Closes https://github.com/matrix-org/synapse/issues/13564.
Closes https://github.com/matrix-org/synapse/issues/14245.
Closes https://github.com/matrix-org/synapse/issues/14543.

### DIFF
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -115,23 +115,6 @@ func TestPartialStateJoin(t *testing.T) {
 		return syncToken
 	}
 
-	// awaitPartialStateJoinCompletion waits until the joined room is no longer partial-stated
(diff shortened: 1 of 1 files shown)


DMRobertson (approved):
I think this means we now automatically call `awaitPartialStateJoinCompletion` via `defer`. We were already doing so in some places. Is there any danger in doing so again? (I think not: we just recheck that Synapse has fully joined by calling /members?)

PR #574, opened 2022-12-20
Add tests for MSC3391

anoadragon453:
This PR adds Complement tests for MSC3391. The MSC is [currently being implemented in Synapse](https://github.com/matrix-org/synapse/pull/14714).

### DIFF
--- a/tests/msc3391_test.go
+++ b/tests/msc3391_test.go
@@ -36,35 +36,23 @@ func TestRemovingAccountData(t *testing.T) {
 
 	// Test deleting global account data.
 	t.Run("Deleting a user's account data via DELETE works", func(t *testing.T) {
-		createAndDeleteUserAccountData(t, alice, true)
+		createUserAccountData(t, alice)
+		deleteUserAccountData(t, alice, true)
 	})
 	t.Run("Deleting a user's account data via PUT works", func(t *testing.T) {
-		createAndDeleteUserAccountData(t, alice, false)
+		createUserAccountData(t, alice)
+		deleteUserAccountData(t, alice, false)
 	})
 
 	// Test deleting room account data.
 	t.Run("Deleting a user's room data via DELETE works", func(t *testing.T) {
-		createAndDeleteRoomAccountData(t, alice, true, roomID)
+		createRoomAccountData(t, alice, roomID)
+		deleteRoomAccountData(t, alice, true, roomID)
 	})
 	t.Run("Deleting a user's room account data via PUT works", func(t *testing.T) {
-		createAndDeleteRoomAccountData(t, alice, false, roomID)
+		createRoomAccountData(t, alice, roomID)
+		deleteRoomAccountData(t, alice, false, roomID)
 	})
-}
-
-func createAndDeleteUserAccountData(t *testing.T, c *client.CSAPI, viaDelete bool) {
-	// Create the account data and check that it has been created successfully
-	createUserAccountData(t, c)
-
-	// Delete the account data and check that it was deleted successfully
-	deleteUserAccountData(t, c, viaDelete)
-}
-
-func createAndDeleteRoomAccountData(t *testing.T, c *client.CSAPI, viaDelete bool, roomID string) {
-	// Create the account data and check that it has been created successfully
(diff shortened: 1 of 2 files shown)


tests/msc3391_test.go
```diff
@@ -0,0 +1,274 @@
...
+	roomID := alice.CreateRoom(t, map[string]interface{}{})
+
+	for _, httpMethod := range []string{"PUT", "DELETE"} {
+		t.Run(fmt.Sprintf("Deleting a user's account data via %s works", httpMethod), func(t *testing.T) {
+			createAndDeleteAccountData(t, alice, httpMethod == "DELETE", nil)
+		})
+		t.Run("Deleting a user's room account data via ", func(t *testing.T) {
+			createAndDeleteAccountData(t, alice, httpMethod == "DELETE", &roomID)
+		})
+	}
```
clokep, line 44: I don't think this loop is really buying us anything TBH.

```suggestion
	// Test deleting global account data.
	t.Run("Deleting a user's account data via DELETE works", func(t *testing.T) {
		createAndDeleteAccountData(t, alice, true, nil)
	})
	t.Run("Deleting a user's account data via PUT works", func(t *testing.T) {
		createAndDeleteAccountData(t, alice, false, nil)
	})

	// Test deleting room account data.
	t.Run("Deleting a user's room data via DELETE works", func(t *testing.T) {
		createAndDeleteAccountData(t, alice, true, &roomID)
	})
	t.Run("Deleting a user's room account data via PUT works", func(t *testing.T) {
		createAndDeleteAccountData(t, alice, false, &roomID)
	})
```

tests/msc3391_test.go
```diff
@@ -0,0 +1,274 @@
...
+	// Create the account data and check that it has been created successfully
+	createAccountData(t, c, roomID)
+
+	// Delete the account data and check that it was deleted successfully
+	deleteAccountData(t, c, viaDelete, roomID)
+}
+
+// createAccountData creates some account data for a user or a room, and checks that it was
+// created successfully by both querying the data afterwards, and ensuring it appears down /sync.
+func createAccountData(t *testing.T, c *client.CSAPI, roomID *string) {
```
clokep, line 57: Could this be separate into two functions? It looks like the only duplicated code would be the initial sync?

If you really want to keep a parameter to `createAndDeleteAccountData` for the room ID, I think the if statement could go in there for clarity.

tests/msc3391_test.go
```diff
@@ -0,0 +1,274 @@
...
+// deleteAccountData removes account data for a user or room.
+//
+// If viaDelete is true, a request is made to the DELETE endpoint for user or
+// room account data. Otherwise, the PUT method is used with an empty content
+// dictionary instead. MSC3391 specifies that a PUT with an empty content body
+// is functionally equivalent to deleting an account data type directly.
+//
+// If roomID is not nil, room account data for the given room ID will be removed.
+// Otherwise, account data from the user will be removed instead.
+func deleteAccountData(t *testing.T, c *client.CSAPI, viaDelete bool, roomID *string) {
```
clokep, line 150: Similar question here of if we can separate the global / room functionality.

anoadragon453: Good point. I've separated out the various test methods into user and room variants in 58913da.

tests/msc3391_test.go
```diff
@@ -0,0 +1,306 @@
...
+	deleteUserAccountData(t, c, viaDelete)
+}
+
+func createAndDeleteRoomAccountData(t *testing.T, c *client.CSAPI, viaDelete bool, roomID string) {
+	// Create the account data and check that it has been created successfully
+	createRoomAccountData(t, c, roomID)
+
+	// Delete the account data and check that it was deleted successfully
+	deleteRoomAccountData(t, c, viaDelete, roomID)
+}
```
clokep, line 68: I'm not sure these two helpers are that useful, but I won't block on this. 👍

clokep (approved):

tests/msc3391_test.go
anoadragon453: Yeah, they really aren't :)

PR #575, opened 2022-12-21
account data: add test for setting and getting

sumnerevans:
This PR adds tests for getting and setting account data.

The tests correspond to the following sytest tests:

* Can add account data
* Can get account data without syncing
* Can add account data to room
* Can get room account data without syncing

### DIFF
--- a/tests/csapi/account_data_test.go
+++ b/tests/csapi/account_data_test.go
@@ -1,0 +1,67 @@
+package csapi_tests
+
+import (
+	"testing"
+
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/match"
+	"github.com/matrix-org/complement/internal/must"
+)
+
+func TestAddAccountData(t *testing.T) {
+	deployment := Deploy(t, b.BlueprintOneToOneRoom)
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+
+	// sytest: Can add account data
+	// sytest: Can get account data without syncing
+	t.Run("Can add global account data", func(t *testing.T) {
+		// Set the account data entry
+		alice.SetGlobalAccountData(t, "test.key", map[string]interface{}{"value": "first"})
+
+		// check that getting the account data returns the correct value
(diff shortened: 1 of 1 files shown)


sumnerevans:
looks like possibly some overlap with #574 with regards to the internal client functions

anoadragon453:
> https://github.com/matrix-org/complement/pull/574

Heads up that https://github.com/matrix-org/complement/pull/574 has now merged.

tests/csapi/account_data_test.go
```diff
@@ -0,0 +1,83 @@
...
+	deployment := Deploy(t, b.BlueprintOneToOneRoom)
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+
+	// sytest: Can add account data
+	// sytest: Can get account data without syncing
+	t.Run("Can add global account data", func(t *testing.T) {
+		// Set the account data entry
+		resp := alice.SetGlobalAccountData(t, "com.beeper.test", map[string]interface{}{"value": "first"})
```
anoadragon453, line 21: Speaking with my SCT hat on; it feels a little off to put a company's identifier into Complement for anything other than an unstable prefix. It comes across as a bit of an advertisement?

tests/csapi/account_data_test.go
```diff
@@ -0,0 +1,83 @@
...
+		must.MatchResponse(t, resp, match.HTTPResponse{
+			StatusCode: 200,
+		})
```
anoadragon453, line 24: [`SetGlobalAccountData`](https://github.com/matrix-org/complement/blob/cdad7e9d553ae6169c88a35c7a3d6d37c160397d/internal/client/client.go#L179-L181) makes use of a [`MustDoFunc`](https://github.com/matrix-org/complement/blob/cdad7e9d553ae6169c88a35c7a3d6d37c160397d/internal/client/client.go#L474-L484) call, which will already fail the test if the status code is not 2xx.

Likewise with the `GetGlobalAccountData` calls, and the get/set room variants below.

anoadragon453 (changes requested):

squahtx (approved):
Looks good!

PR #576, opened 2023-01-03
Add tests for MSC3890: Remotely silence local notifications

anoadragon453:
Implements tests for MSC3890. The server-side of this MSC is quite small: simply delete a bit of account data when a user's device is logged out/deleted.

~~I ran into some complications when writing this test. Complement doesn't make it easy to manipulate multiple devices of the same user at once (at least as far as I can tell). I created a helper method to log in to an existing user account in order to create another device. From there I did a bit of a hack to create another CSAPI client. I'd be interested if people know a better way to manipulate multiple devices of a single user in Complement, or have ideas on how to improve the status quo.~~ [cd70eb9](https://github.com/matrix-org/complement/pull/576/commits/cd70eb98d0ba71e3666a66358c1412a7c1ddb8d5) improves this substantially by adding a `LoginUser` method to `deployment`. See [this test code](https://github.com/matrix-org/complement/pull/576/files#diff-8c7b7eed95307d27392994ab0e3d05df8ce00e98aeaf53035f75f3c36d0861daR31) for how this can be used to create new devices on an existing user.


tests/msc3890_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Log in to another device on this user
+	_, secondDeviceAccessToken, secondDeviceID := alice.LoginUser(t, alicePassword)
+	secondDevice := &client.CSAPI{
+		AccessToken:      secondDeviceAccessToken,
+		BaseURL:          deployment.HS["hs1"].BaseURL,
+		Client:           client.NewLoggedClient(t, "hs1", nil),
+		SyncUntilTimeout: 5 * time.Second,
+		Debug:            alice.Debug,
+	}
```
squahtx, line 38: I'd be in favour of factoring this out into a method on `Deployment`, analogous to `RegisterUser`.
It's somewhat unexpected to use an already logged in client to log in again.

tests/msc3890_test.go
```diff
@@ -0,0 +1,87 @@
...
+		)
+
+		// Log out the second device
+		secondDevice.MustDoFunc(t, "POST", []string{"_matrix", "client", "v3", "logout"})
+
+		// Using the first device, check that the local notification setting account data for the deleted device was removed.
+		res := alice.DoFunc(t, "GET", []string{"_matrix", "client", "v3", "user", alice.UserID, "account_data", accountDataType})
+		must.MatchResponse(t, res, match.HTTPResponse{
+			StatusCode: 404,
+		})
```
squahtx, line 85: Could we also test this endpoint just before logging out the second device?
Just so we are 100% sure the test is testing the right thing and it's not 404ing for some other reason.

tests/msc3890_test.go
anoadragon453: Hmm, oddly I tried adding a `LoginUser` method initially in this PR and for whatever reason I was finding it difficult. But doing so again today in [cd70eb9](https://github.com/matrix-org/complement/pull/576/commits/cd70eb98d0ba71e3666a66358c1412a7c1ddb8d5) and it was a breeze and the test code is much cleaner!

So thanks for the suggestion :)

internal/docker/deployment.go
```diff
@@ -118,6 +118,32 @@ func (d *Deployment) RegisterUser(t *testing.T, hsName, localpart, password stri
...
+		BaseURL:          dep.BaseURL,
+		Client:           client.NewLoggedClient(t, hsName, nil),
+		SyncUntilTimeout: 5 * time.Second,
+		Debug:            d.Deployer.debugLogging,
+	}
+	dep.CSAPIClients = append(dep.CSAPIClients, client)
+	userID, accessToken, deviceID := client.LoginUser(t, localpart, password)
+
+	// remember the token so subsequent calls to deployment.Client return the user
+	dep.AccessTokens[userID] = accessToken
```
squahtx, line 139: I wonder if we should leave the original access token in place. I can't convince myself that we should do one or the other, so I'm happy to leave the code as-is.

flake.lock
```diff
@@ -0,0 +1,40 @@
+{
```
squahtx, line 1: I didn't notice this before, but could we move it to a separate PR or confirm that the dendrite/complement team are happy to maintain it?

internal/docker/deployment.go
anoadragon453: Hmm, I think this might be a bit magic actually. It makes more sense to me if the original device is always referred to by `deployment.Client`, and you can create devices "on the side" if you like.

I'm partial to removing this assignment, and just calling this detail out in the method's docstring.

flake.lock
anoadragon453: Oh lol, that shouldn't be there. I do want to PR some nix stuff soon to make my life easier, but not in this PR certainly.

squahtx (approved):
lgtm!

PR #577, opened 2023-01-06
Add `GetAllPushRules`, `GetPushRule` and `SetPushRule` client helper methods

anoadragon453:
Some helper methods to get and set push rules. We also refactor [tests/csapi/push_test.go](https://github.com/matrix-org/complement/pull/577/files#diff-538b85fd27b45b086bd4368eac7da9042a0d290e62de779779792572203037f0) to use them.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -235,8 +235,9 @@ func (c *CSAPI) GetPushRule(t *testing.T, scope string, kind string, ruleID stri
 }
 
 // SetPushRule creates a new push rule on the user, or modifies an existing one.
-// `before` and `after` parameters can be provided, which if set will map to the
-// `before` and `after` query parameters on the set push rules endpoint:
+// If `before` or `after` parameters are not set to an empty string, their values
+// will be set as the `before` and `after` query parameters respectively on the
+// "set push rules" client endpoint:
 // https://spec.matrix.org/v1.5/client-server-api/#get_matrixclientv3pushrules.
 //
 // Example of setting a push rule with ID 'com.example.rule2' that must come after 'com.example.rule1':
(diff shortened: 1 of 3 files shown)


internal/client/client.go
```diff
@@ -188,6 +188,77 @@ func (c *CSAPI) SetRoomAccountData(t *testing.T, roomID string, eventType string
...
+	t.Helper()
+
+	// If the `before` or `after` arguments have been provided, construct same-named query parameters
+	queryParams := url.Values{}
+	if before != nil {
+		queryParams.Add("before", *before)
+	}
+	if after != nil {
+		queryParams.Add("after", *after)
+	}
```
DMRobertson, line 257: Was slightly surprised that `before` and `after` are passed as pointers. I guess the point is to allow to construct requests of the form `...?before=`? (one param provided as empty string, another omitted).

(I note that the spec doesn't mandate that `rule_id`s should be nonempty strings, so I suppose `before` and `after` could be take to be possibly empty too.)

DMRobertson (approved):

internal/client/client.go
anoadragon453: Hmm, I did this as I figured passing `nil` to mean "omit the parameter" would be more natural than passing `""` to mean the same thing. But I suppose it does allow callers to pass `""` and thus send up an empty string value.

Which do you think is more intuitive? Whichever way we choose, I should probably include a clarifying line in the docstring.

anoadragon453: Looking around `client.go`, I don't see pointers being used as optional parameters anywhere else. So I think I'll revert to non-pointers for consistency (and to avoid people accidentally passing `""` when they meant `nil` :).

DMRobertson: Don't have any strong opinions (unusually); I'd try to be consistent with whatever complement already does and whatever's considered idomatic go.

anoadragon453: Sounds good! I'm going to go ahead with an empty string then - it's still relatively easy to use and prevents a potential `""` footgun.

PR #579, opened 2023-01-09
Fix panics due to closed channel in some tests

erikjohnston:
A number of tests start goroutines that write into channels when completed, however the tests would close the channel if it exited early (e.g. due to a timeout).

Instead, make the goroutine that writes into the channel the one that closes it.

This does mean that the goroutines will continue past the end of the test (?), but they did that before too. I am wondering if there is a better way of doing these things...

### DIFF
--- a/tests/csapi/push_test.go
+++ b/tests/csapi/push_test.go
@@ -111,6 +111,9 @@ func checkWokenUp(t *testing.T, csapi *client.CSAPI, syncReq client.SyncReq, fn 
 	errChan := make(chan error, 1)
 	syncStarted := make(chan struct{})
 	go func() {
+		defer close(errChan)
+		defer close(syncStarted)
+
 		var syncResp gjson.Result
 		syncStarted <- struct{}{}
 		syncResp, nextBatch = csapi.MustSync(t, syncReq)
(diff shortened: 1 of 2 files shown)


anoadragon453 (approved):
So the problem is that `Fatalf` or `Errorf` etc. will run and the test exits early and all `defer`s (which will close the channels that the goroutines write to) will close. Then execution of the goroutine will continue (as a `panic` only fails up the stack) and when it tries to write to the channel it will then panic itself and fill the logs with noise.

I suppose the *right* thing to do is to have the goroutine check a "signal channel" after it completes long-running operations (such as `/sync`) to make sure that it should continue execution before writing to channels. Something like the following:

```go
errChan := make(chan error, 1)
quitChan := make(chan bool, 1)
defer close(quitChan)
defer close(errChan)
defer quitChan <- true
go func() {
	syncResp, nextBatch := csapi.MustSync(t, syncReq)
	select {
	case <- quitChan:
	    // We have been told to quit
		return
	default:
		// process the request and return
	    if ... {
	        errChan <- fmt.Errorf("...")
	    }
	    errChan <- nil
	}
}
```

What you propose technically works as well though.

erikjohnston:
> I suppose the _right_ thing to do is to have the goroutine check a "signal channel" after it completes long-running operations (such as `/sync`) to make sure that it should continue execution before writing to channels.

I suppose that might make sense for the one goroutine that is multiple statements, but I don't see the benefit for the others (like the one in the example)?

erikjohnston:
(merging now to unblock fixing flakey tests)

PR #581, opened 2023-01-10
Baby steps for testing leaves during a faster join

DMRobertson:
part of matrix-org/synapse#12802

- Check that leaves work correctly today, even if they block on resync
- Dummy test that leaves will work before resync completes
- Ideas for future tests

### DIFF
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -3181,12 +3181,17 @@ func TestPartialStateJoin(t *testing.T) {
 		must.MatchResponse(t, response, spec)
 	})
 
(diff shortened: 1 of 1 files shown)


squahtx (approved):
lgtm!

Thoroughly logging what the test is doing in human-readable terms is a nice touch.
We've been lax at it when writing the previous tests and as a result it's been painful to understand where they're failing when they go wrong.

DMRobertson:
> Thoroughly logging what the test is doing in human-readable terms is a nice touch.
> We've been lax at it when writing the previous tests and as a result it's been painful to understand where they're failing when they go wrong.

Glad you like it! (I was writing these as `// comments` anyway and was finding it painful to keep track when debugging.)

PR #585, opened 2023-01-18
Add test to determine that the public room list is correctly updated after lazy load finishes syncing state

H-Shay:
As the title states.

### DIFF
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -3295,6 +3295,61 @@ func TestPartialStateJoin(t *testing.T) {
 		}
 	})
 
+	t.Run("Room stats are correctly updated once state re-sync completes", func(t *testing.T) {
+		deployment := Deploy(t, b.BlueprintAlice)
+		defer deployment.Destroy(t)
+
+		// create a user with admin powers as we will need this power to make the remote room visible in the
+		// local room list
+		terry := deployment.RegisterUser(t, "hs1", "terry", "pass", true)
+
+		server := createTestServer(t, deployment)
+		cancel := server.Listen()
+		defer cancel()
+		serverRoom := createTestRoom(t, server, terry.GetDefaultRoomVersion(t))
+
+		// start a partial state join
+		psjResult := beginPartialStateJoin(t, server, serverRoom, terry)
+		defer psjResult.Destroy(t)
+
+		// make the remote room visible in the local room list
+		reqBody := client.WithJSONBody(t, map[string]interface{}{
+			"visibility": "public",
+		})
+		terry.MustDoFunc(t, "PUT", []string{"_matrix", "client", "v3", "directory", "list", "room", serverRoom.RoomID}, reqBody)
+
+		// sanity check - before the state has completed syncing state we would expect only one user
+		// to show up in the room list
+		res := terry.MustDoFunc(t, "GET", []string{"_matrix", "client", "v3", "publicRooms"})
+		body, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			t.Fatalf("something broke: %v", err)
+		}
+		numJoinedMembers := gjson.GetBytes(body, "chunk.0.num_joined_members")
+		if numJoinedMembers.Int() != 1 {
+			t.Fatalf("Expected 1 users, returned %s", numJoinedMembers)
+		}
+
+		// finish syncing the state
+		psjResult.FinishStateRequest()
+		terry.MustSyncUntil(t,
+			client.SyncReq{},
+			client.SyncJoinedTo(terry.UserID, psjResult.ServerRoom.RoomID),
+		)
+
+		// check the number of joined users, it should now be 3
+		res2 := terry.MustDoFunc(t, "GET", []string{"_matrix", "client", "v3", "publicRooms"})
+		body2, err2 := ioutil.ReadAll(res2.Body)
+		if err2 != nil {
+			t.Fatalf("something broke: %v", err2)
+		}
+		numJoinedMembers2 := gjson.GetBytes(body2, "chunk.0.num_joined_members")
+		if numJoinedMembers2.Int() != 3 {
+			t.Fatalf("Expected 3 users, returned %s", numJoinedMembers2)
+		}
+
+	})
+
 	// TODO: tests which assert that:
 	//   - Join+Join+Leave+Leave works
 	//   - Join+Leave+Join works
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -3344,7 +3344,7 @@ func TestPartialStateJoin(t *testing.T) {
 				}
 				return false
 			}))
-	})
+		})
 
 	// TODO: tests which assert that:
 	//   - Join+Join+Leave+Leave works
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -3329,9 +3329,9 @@ func TestPartialStateJoin(t *testing.T) {
 		psjResult.FinishStateRequest()
 		awaitPartialStateJoinCompletion(t, psjResult.ServerRoom, terry)
 
-		// the rooms stats are updated by a background job which is not guaranteed to have completed by the time
-		// the state sync has completed. We check for up to 3 seconds that the job has completed. The number of joined
-		// users should now be 3: one local user (terry) and two remote (charlie and derek)
+		// In Synapse rooms stats are updated by a background job which is not guaranteed to have completed by the time
+		// the state sync has completed. To account for that, we check for up to 3 seconds that the job has completed.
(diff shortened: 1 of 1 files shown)


tests/federation_room_join_partial_state_test.go
```diff
@@ -3295,6 +3295,61 @@ func TestPartialStateJoin(t *testing.T) {
...
+		defer psjResult.Destroy(t)
+
+		// make the remote room visible in the local room list
+		reqBody := client.WithJSONBody(t, map[string]interface{}{
+			"visibility": "public",
+		})
+		terry.MustDoFunc(t, "PUT", []string{"_matrix", "client", "v3", "directory", "list", "room", serverRoom.RoomID}, reqBody)
+
+		// sanity check - before the state has completed syncing state we would expect only one user
+		// to show up in the room list
```
DMRobertson, line 3322: Presumably terry?

I'm a little surprised by this---what's the logic/thinking here? Are we only tracking counting local users until the resync completes?

[The room is created by a user called `charlie` on the complement-controlled HS.](https://github.com/matrix-org/complement/blob/fe5887f0907b88a9c562a67009200b8b7af26545/tests/federation_room_join_partial_state_test.go#L94-L103) I would expect Charlie's membership to show up in the /send_join response: Charlie's membership should be an auth_event of the power levels event that Charlie creates, and the PL event is an auth_event of Terry's join. Therefore I'd expect Terry to be able to see Charlie (but not Derek).

tests/federation_room_join_partial_state_test.go
```diff
@@ -3295,6 +3295,61 @@ func TestPartialStateJoin(t *testing.T) {
...
+		res := terry.MustDoFunc(t, "GET", []string{"_matrix", "client", "v3", "publicRooms"})
+		body, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			t.Fatalf("something broke: %v", err)
+		}
+		numJoinedMembers := gjson.GetBytes(body, "chunk.0.num_joined_members")
+		if numJoinedMembers.Int() != 1 {
+			t.Fatalf("Expected 1 users, returned %s", numJoinedMembers)
+		}
```
DMRobertson, line 3331: I think Kegan's intention for complement is to make more use of the `must` and `match` libraries. You might be able to use `must.MatchResponse` here to make it neater, as in e.g.

https://github.com/matrix-org/complement/blob/ec6e09c17eddb6c0b4686c76c3a2f7272cb708f2/tests/csapi/room_members_test.go#L175-L187

something like 

```go
must.matchResponse(t, res, match.HTTPResponse{
    StatusCode: 200,
    JSON: []match.JSON{
        match.JSONKeyEqual("chunk.0.num_joined_members", 1)
    }
})

I don't know if you find that clearer. It's a little more declarative and delegates the error handling I guess. :shrug:

File: tests/federation_room_join_partial_state_test.go [hunk 3]
```diff
@@ -3295,6 +3295,61 @@ func TestPartialStateJoin(t *testing.T) {
...
+
+		// finish syncing the state
+		psjResult.FinishStateRequest()
+		terry.MustSyncUntil(t,
+			client.SyncReq{},
+			client.SyncJoinedTo(terry.UserID, psjResult.ServerRoom.RoomID),
+		)
+
+		// check the number of joined users, it should now be 3
```
DMRobertson, line 3340: Presumably Terry, Derek and Charlie? I'd suggest spelling this out. E.g. "Now that the resync has completed, we see remote members, not just local ones"

DMRobertson (commented):
...I've only just now seen that this is a draft. Oh well. I hope the thoughts are useful anyway

tests/federation_room_join_partial_state_test.go
squahtx: While Charlie's membership does show up in the /send_join response, it's in the auth chain only and not the current room state. Room stats gets computed off the current state only.

tests/federation_room_join_partial_state_test.go
```diff
@@ -3295,6 +3295,61 @@ func TestPartialStateJoin(t *testing.T) {
...
+		res2 := terry.MustDoFunc(t, "GET", []string{"_matrix", "client", "v3", "publicRooms"})
+		body2, err2 := ioutil.ReadAll(res2.Body)
+		if err2 != nil {
+			t.Fatalf("something broke: %v", err2)
+		}
+		numJoinedMembers2 := gjson.GetBytes(body2, "chunk.0.num_joined_members")
+		if numJoinedMembers2.Int() != 3 {
+			t.Fatalf("Expected 3 users, returned %s", numJoinedMembers2)
+		}
```
squahtx, line 3349: I ran this locally and it flakes randomly. I think it's because the stats updater is its own background loop in synapse and isn't guaranteed to be done when the partial state join completes.
For things like this, I think we do a loop with a timeout in other parts of complement (which I can't find examples of at the moment).

I also ran the test with workers and it passed (which was surprising because I thought synapse was missing a bit to poke the notifier when updating the current state for a partial state room).

tests/federation_room_join_partial_state_test.go
```diff
@@ -3295,6 +3295,61 @@ func TestPartialStateJoin(t *testing.T) {
...
+		if numJoinedMembers.Int() != 1 {
+			t.Fatalf("Expected 1 users, returned %s", numJoinedMembers)
+		}
+
+		// finish syncing the state
+		psjResult.FinishStateRequest()
+		terry.MustSyncUntil(t,
+			client.SyncReq{},
+			client.SyncJoinedTo(terry.UserID, psjResult.ServerRoom.RoomID),
+		)
```
squahtx, line 3338: Since Mat's PR to address non-lazy loading /syncs hasn't landed yet, /sync can return the room join before the room has been fully un-partial stated. Elsewhere we use:
```suggestion
		awaitPartialStateJoinCompletion(t, psjResult.ServerRoom, terry)
```
which blocks on /joined_members.

H-Shay: This should address the random flakes, correct?

tests/federation_room_join_partial_state_test.go
H-Shay: Is it a problem that the test passes with workers (i.e. does that violate any assumptions that we should look deeper into) or is it fine?

tests/federation_room_join_partial_state_test.go
H-Shay: Err actually it probably wouldn't, thinking about it for one second more.

squahtx: Good question. I think you're right in that it's not sufficient to address the flakes.

tests/federation_room_join_partial_state_test.go
squahtx: A little. I'm looking why it passes with workers at the moment.

tests/federation_room_join_partial_state_test.go
```diff
@@ -3321,33 +3318,32 @@ func TestPartialStateJoin(t *testing.T) {
...
-		numJoinedMembers2 := gjson.GetBytes(body2, "chunk.0.num_joined_members")
-		if numJoinedMembers2.Int() != 3 {
-			t.Fatalf("Expected 3 users, returned %s", numJoinedMembers2)
-		}
-
+		awaitPartialStateJoinCompletion(t, psjResult.ServerRoom, terry)
+
+		// the rooms stats are updated by a background job which is not guaranteed to have completed by the time
+		// the state sync has completed. We check for up to 3 seconds that the job has completed. The number of joined
+		// users should now be 3: one local user (terry) and two remote (charlie and derek)
```
squahtx, line 3334: The background job is a synapse implementation detail. I'd either clarify that this refers to synapse or just state that there's no guarantee, without referring to the background job.

tests/federation_room_join_partial_state_test.go
squahtx: Thanks for resolving the flake!

tests/federation_room_join_partial_state_test.go
H-Shay: I added info about the background job being Synapse, hopefully it's clearer.

tests/federation_room_join_partial_state_test.go
```diff
@@ -3295,6 +3295,57 @@ func TestPartialStateJoin(t *testing.T) {
...
+				if err != nil {
+					t.Fatalf("something broke: %v", err)
+				}
+				numJoinedMembers := gjson.GetBytes(body, "chunk.0.num_joined_members")
+				if numJoinedMembers.Int() == 3 {
+					return true
+				}
+				return false
+			}))
+	})
```
squahtx, line 3347: small indentation fix
```suggestion
		})
```

squahtx (approved):
Looks good! Big thank you for getting the tests written so quickly.

H-Shay:
Merged, thank you.

PR #586, opened 2023-01-18
Add a test using a sync token for relations.

clokep:
Tests for matrix-org/synapse#14866.


tests/csapi/room_relations_test.go
```diff
@@ -205,3 +206,91 @@ func TestRelationsPagination(t *testing.T) {
 		},
 	})
 }
+
+func TestRelationsPaginationSync(t *testing.T) {
+  runtime.SkipIf(t, runtime.Dendrite) // FIXME: https://github.com/matrix-org/dendrite/issues/2944
```
DMRobertson, line 211: Indentation looks a bit off here. Might need to run `gofmt`?

tests/csapi/room_relations_test.go
```diff
@@ -205,3 +206,91 @@ func TestRelationsPagination(t *testing.T) {
...
+
+	deployment := Deploy(t, b.BlueprintAlice)
+	defer deployment.Destroy(t)
+
+	alice := deployment.Client(t, "hs1", "@alice:hs1")
+	roomID := alice.CreateRoom(t, map[string]interface{}{"preset": "public_chat"})
+	_, token := alice.MustSync(t, client.SyncReq{})
+
+	res := alice.MustDoFunc(t, "PUT", []string{"_matrix", "client", "v3", "rooms", roomID, "send", "m.room.message", "txn-1"}, client.WithJSONBody(t, map[string]interface{}{
```
DMRobertson, line 220: Nothing wrong with this, but you might prefer to use `SendEventUnsynced` here

https://github.com/matrix-org/complement/blob/70f58acc4576daddaba968a8a29db55b04418c5d/internal/client/client.go#L263-L276

tests/csapi/room_relations_test.go
```diff
@@ -205,3 +206,91 @@ func TestRelationsPagination(t *testing.T) {
...
+			"m.relates_to": map[string]interface{}{
+				"event_id": rootEventID,
+				"rel_type": "m.thread",
+			},
+		}))
+		event_id = client.GetJSONFieldStr(t, client.ParseJSON(t, res), "event_id")
+	}
+
+	// Sync and keep the token.
+	nextBatch := alice.MustSyncUntil(t, client.SyncReq{Since: token}, client.SyncTimelineHas(roomID, func(r gjson.Result) bool {
```
DMRobertson, line 241: There is a `SyncTimelineHasEventID` helper which might save some typing:

https://github.com/matrix-org/complement/blob/70f58acc4576daddaba968a8a29db55b04418c5d/internal/client/client.go#L793-L798

DMRobertson (commented):
I've left some suggestions about helper functions, but take them with a pinch of salt.

Could we have a comment summarising what the test does and what it checks? Suggestion

```
Alice starts a thread and sends 10 messages into it.
Check she can paginate backwards through the thread's history, starting from a sync token midway through the thread.
```

Otherwise LGTM

tests/csapi/room_relations_test.go
```diff
@@ -205,3 +206,91 @@ func TestRelationsPagination(t *testing.T) {
...
+		return r.Get("event_id").Str == event_id
+	}))
+
+	// Create more related events.
+	event_ids := [5]string{}
+	for i := 0; i < 5; i++ {
+		res = alice.MustDoFunc(t, "PUT", []string{"_matrix", "client", "v3", "rooms", roomID, "send", "m.room.message", fmt.Sprintf("txn-%d", 2+i)}, client.WithJSONBody(t, map[string]interface{}{
```
DMRobertson, line 248: Err, actually, is this reusing transaction IDs from the first batch of 5 events?

clokep: Hm, maybe?

DMRobertson: The first message is sent with `txn-1`. Then we send five events in `txn-2`, ..., `txn-6`. AFAICS we then repeat `txn-2`, ... `txn-6` here.

DMRobertson: Perhaps `event_ids[4]` ends up being equal to `event_id` in this case?

clokep: I'll need to double check, good catch.

tests/csapi/room_relations_test.go
clokep: Huh, was that always there?

DMRobertson: No, I split it out of `SendEventSynced` recently, in #588

tests/csapi/room_relations_test.go
clokep: Ah, this showed a but. I was using the wrong parameters in the subsequent sync call too.

clokep:
Thanks for the review @DMRobertson! 🎉 Would you mind taking another look?

tests/csapi/room_relations_test.go
```diff
@@ -205,3 +206,98 @@ func TestRelationsPagination(t *testing.T) {
...
+	}))
+
+	// Create more related events.
+	event_ids := [5]string{}
+	for i := 0; i < 5; i++ {
+		event_ids[i] = alice.SendEventUnsynced(t, roomID, b.Event{
+			Type: "m.room.message",
+			Content: map[string]interface{}{
+				"msgtype": "m.text",
+				"body":    fmt.Sprintf("reply %d", i),
```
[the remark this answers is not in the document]
clokep: Yeah, it didn't really matter. I can tweak them so that they differ.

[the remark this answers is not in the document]
DMRobertson: ahh, that's a nice way to do it :)

DMRobertson (approved):
I'm happy if CI is!

PR #587, opened 2023-01-18
Add a test to verify that the user directory is properly updated after a lazy load join finishes syncing

H-Shay:
As the title states.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -414,7 +414,7 @@ func (c *CSAPI) LoginUser(t *testing.T, localpart, password string) (userID, acc
 			"user": localpart,
 		},
 		"password": password,
-		"type": "m.login.password",
+		"type":     "m.login.password",
 	}
 	res := c.MustDoFunc(t, "POST", []string{"_matrix", "client", "v3", "login"}, WithJSONBody(t, reqBody))
 
@@ -521,7 +521,11 @@ func (c *CSAPI) GetDefaultRoomVersion(t *testing.T) gomatrixserverlib.RoomVersio
 // WithRawBody sets the HTTP request body to `body`
 func WithRawBody(body []byte) RequestOpt {
 	return func(req *http.Request) {
-		req.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+		req.Body = ioutil.NopCloser(bytes.NewReader(body))
+		req.GetBody = func() (io.ReadCloser, error) {
+			r := bytes.NewReader(body)
+			return ioutil.NopCloser(r), nil
+		}
 		// we need to manually set this because we don't set the body
 		// in http.NewRequest due to using functional options, and only in NewRequest
 		// does the stdlib set this for us.
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -3311,17 +3311,22 @@ func TestPartialStateJoin(t *testing.T) {
 		psjResult := beginPartialStateJoin(t, server, serverRoom, rocky)
 		defer psjResult.Destroy(t)
 
-		// sanity check - before the state has completed syncing state we would expect rocky to show up
+		// sanity check - before the homeserver has completed syncing state we would expect rocky to show up
 		reqBody := client.WithJSONBody(t, map[string]interface{}{
 			"search_term": rocky.UserID,
 		})
-		time.Sleep(time.Second * 1)
-		res := rocky.MustDoFunc(t, "POST", []string{"_matrix", "client", "v3", "user_directory", "search"}, reqBody)
-		must.MatchResponse(t, res, match.HTTPResponse{
-			StatusCode: 200,
-			JSON: []match.JSON{
-				match.JSONKeyEqual("results.0.user_id", "@rocky:hs1"),
-			}})
+		rocky.MustDoFunc(t, "POST", []string{"_matrix", "client", "v3", "user_directory", "search"}, reqBody,
+			client.WithRetryUntil(time.Second*3, func(res *http.Response) bool {
+				body, err := ioutil.ReadAll(res.Body)
+				if err != nil {
+					t.Fatalf("something broke: %v", err)
+				}
+				user_id := gjson.GetBytes(body, "results.0.user_id")
+				if user_id.Str == rocky.UserID {
+					return true
+				}
+				return false
+			}))
 
 		// .. but not rod's
 		reqBody2 := client.WithJSONBody(t, map[string]interface{}{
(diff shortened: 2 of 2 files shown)


tests/federation_room_join_partial_state_test.go
```diff
@@ -3295,6 +3295,85 @@ func TestPartialStateJoin(t *testing.T) {
...
+			}})
+
+		// finish syncing the state
+		psjResult.FinishStateRequest()
+		awaitPartialStateJoinCompletion(t, psjResult.ServerRoom, rocky)
+
+		// the rooms stats are updated by a background job in Synapse which is not guaranteed to have completed by the time
+		// the state sync has completed. We check for up to 3 seconds that the job has completed, after which Charlie
+		// the job should have finished and Charlie and Derek should be visible in the user directory
+		time.Sleep(time.Second * 3)
```
H-Shay, line 3339: In the interest of expediency I stuck a sleep here. I attempted on lines 3363-3374 (I commented out the attempt in case it's helpful to see) to use `client.WithRetryUntil` to loop until the condition was met, but I could not figure out how to make that function work with a request body, which is necessary for this API (I think?). It worked for the first iteration but then complained about an empty body for the next iteration, my best guess is that the request body is being consumed on the first attempt? I tried a few different formulations to feed the request body in but failed, any insight here would be appreciated.

H-Shay:
Related: https://github.com/matrix-org/synapse/issues/12815

tests/federation_room_join_partial_state_test.go
```diff
@@ -3295,6 +3295,89 @@ func TestPartialStateJoin(t *testing.T) {
...
+
+		// start a partial state join
+		psjResult := beginPartialStateJoin(t, server, serverRoom, rocky)
+		defer psjResult.Destroy(t)
+
+		// sanity check - before the state has completed syncing state we would expect rocky to show up
+		reqBody := client.WithJSONBody(t, map[string]interface{}{
+			"search_term": rocky.UserID,
+		})
+		res := rocky.MustDoFunc(t, "POST", []string{"_matrix", "client", "v3", "user_directory", "search"}, reqBody)
```
H-Shay, line 3318: This could flake as the user directory might be slow in updating, it might make sense to stick a retry loop (if I can figure that out, see below) here as well.

tests/federation_room_join_partial_state_test.go
```diff
@@ -3295,6 +3295,90 @@ func TestPartialStateJoin(t *testing.T) {
...
+
+		// start a partial state join
+		psjResult := beginPartialStateJoin(t, server, serverRoom, rocky)
+		defer psjResult.Destroy(t)
+
+		// sanity check - before the state has completed syncing state we would expect rocky to show up
+		reqBody := client.WithJSONBody(t, map[string]interface{}{
+			"search_term": rocky.UserID,
+		})
+		time.Sleep(time.Second * 1)
```
H-Shay, line 3318: This could flake as the user directory might be slow in updating, it might make sense to stick a retry loop (if I can figure that out, see below) here as well, I've added a sleep for the time being.

tests/federation_room_join_partial_state_test.go
squahtx: > the request body is being consumed on the first attempt

That looks like what is happening. `req.Body` is set to an `io.Reader` containing the JSON. At the end of the first request, the read position lies at the end of the JSON, and so the retry reads 0 bytes.

There's a `GetBody` field that can be set on `http.Request`. If we modify `WithRawBody`, we can get the retries to work:
```diff
diff --git a/internal/client/client.go b/internal/client/client.go
index 7ffacff..05dc056 100644
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -522,6 +522,10 @@ func (c *CSAPI) GetDefaultRoomVersion(t *testing.T) gomatrixserverlib.RoomVersio
 // WithRawBody sets the HTTP request body to `body`
 func WithRawBody(body []byte) RequestOpt {
        return func(req *http.Request) {
-               req.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+               req.Body = ioutil.NopCloser(bytes.NewReader(body))
+               req.GetBody = func() (io.ReadCloser, error) {
+                       r := bytes.NewReader(body)
+                       return ioutil.NopCloser(r), nil
+               }
                // we need to manually set this because we don't set the body
                // in http.NewRequest due to using functional options, and only in NewRequest
                // does the stdlib set this for us.
```

We replace `NewBuffer` with `NewReader`, because the docs say
> NewBuffer creates and initializes a new Buffer using buf as its initial contents. The new Buffer takes ownership of buf, and the caller should not use buf after this call.

tests/federation_room_join_partial_state_test.go
```diff
@@ -3295,6 +3295,90 @@ func TestPartialStateJoin(t *testing.T) {
...
+				match.JSONKeyEqual("results", [0]string{}),
+			}})
+
+		// finish syncing the state
+		psjResult.FinishStateRequest()
+		awaitPartialStateJoinCompletion(t, psjResult.ServerRoom, rocky)
+
+		// the rooms stats are updated by a background job in Synapse which is not guaranteed to have completed by the time
+		// the state sync has completed. We check for up to 3 seconds that the job has completed, after which Charlie
+		// the job should have finished and Charlie and Derek should be visible in the user directory
```
squahtx, line 3343: This comment has gone a little stale and could use an update.

tests/federation_room_join_partial_state_test.go
```diff
@@ -3295,6 +3295,90 @@ func TestPartialStateJoin(t *testing.T) {
...
+
+		// add some new users to avoid the test being polluted by previous tests
+		serverRoom.AddEvent(createJoinEvent(t, server, serverRoom, server.UserID("rod")))
+		serverRoom.AddEvent(createJoinEvent(t, server, serverRoom, server.UserID("todd")))
+
+		// start a partial state join
+		psjResult := beginPartialStateJoin(t, server, serverRoom, rocky)
+		defer psjResult.Destroy(t)
+
+		// sanity check - before the state has completed syncing state we would expect rocky to show up
```
squahtx, line 3314: small typo?
```suggestion
		// sanity check - before the homeserver has completed syncing state we would expect rocky to show up
```

tests/federation_room_join_partial_state_test.go
```diff
@@ -3295,6 +3295,90 @@ func TestPartialStateJoin(t *testing.T) {
...
+
+		// the rooms stats are updated by a background job in Synapse which is not guaranteed to have completed by the time
+		// the state sync has completed. We check for up to 3 seconds that the job has completed, after which Charlie
+		// the job should have finished and Charlie and Derek should be visible in the user directory
+		time.Sleep(time.Second * 3)
+
+		reqBody3 := client.WithJSONBody(t, map[string]interface{}{
+			"search_term": "rod",
+		})
+		rodFullId := "@rod:" + server.ServerName()
```
squahtx, line 3349: We would usually write this as `server.UserID("rod")`
and the same for todd below.

squahtx:
I forgot to mention: this passed when I ran it locally with workers, modulo the missing `Notifier.notify_replication` call of course.

tests/federation_room_join_partial_state_test.go
H-Shay: I put this change in this PR, let me know if it should be in a separate PR instead.

internal/client/client.go
```diff
@@ -414,7 +414,7 @@ func (c *CSAPI) LoginUser(t *testing.T, localpart, password string) (userID, acc
 			"user": localpart,
 		},
 		"password": password,
-		"type": "m.login.password",
+		"type":     "m.login.password",
```
H-Shay, line 417: My editor insisted on this change, I can edit the file here if it's a problem.

squahtx: Can we revert this change please?

squahtx (approved):
Looks good!

If you're feeling up for it, you could try factoring out the common code from the 3 checks into an `assertUserDirectoryContains` local function. `createJoinEvent` has an example of the syntax.

DMRobertson:
> If you're feeling up for it, you could try factoring out the common code from the 3 checks into an `assertUserDirectoryContains` local function. `createJoinEvent` has an example of the syntax.

In the interests of getting faster joins out there I'm going to merge this as-is; though by all means do feel free to factor this out.

PR #588, opened 2023-01-19
Tests for eager-sync behaviour during and after a partial join

DMRobertson:
Based on #584 by @MatMaul.

Matches https://github.com/matrix-org/synapse/pull/14870

### DIFF
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -1187,6 +1187,8 @@ func TestPartialStateJoin(t *testing.T) {
 		// reply to hs2 with a bogus /state_ids response
 		fedStateIdsSendResponseWaiter.Finish()
 
+		// We expect hs2 to fall back to requesting state from hs1, in order to
+		// complete the partial state join
 		nextBatch = charlie.MustSyncUntil(
 			t,
 			client.SyncReq{Since: nextBatch},
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -255,7 +255,7 @@ func TestPartialStateJoin(t *testing.T) {
 	})
 
 	t.Run("EagerIncrementalSyncDuringPartialStateJoin", func(t *testing.T) {
-		eagerSyncDuringPartialStateJoinTest(t, "initial", true)
+		eagerSyncDuringPartialStateJoinTest(t, "incremental", true)
 	})
 
 	// when Alice does a lazy-loading sync, she should see the room immediately
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -118,9 +118,9 @@ func TestPartialStateJoin(t *testing.T) {
 	deployment := Deploy(t, b.BlueprintAlice)
 	defer deployment.Destroy(t)
 
-	// test that a regular /sync request made during a partial-state /send_join
+	// test that a non lazy /sync request made during a partial-state /send_join
 	// request does not return the room until the state is correctly synced.
-	t.Run("RegularSyncDuringPartialStateJoin", func(t *testing.T) {
+	t.Run("NonLazySyncDuringPartialStateJoin", func(t *testing.T) {
 		alice := deployment.RegisterUser(t, "hs1", "t1alice", "secret", false)
 
 		server := createTestServer(t, deployment)
@@ -132,12 +132,12 @@ func TestPartialStateJoin(t *testing.T) {
 
 		// Alice has now joined the room, and the server is syncing the state in the background.
 
-		// Regular sync shouldn't include the room yet
+		// sync shouldn't include the room yet
 		response, nextBatch := alice.MustSync(t, client.SyncReq{})
 
 		syncJoinedRoomPath := "rooms.join." + client.GjsonEscape(serverRoom.RoomID)
 		if response.Get(syncJoinedRoomPath).Exists() {
-			t.Fatal("Regular sync shouldn't include the joined room until resync is over")
+			t.Fatal("Sync shouldn't include the joined room until resync is over")
 		}
 
 		// wait for the state_ids request to arrive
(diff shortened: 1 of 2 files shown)


tests/federation_room_join_partial_state_test.go
```diff
@@ -933,44 +1088,35 @@ func TestPartialStateJoin(t *testing.T) {
...
 		psjResult.FinishStateRequest()
 
 		// the /sync request should now complete, with the new room
-		var syncRes gjson.Result
-		select {
-		case <-time.After(1 * time.Second):
-			t.Fatalf("/sync request request did not complete")
-		case syncRes = <-syncResponseChan:
-		}
+		response, _ = alice.MustSync(t, client.SyncReq{Since: nextBatch})
```
squahtx, line 1116: Now that eager /syncs no longer block, shouldn't we use a `MustSyncUntil` or `awaitPartialStateJoinCompletion`, to avoid the race where the /sync response comes before finishing the partial state join?

tests/federation_room_join_partial_state_test.go
```diff
@@ -177,7 +259,78 @@ func TestPartialStateJoin(t *testing.T) {
...
+		select {
+		case <-syncStarted:
+			// wait for the state_ids request to arrive
+			psjResult.AwaitStateIdsRequest(t)
+			// release the federation /state response
+			psjResult.FinishStateRequest()
+		case <-time.After(time.Second * 5):
+			// even though this should mostly be impossible, make sure we have a timeout
+			t.Fatalf("goroutine didn't start")
+		}
```
squahtx, line 320: This isn't a strong guarantee that the homeserver has received the /sync and started waiting for something to happen yet. It's possible for us to call `FinishStateRequest` before the /sync request is sent out.

I'd suggest
1. sleeping for a few seconds to confirm the homeserver is waiting (and maybe check `responseChan` for sanity)
2. then allowing the partial state join to complete
3. then expecting the sync response

tests/federation_room_join_partial_state_test.go
```diff
@@ -1029,34 +1175,21 @@ func TestPartialStateJoin(t *testing.T) {
...
-		case syncRes = <-syncResponseChan:
-		}
+		response, _ = charlie.MustSync(t, client.SyncReq{Since: nextBatch})
 
-		roomRes := syncRes.Get("rooms.join." + client.GjsonEscape(roomID))
-		if !roomRes.Exists() {
-			t.Fatalf("hs2 /sync completed without join to new room\n")
+		if !response.Get(syncJoinedRoomPath).Exists() {
+			t.Fatal("hs2 /sync completed without join to new room")
 		}
```
squahtx, line 1193: We could do with a comment here explaining that we expect hs2 to fall back to requesting state from hs1 and complete the partial state join.

Also syncs no longer block until joins complete, so we have to use `MustSyncUntil` or `awaitPartialStateJoinCompletion` to avoid creating more flakes.

tests/federation_room_join_partial_state_test.go
```diff
@@ -118,52 +118,134 @@ func TestPartialStateJoin(t *testing.T) {
...
+
+		getEagerSyncReq := func() client.SyncReq {
+			// We track the `next_batch` returned by Alice's eager syncs. However, we
+			// will only _use_ it when we are testing the behaviour of incremental syncs.
+			if incremental {
+				return client.SyncReq{Since: eagerSyncToken}
+			} else {
+				// TODO: making repeated initial syncs might mean Synapse gives us a
+				// sync result from cache. Can we prevent that somehow? Restart the
+				// server would do it, but seems nuclear.
```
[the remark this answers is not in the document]
DMRobertson: Aha, many thanks. (That would explain why this wasn't doing any caching when I trawled through the logs)

tests/federation_room_join_partial_state_test.go
```diff
@@ -118,52 +118,134 @@ func TestPartialStateJoin(t *testing.T) {
...
+	// incremental sync (has a `since` token) separately.  Do this as follows
+	//
+	// 1. Partial join Alice to a remote room.
+	// 2. Have Alice lazy-sync until she sees (1).
+	// 3. Have Alice eager sync. The response should omit the remote room.
+	// 4. Have Alice send a message to the remote room.
+	// 5. Have Alice lazy-sync until she sees (4).
+	// 6. Have Alice eager-sync. The response should omit the remote room.
+	// 7. Allow the resync to complete.
+	// 8. Have Alice eager-sync until she sees the remote room.
```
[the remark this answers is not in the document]
DMRobertson: > Is the idea that...

That's exactly right. I'm trying to test a few things simultaneously (which you might argue is bad practice)L

- (1--3): partial-joining a room doesn't make it appear in an eager sync
- (4--6): activity during the resync doesn't make the room appear in an eager sync
- (7--8): but the resync completion does make the room appear in an eager sync.

tests/federation_room_join_partial_state_test.go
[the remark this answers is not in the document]
DMRobertson: 0beaabf

tests/federation_room_join_partial_state_test.go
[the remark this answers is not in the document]
DMRobertson: That's right.

(3) tests that (1) doesn't show up in an eager sync. (2) is there to try and ensure that the server finished processing (1) before we try the sync.

Similarly, (6) tests (4) doesn't show up in an eager sync, using (5) to try and avoid races.

Then (8) tests that (7) shows up in the eager sync response.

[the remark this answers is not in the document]
DMRobertson: I agree that (3--5) aren't strictly necessary, but I was trying to make the test a little more robust. I, err, don't fully grok synapse's /sync machinery, so could believe that joins or memberships were handled differently to regular message events.

tests/federation_room_join_partial_state_test.go
DMRobertson: 8301105

DMRobertson: err, actually 76d2e5c

tests/federation_room_join_partial_state_test.go
DMRobertson: MustSyncUntil in 76d2e5c

DMRobertson: And comment in e1d3bbe

DMRobertson:
As with the branch this tests, if CI passes I'm inclined to merge given that Sean is away.

PR #590, opened 2023-01-20
Ignore presence updates in test

erikjohnston:
Otherwise the test fails if the server receives a presence update during the partial join


DMRobertson (approved):
I'd also be fine with `func(e gomatrixserverlib.EDU) {}` :shrug:

PR #591, opened 2023-01-26
Fix flaky user directory test

H-Shay:
Hopefully fixes https://github.com/matrix-org/synapse/issues/14902. In the test, there are two directory lookups, one for a user rod and one for a user todd. When I wrote the test, my assumption was that if rod was showing up in the user directory then todd would show up as well, so I did not retry the lookup for todd. It seems now that I was wrong. I added a retry to todd's lookup, and ran the test ~15 times locally. Not only did it not fail, but I saw several runs where todd's lookup was retried - which in the previous iteration would have been a failure -  so I suspect this was the problem.


squahtx (approved):
Looks good. Thanks for fixing the flake!

H-Shay:
Merged, thanks!

Issue #580, opened 2022-09-29
Complement `TestDeviceListUpdates/when_joining_a_room_with_a_remote_user` is flakey


DMRobertson:
https://github.com/matrix-org/synapse/actions/runs/3149623842/jobs/5121449618#step:5:8174

Happened in worker mode---would guess there's some race the test isn't expecting?

richvdh:
https://github.com/matrix-org/synapse/actions/runs/3884827892/jobs/6627963686

DMRobertson:
https://github.com/matrix-org/synapse/actions/runs/3996958036/jobs/6857752983

clokep:
I had this fail twice yesterday on me:

https://github.com/matrix-org/synapse/actions/runs/4127804842/jobs/7131517396
https://github.com/matrix-org/synapse/actions/runs/4127804842/jobs/7131517530

PR #592, opened 2023-01-31
Increase thread safety on variables

realtyem:
While experimenting with Complement and trying to speed up a few things, I noticed some race conditions causing inconsistent results. I started tracking them down one-by-one using the `-race` flag on `go test`. Fixing these allows heavier usage of `t.Parallel()` during tests while sharing a deployment and registering fresh users with `RegisterUser()`(ala, federation_room_join_partial_state_test.go and #471) and it's siblings and derivatives. This alone will not(unfortunately) speed up Complement on Synapse(especially on Github CI), but will pave the way for less odd and inconsistent errors in the future. Hopefully, some flakes will go away as well(looking at you `Client Timeout while awaiting headers`).

Mutex's and RWMutex's were the main go-to, but atomic.AddInt64() was also used in a few places where I felt it was more appropriate.

There is an entire set of data races I couldn't touch from this repo, those caused by [gorilla's mux](https://github.com/gorilla/mux) package. The problem stems from the complement test server that is created to test federation responses. Sometimes a test server is created, then started with `Listen()`, and ***then*** an additional Handler is added on to the server(which is necessary to get the results of the test). This causes a data race in the Router.routes slice when Router.Match is concurrently called.  Unfortunately, this package is [no longer maintained and has been archived](https://github.com/gorilla/mux/issues/659). If someone at matrix.org would like to fork it, I can submit the PR to patch it. I'm still exploring other packages that might meet the requirements as a replacement, but would appreciate suggestions or a fuller list of said requirements than I can get by just reading the source code.

Should be reviewable commit-wise.

Useful links:
* https://stackoverflow.com/questions/36167200/how-safe-are-golang-maps-for-concurrent-read-write-operations

Signed-off-by: Jason Little <realtyem@gmail.com>

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -15,6 +15,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -93,7 +94,7 @@ type CSAPI struct {
 	// True to enable verbose logging
 	Debug bool
 
-	txnID int
+	txnID int64
 }
 
 // UploadContent uploads the provided content with an optional file name. Fails the test on error. Returns the MXC URI.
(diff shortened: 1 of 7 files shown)


MatMaul:
Thanks a lot for this PR, some flakes have been annoying us for a while, like [`TestDeviceListUpdates`](https://github.com/matrix-org/complement/issues/580).

We think your changes make sense, but we also think that having to think about that kind of stuffs in a testing framework is... really annoying :) 

As a first, could you split out the part with the changes on the `internal` framework ?
Those are mostly hidden to people writing tests so we think this is probably ok to have a bit of locking wizardy there.

We are going to put a bit more thinking about that in `#complement:matrix.org` so feel free to come around there.

squahtx:
This mostly touches Complement internals, so I'm taking it off the Synapse team's review queue.

kegsay (approved):
The internal package changes are very sane, thanks!

Complement internals are generally not thread-safe to use, but yes in practice we should make them (specifically the Deployment/Server/ServerRoom).

MatMaul:
For context we also had a [lengthy discussion](https://matrix.to/#/!alCakyySsFIAVfZLDL:matrix.org/$XdzmuxFwlxVd6624eE3A6cIja2-TZFZ_Mu9on4nVvQI?via=matrix.org&via=element.io&via=librepush.net) on should we really use `Parallel` and locking, and we probably want to avoid that later on, but it needs work to split tests in more packages so `go test -p X` can parallelized them.

In the meantime, thanks a lot for that @realtyem , less flakes !

PR #593, opened 2023-02-01
Fix `TestDeviceListUpdates` flakes

### LINKED ISSUE #580: Complement `TestDeviceListUpdates/when_joining_a_room_with_a_remote_user` is flakey

squahtx:
When a remote user has a device list update, an entry is expected to appear in
`device_lists.changed` in the `/sync` response. The local homeserver may then
query the remote homeserver for the update. Some homeservers (Synapse) may emit
extra `device_lists.changed` updates in `/sync` responses after querying keys.

Previously, we attempted to skip over the spurious `device_lists.changed`
updates by doing another sync. This wasn't sufficient, so we make another
remote user send a device list update, and wait for that to appear.

Fixes #580.

---

Fixes https://github.com/matrix-org/synapse/issues/14103.

### DIFF
--- a/tests/csapi/device_lists_test.go
+++ b/tests/csapi/device_lists_test.go
@@ -98,6 +98,43 @@ func TestDeviceListUpdates(t *testing.T) {
 		}
 	}
 
+	// makeBarrier returns a function which tries to act as a barrier for `device_lists.changed`
+	// updates.
+	//
+	// When a remote user has a device list update, an entry is expected to appear in
+	// `device_lists.changed` in the `/sync` response. The local homeserver may then query the
+	// remote homeserver for the update. Some homeservers (Synapse) may emit extra
+	// `device_lists.changed` updates in `/sync` responses after querying keys.
+	//
+	// The barrier tries to ensure that `device_lists.changed` entries resulting from device list
+	// updates and queries before the barrier do not appear in `/sync` responses after the barrier.
+	makeBarrier := func(
+		t *testing.T,
+		deployment *docker.Deployment,
+		observingUser *client.CSAPI,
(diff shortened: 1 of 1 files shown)


tests/csapi/device_lists_test.go
```diff
@@ -98,13 +98,51 @@ func TestDeviceListUpdates(t *testing.T) {
...
+	// makeBarrier returns a function which tries to act as a barrier for `device_lists.changed`
+	// updates.
+	//
+	// When a remote user has a device list update, an entry is expected to appear in
+	// `device_lists.changed` in the `/sync` response. The local homeserver may then query the
+	// remote homeserver for the update. Some homeservers (Synapse) may emit extra
+	// `device_lists.changed` updates in `/sync` responses after querying keys.
+	//
+	// The barrier tries to ensure that `device_lists.changed` entries resulting from device list
+	// updates and queries before the barrier do not appear in `/sync` responses after the barrier.
```
DMRobertson, line 110: Have I understood this correctly? 

We make something new happen in the device lists stream for `barry` and wait to see the result happen. Because 

- barry's change happens after any "no-op" device_list.changed entries, and
- the device list stream is sequential/linear(?)

by the time we've seen `device_lists.changed` for barry we will have already seen the  "no-op" `device_list.changed" (if it was generated at all).

squahtx: That's the idea, yes.

DMRobertson (approved):

PR #597, opened 2023-02-03
Pass lazy loading filter to sync reqs in room summary tests

Fizzadar:
This aligns correctly with the spec and is also seemingly the only reason the test is disabled on synapse, now re-enabled.

### DIFF
--- a/tests/csapi/sync_test.go
+++ b/tests/csapi/sync_test.go
@@ -434,19 +434,33 @@ func TestPresenceSyncDifferentRooms(t *testing.T) {
 	})
 }
 
+func getLazyLoadingSyncFilter() string {
+	j, _ := json.Marshal(map[string]interface{}{
+		"room": map[string]interface{}{
+			"timeline": map[string]interface{}{
+				"lazy_load_members": true,
+			},
+			"state": map[string]interface{}{
+				"lazy_load_members": true,
+			},
+		},
+	})
+	return string(j)
+}
+
 func TestRoomSummary(t *testing.T) {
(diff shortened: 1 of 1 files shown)


S7evinK:
Also see https://github.com/matrix-org/complement/pull/553 and https://github.com/matrix-org/matrix-spec/issues/1325, where we came to the conclusion, for now, that it's not clear by the spec when this field should be added.
Looking at `RoomSummary` and the `m.heroes`, it states that this is **required**, if the room has neither `m.room.name` nor `m.room.canonical_alias`. So given a room can have neither (e.g. DMs), shouldn't Synapse include a `RoomSummary`, no matter if lazy loading or not?

squahtx:
> Also see #553 and [matrix-org/matrix-spec#1325](https://github.com/matrix-org/matrix-spec/issues/1325), where we came to the conclusion, for now, that it's not clear by the spec when this field should be added. Looking at `RoomSummary` and the `m.heroes`, it states that this is **required**, if the room has neither `m.room.name` nor `m.room.canonical_alias`. So given a room can have neither (e.g. DMs), shouldn't Synapse include a `RoomSummary`, no matter if lazy loading or not?

The spec's a little unclear, since it doesn't directly mention when `summary` is required. However it makes sense for the server to send `m.heroes` for DMs, for example. If we take `summary` to be required when any of its fields are required, then Synapse is not spec compliant here.

I've filed this as https://github.com/matrix-org/synapse/issues/15017.

kegsay:
This is over 2 years old. Closing. Please open a new PR if you want to land this.

PR #598, opened 2023-02-06
Encourage (require) sign-offs on PRs

DMRobertson:


squahtx (approved):
lgtm.

It seems a bit weird to be linking to the Synapse contributing guide from complement. Perhaps we need a matrix-org-wide one?

DMRobertson:
> It seems a bit weird to be linking to the Synapse contributing guide from complement. Perhaps we need a matrix-org-wide one?

Agreed. Was just trying to put something together here!

PR #600, opened 2023-02-07
TestRoomState: use non-default power levels to test powerlevels event retrieval

sumnerevans:

### Pull Request Checklist

### DIFF
--- a/tests/csapi/apidoc_room_state_test.go
+++ b/tests/csapi/apidoc_room_state_test.go
@@ -64,19 +64,45 @@ func TestRoomState(t *testing.T) {
 				"visibility":      "public",
 				"preset":          "public_chat",
 				"room_alias_name": "room_alias",
+
+				// Add power level overrides so that it doesn't create the
+				// default power levels and forces the fields to exist.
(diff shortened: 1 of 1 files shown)


tests/csapi/apidoc_room_state_test.go
```diff
@@ -64,19 +64,45 @@ func TestRoomState(t *testing.T) {
...
+					"invite":         2,
+					"kick":           52,
+					"notifications": map[string]interface{}{
+						"room": 53,
+					},
+					"redact":        54,
+					"state_default": 55,
+					// An empty users map should get merged with the default
+					// events map which will include the creator as an admin.
+					"users":         map[string]interface{}{},
```
squahtx, line 85: Synapse is not doing any merging and failing the test as a result. If I remember rightly, the spec is a little unclear about whether merging should happen.

sumnerevans: Yep, I mentioned that here: https://github.com/matrix-org/matrix-spec/issues/492#issuecomment-1421029718

Synapse only merges one level deep.

PR #603, opened 2023-02-09
Faster room joins: Refactor test for user directory update once synced

squahtx:


DMRobertson (approved):
To check: this is just a refactor, and doesn't change test behaviour?

squahtx:
This shouldn't change test behaviour at all.

PR #608, opened 2023-02-13
Test that Synapse will purge a room during resync

DMRobertson:
See https://github.com/matrix-org/synapse/pull/15068.

IDK if we want Synapse-specific tests in Synapse, but short of https://github.com/matrix-org/complement/issues/226 I think this is the least-bad option.

### DIFF
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -3652,7 +3652,9 @@ func TestPartialStateJoin(t *testing.T) {
 
 		serverRoom := createTestRoom(t, server, alice.GetDefaultRoomVersion(t))
 		psjResult := beginPartialStateJoin(t, server, serverRoom, alice)
-		defer psjResult.Destroy(t)
+		// NB: because we do not end up joined to this room at the end of the test,
+		// we do not `defer psjResult.Destroy(t)` as usual; see the comments below
+		// about races.
 
 		t.Log("Alice waits to see her join")
 		alice.MustSyncUntil(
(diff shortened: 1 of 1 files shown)


tests/federation_room_join_partial_state_test.go
```diff
@@ -3636,6 +3637,49 @@ func TestPartialStateJoin(t *testing.T) {
...
 	//   - Join + remote ban works, then cannot rejoin
+
+	t.Run("Purge during resync", func(t *testing.T) {
+		if runtime.Homeserver != runtime.Synapse {
+			// TOOD: Pull this into a Synapse-specific suite when someone figures out how
+			// to do that (https://github.com/matrix-org/complement/issues/226)
+			t.Skipf("Skipping test of Synapse-internal API on %s", runtime.Homeserver)
+		}
+		t.Log("Alice begins a partial join to a room")
+		alice := deployment.RegisterUser(t, "hs1", "t41alice", "secret", true)
```
squahtx, line 3648: I believe t46 is the next unused user ID.
```suggestion
		alice := deployment.RegisterUser(t, "hs1", "t46alice", "secret", true)
```

squahtx (approved):
Test looks good. Thanks for picking this up!

tests/federation_room_join_partial_state_test.go
DMRobertson: Oops, thanks. Copy-paste error.

DMRobertson:
There is a problem with this test: the partial state infrastructure in complement checks to see that a partial state has succeeded by querying /members. But that will fail after the room has been purged. Will think on this tomorrow.

DMRobertson:
Thanks Sean for talking this over with me this morning. Could you take another look?

squahtx (approved):
Looks good!

PR #610, opened 2023-02-14
Fix `TestDeviceListUpdates` flakes, again

squahtx:
When a homeserver (Synapse) performs a partial state join to a room, it
may emit `device_lists.changed` entries for users it thinks are in the
room, despite the room not appearing in the /sync response yet. This is
questionable behaviour, but not forbidden by the spec. Once the
homeserver is fully joined to the room, it may emit another round of
`device_lists.changed` entries for users it knows are definitely in the
room.

These extra entries were causing test cases to advance prematurely and
randomly fail. To fix the tests, we explicitly check for join events
alongside the `device_lists.changed` entries produced by joins. In
effect, we ensure that homeservers are fully joined to the test rooms
before continuing the rest of the tests.

For consistency, we also check for leave events alongside
`device_lists.left` entries.

---

Fixes https://github.com/matrix-org/synapse/issues/14103.

### DIFF
--- a/tests/csapi/device_lists_test.go
+++ b/tests/csapi/device_lists_test.go
@@ -159,6 +159,7 @@ func TestDeviceListUpdates(t *testing.T) {
 		aliceNextBatch = alice.MustSyncUntil(
 			t,
 			client.SyncReq{Since: aliceNextBatch},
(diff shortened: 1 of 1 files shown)


DMRobertson (approved):
Huge thanks for investigating this!

PR #611, opened 2023-02-15
Fix `TestKnocking` flakiness by waiting for knock membership to arrive

squahtx:
Wait for up to 5 seconds for the knock membership to arrive over
federation before failing the "Users in the room see a user's membership
update when they knock" test case.

---

Fixes https://github.com/matrix-org/synapse/issues/14226.

### DIFF
--- a/tests/knocking_test.go
+++ b/tests/knocking_test.go
@@ -143,8 +143,15 @@ func knockingBetweenTwoUsersTest(t *testing.T, roomID string, inRoomUser, knocki
 	})
 
 	t.Run("Users in the room see a user's membership update when they knock", func(t *testing.T) {
-		// check the membership seen over the federation
+		// wait for the membership to arrive over federation
+		start := time.Now()
 		knockerState := serverRoom.CurrentState("m.room.member", knockingUser.UserID)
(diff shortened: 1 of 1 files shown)


tests/knocking_test.go
```diff
@@ -143,8 +143,15 @@ func knockingBetweenTwoUsersTest(t *testing.T, roomID string, inRoomUser, knocki
...
 
 	t.Run("Users in the room see a user's membership update when they knock", func(t *testing.T) {
-		// check the membership seen over the federation
+		// wait for the membership to arrive over federation
+		start := time.Now()
 		knockerState := serverRoom.CurrentState("m.room.member", knockingUser.UserID)
+		for knockerState == nil && time.Since(start) < 5 * time.Second {
+			time.Sleep(100 * time.Millisecond)
+			knockerState = serverRoom.CurrentState("m.room.member", knockingUser.UserID)
+		}
```
squahtx, line 152: The membership we're waiting for is created by the two test cases above.
I'm not sure why this test only started flaking recently. It should always have been flaky.

DMRobertson (approved):
Many thanks for digging into this. LGTM.

PR #612, opened 2023-02-15
Add test for failover of federated joins

squahtx:
When performing a federated join, the homeserver accepts a list of
servers to try joining via. When the first remote server in the list is
not in the room, the federated join ought to try the rest of the list
instead of failing immediately.

---

Synapse currently fails the test. https://github.com/matrix-org/synapse/pull/15074 fixes it.


DMRobertson (approved):
SGTM!

PR #617, opened 2023-02-17
More tests for leaves during a partial state resync

DMRobertson:
This
- tests https://github.com/matrix-org/synapse/issues/12802 (which we seem to have got for free by solving https://github.com/matrix-org/synapse/issues/12801)
- Unskips https://github.com/matrix-org/complement/pull/581

Closes https://github.com/matrix-org/synapse/issues/12802

### DIFF
--- a/tests/federation_room_join_partial_state_test.go
+++ b/tests/federation_room_join_partial_state_test.go
@@ -3520,8 +3520,62 @@ func TestPartialStateJoin(t *testing.T) {
 			}
 		})
 
+		// Test that the original joiner can leave during the resync, even after someone else has joined
+		t.Run("works after a second partial join", func(t *testing.T) {
+			alice := deployment.RegisterUser(t, "hs1", "t47alice", "secret", false)
+			bob := deployment.RegisterUser(t, "hs1", "t47bob", "secret", false)
+			handleTransactions := federation.HandleTransactionRequests(
+				// Accept all PDUs and EDUs
+				func(e *gomatrixserverlib.Event) {},
+				func(e gomatrixserverlib.EDU) {},
+			)
+			server := createTestServer(t, deployment, handleTransactions)
+			cancel := server.Listen()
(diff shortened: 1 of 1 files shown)


tests/federation_room_join_partial_state_test.go
```diff
@@ -3719,8 +3719,61 @@ func TestPartialStateJoin(t *testing.T) {
 			// Cleanup.
 			psjResult.FinishStateRequest()
 		})
-		// TODO: tests which assert that:
-		//   - Join + remote ban works, then cannot rejoin
```
H-Shay: I see this comment is being removed but it seems that the test below is testing something slightly different (as it only tests that the ban causes a leave, rather than then also testing after the successful leave-ban that she cannot rejoin)?

tests/federation_room_join_partial_state_test.go
```diff
@@ -3398,128 +3398,382 @@ func TestPartialStateJoin(t *testing.T) {
...
+
+			t.Log("Alice waits to see her join")
+			aliceNextBatch := alice.MustSyncUntil(
+				t,
+				client.SyncReq{Filter: buildLazyLoadingSyncFilter(nil)},
+				client.SyncJoinedTo(alice.UserID, serverRoom.RoomID),
+			)
+
+			t.Log("A resident server user bans Alice from the room.")
+			kickEvent := server.MustCreateEvent(t, serverRoom, b.Event{
```
DMRobertson, line 3748: this should be called `banEvent`.

tests/federation_room_join_partial_state_test.go
DMRobertson: Actually this was fine. (I was worried that I'd have to change the send_join response, but if I do it before the resync I can test that the joining HS still enforces the ban.)

H-Shay (approved):
LGTM, thanks for adding these!

PR #619, opened 2023-03-01
Fix test for forgotten rooms not showing up in initial /sync

squahtx:
Previously, the test would always pass as it did not provide the
`include_leave` option in the room filter and did not check the
`rooms.leave` key in the response.

### DIFF
--- a/tests/csapi/apidoc_room_forget_test.go
+++ b/tests/csapi/apidoc_room_forget_test.go
@@ -1,6 +1,7 @@
 package csapi_tests
 
 import (
+	"encoding/json"
 	"net/http"
 	"net/url"
 	"testing"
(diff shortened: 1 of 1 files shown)


clokep (approved):
Looks like the original sytest https://github.com/matrix-org/sytest/blob/4784b7ed3337962e7938d368d2748953d627d5bf/tests/30rooms/31forget.pl#L88-L99 made this mistake too?

(As an aside -- should we delete that test since it was ported?)

squahtx:
> Looks like the original sytest https://github.com/matrix-org/sytest/blob/4784b7ed3337962e7938d368d2748953d627d5bf/tests/30rooms/31forget.pl#L88-L99 made this mistake too?
> 
> (As an aside -- should we delete that test since it was ported?)

Yep, the original sytest looks equally buggy.
I've PRed the deletion of the sytest in https://github.com/matrix-org/sytest/pull/1338.

squahtx:
Thanks for the reviews!

PR #621, opened 2023-03-03
Add a test which checks that the transaction ID is serialised in the event

sandhose:
This adds a test which checks that the `transaction_id` is indeed present when querying an event with `/rooms/{roomId}/event/{eventId}`. Synapse implementation was bugged, and I fixed it in matrix-org/synapse#15174

### DIFF
--- a/tests/csapi/txnid_test.go
+++ b/tests/csapi/txnid_test.go
@@ -1,0 +1,40 @@
+package csapi_tests
+
+import (
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/runtime"
+	"github.com/tidwall/gjson"
+	"testing"
+)
+
+// TestTxnInEvent checks that the transaction ID is present when getting the event from the /rooms/{roomID}/event/{eventID} endpoint.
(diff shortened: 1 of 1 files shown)


tests/csapi/txnid_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/matrix-org/complement/internal/b"
+	"github.com/matrix-org/complement/internal/client"
+	"github.com/matrix-org/complement/runtime"
+	"github.com/tidwall/gjson"
+	"testing"
+)
+
+// TestTxnInEvent checks that the transaction ID is present when getting the event from the /rooms/{roomID}/event/{eventID} endpoint.
+func TestTxnInEvent(t *testing.T) {
+	// Dendrite implementation is broken
```
clokep, line 13: If you could file an issue (and reference it here) that would be lovely!

clokep (approved):
Seems reasonable.

clokep:
(Well assuming tests pass after we merge the Synapse PR. 😄 )

clokep:
LGTM with the dendrite issue filed. 👍

tests/csapi/txnid_test.go
sandhose: Done!

PR #622, opened 2023-03-04
Tests for transaction ID semantics

hughns:
This adds Client-Server API tests relation to transaction identifiers:

1. That requests are idempotent within a single client session
2. That transaction ID can be re-used across client sessions
3. That the transaction ID is present in /sync response to deal with "local echo"

Signed-off-by: Hugh Nimmo-Smith <hughns@element.io>

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -264,8 +264,14 @@ func (c *CSAPI) SetPushRule(t *testing.T, scope string, kind string, ruleID stri
 // SendEventUnsynced sends `e` into the room.
 // Returns the event ID of the sent event.
 func (c *CSAPI) SendEventUnsynced(t *testing.T, roomID string, e b.Event) string {
-	t.Helper()
 	txnID := int(atomic.AddInt64(&c.txnID, 1))
+	return c.SendEventUnsyncedWithTxnID(t, roomID, e, txnID)
+}
+
+// SendEventUnsynced sends `e` into the room.
+// Returns the event ID of the sent event.
+func (c *CSAPI) SendEventUnsyncedWithTxnID(t *testing.T, roomID string, e b.Event, txnID int) string {
+	t.Helper()
 	paths := []string{"_matrix", "client", "v3", "rooms", roomID, "send", e.Type, strconv.Itoa(txnID)}
 	if e.StateKey != nil {
 		paths = []string{"_matrix", "client", "v3", "rooms", roomID, "state", e.Type, *e.StateKey}
@@ -438,7 +444,34 @@ func (c *CSAPI) LoginUser(t *testing.T, localpart, password string) (userID, acc
 	return userID, accessToken, deviceID
 }
 
-//RegisterUser will register the user with given parameters and
+// LoginUserWithDeviceID will log in to a homeserver on an existing device
+func (c *CSAPI) LoginUserWithDeviceID(t *testing.T, localpart, password, deviceID string) (userID, accessToken string) {
+	t.Helper()
+	reqBody := map[string]interface{}{
+		"identifier": map[string]interface{}{
+			"type": "m.id.user",
+			"user": localpart,
+		},
+		"device_id": deviceID,
+		"password":  password,
+		"type":      "m.login.password",
+	}
+	res := c.MustDoFunc(t, "POST", []string{"_matrix", "client", "v3", "login"}, WithJSONBody(t, reqBody))
+
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		t.Fatalf("unable to read response body: %v", err)
+	}
+
+	userID = gjson.GetBytes(body, "user_id").Str
+	accessToken = gjson.GetBytes(body, "access_token").Str
+	if gjson.GetBytes(body, "device_id").Str != deviceID {
+		t.Fatalf("device_id returned by login does not match the one requested")
+	}
+	return userID, accessToken
+}
+
+// RegisterUser will register the user with given parameters and
 // return user ID & access token, and fail the test on network error
 func (c *CSAPI) RegisterUser(t *testing.T, localpart, password string) (userID, accessToken, deviceID string) {
 	t.Helper()
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -266,14 +266,14 @@ func (c *CSAPI) SetPushRule(t *testing.T, scope string, kind string, ruleID stri
 func (c *CSAPI) SendEventUnsynced(t *testing.T, roomID string, e b.Event) string {
 	t.Helper()
 	txnID := int(atomic.AddInt64(&c.txnID, 1))
-	return c.SendEventUnsyncedWithTxnID(t, roomID, e, txnID)
+	return c.SendEventUnsyncedWithTxnID(t, roomID, e, strconv.Itoa(txnID))
 }
 
 // SendEventUnsynced sends `e` into the room.
 // Returns the event ID of the sent event.
-func (c *CSAPI) SendEventUnsyncedWithTxnID(t *testing.T, roomID string, e b.Event, txnID int) string {
+func (c *CSAPI) SendEventUnsyncedWithTxnID(t *testing.T, roomID string, e b.Event, txnID string) string {
 	t.Helper()
-	paths := []string{"_matrix", "client", "v3", "rooms", roomID, "send", e.Type, strconv.Itoa(txnID)}
+	paths := []string{"_matrix", "client", "v3", "rooms", roomID, "send", e.Type, txnID}
 	if e.StateKey != nil {
 		paths = []string{"_matrix", "client", "v3", "rooms", roomID, "state", e.Type, *e.StateKey}
 	}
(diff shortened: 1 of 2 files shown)


internal/client/client.go
```diff
@@ -264,8 +264,14 @@ func (c *CSAPI) SetPushRule(t *testing.T, scope string, kind string, ruleID stri
 // SendEventUnsynced sends `e` into the room.
 // Returns the event ID of the sent event.
 func (c *CSAPI) SendEventUnsynced(t *testing.T, roomID string, e b.Event) string {
-	t.Helper()
 	txnID := int(atomic.AddInt64(&c.txnID, 1))
```
kegsay, line 267:
```suggestion
	t.Helper()
	txnID := int(atomic.AddInt64(&c.txnID, 1))
```

All helper functions need `t.Helper()` on them.

tests/csapi/txnid_test.go
```diff
@@ -0,0 +1,178 @@
...
+)
+
+func mustHaveTransactionID(t *testing.T, roomID, eventID string) client.SyncCheckOpt {
+	return client.SyncTimelineHas(roomID, func(r gjson.Result) bool {
+		if r.Get("event_id").Str == eventID {
+			if !r.Get("unsigned.transaction_id").Exists() {
+				t.Fatalf("Event %s in room %s should have a 'transaction_id', but it did not", eventID, roomID)
+			}
+
+			return true
```
kegsay, line 19: You don't ever seem to assert that the `transaction_id` is actually the thing you gave originally? The type isn't even asserted currently so this would pass:
```js
{
  content: { ... },
  unsigned: {
    transaction_id: {
      key: "val"
    }
  }
}
```

tests/csapi/txnid_test.go
```diff
@@ -0,0 +1,178 @@
...
+		return false
+	})
+}
+
+func mustNotHaveTransactionID(t *testing.T, roomID, eventID string) client.SyncCheckOpt {
+	return client.SyncTimelineHas(roomID, func(r gjson.Result) bool {
+		if r.Get("event_id").Str == eventID {
+			res := r.Get("unsigned.transaction_id")
+			if res.Exists() {
+				t.Fatalf("Event %s in room %s should NOT have a 'transaction_id', but it did (%s)", eventID, roomID, res.Str)
```
kegsay, line 31: Please put the complete path here, lest poor developers end up looking for a top-level `transaction_id` field. You could even log the entire event here with `r.Raw`.

tests/csapi/txnid_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	if eventID1 != eventID2 {
+		t.Fatalf("Expected event IDs to be the same, but they were not")
+	}
+
+	// even if we change the content we should still get back the same event ID as transaction ID is the same
+	eventID3 := c1.SendEventUnsyncedWithTxnID(t, roomID1, event2, txnId)
+
+	if eventID1 != eventID3 {
+		t.Fatalf("Expected event IDs to be the same even with different content, but they were not")
```
kegsay, line 169: These error messages are okay but could be more ergonomic. It would be nice to log what the inputs were, in this case `eventID1` and `eventID3`, which you can do by also using an existing heper, in this case `must.EqualStr(got, want, msg)`.

internal/client/client.go
```diff
@@ -438,7 +444,34 @@ func (c *CSAPI) LoginUser(t *testing.T, localpart, password string) (userID, acc
 	return userID, accessToken, deviceID
 }
 
-//RegisterUser will register the user with given parameters and
+// LoginUserWithDeviceID will log in to a homeserver on an existing device
+func (c *CSAPI) LoginUserWithDeviceID(t *testing.T, localpart, password, deviceID string) (userID, accessToken string) {
```
kegsay, line 448: We already have a login method: `LoginUser`. The reason why you don't want to use it is because you can't specify the device ID. Rather than duplicating the entire `LoginUser` function, add functional options to `LoginUser` to set the device ID. Something like:
```go
type LoginOpt func(map[string]interface{})

func WithDeviceID(deviceID string) LoginOpt {
    return func(loginBody map[string]interface{}) {
        loginBody["device_id"] = deviceID
    }
}
```

then change `LoginUser` to be:
```go
func (c *CSAPI) LoginUser(t *testing.T, localpart, password string, opts ...LoginOpt) (userID, accessToken, deviceID string) {
    // .. existing code up to creation of reqBody
    for _, opt := range opts {
        opt(reqBody)
    }
}
```

This is further described in https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

kegsay (changes requested):

tests/csapi/txnid_test.go
hughns: This was intentional as the "local echo" the presence is the pertinent aspect. However, I will add this for completeness.

internal/client/client.go
```diff
@@ -266,7 +266,14 @@ func (c *CSAPI) SetPushRule(t *testing.T, scope string, kind string, ruleID stri
 func (c *CSAPI) SendEventUnsynced(t *testing.T, roomID string, e b.Event) string {
 	t.Helper()
 	txnID := int(atomic.AddInt64(&c.txnID, 1))
-	paths := []string{"_matrix", "client", "v3", "rooms", roomID, "send", e.Type, strconv.Itoa(txnID)}
+	return c.SendEventUnsyncedWithTxnID(t, roomID, e, strconv.Itoa(txnID))
+}
+
+// SendEventUnsynced sends `e` into the room.
```
DMRobertson, line 272:
```suggestion
// SendEventUnsyncedWithTxnID sends `e` into the room with a prescribed transaction ID.
// This is useful for writing tests that interrogate transaction semantics.
```

internal/client/client.go
```diff
@@ -438,7 +458,7 @@ func (c *CSAPI) LoginUser(t *testing.T, localpart, password string) (userID, acc
 	return userID, accessToken, deviceID
 }
 
-//RegisterUser will register the user with given parameters and
+// RegisterUser will register the user with given parameters and
 // return user ID & access token, and fail the test on network error
```
DMRobertson, line 462:
```suggestion
// return user ID, access token and device ID. It fails the test on network error.
```

tests/csapi/txnid_test.go
```diff
@@ -22,20 +25,192 @@ func TestTxnInEvent(t *testing.T) {
...
 		Type: "m.room.message",
 		Content: map[string]interface{}{
 			"msgtype": "m.text",
 			"body":    "first",
 		},
-	})
+	}, txnId)
 
 	// The transaction ID should be present on the GET /rooms/{roomID}/event/{eventID} response.
 	res := c.MustDoFunc(t, "GET", []string{"_matrix", "client", "v3", "rooms", roomID, "event", eventID})
```
DMRobertson, line 39: Slightly worried that making this unsynced introduces races. OTOH we're being quite reasonable in requiring a read-after-write consistency here; it's probably fine.

tests/csapi/txnid_test.go
```diff
@@ -22,20 +25,192 @@ func TestTxnInEvent(t *testing.T) {
...
 	body := client.ParseJSON(t, res)
 	result := gjson.ParseBytes(body)
-	if !result.Get("unsigned.transaction_id").Exists() {
-		t.Fatalf("Event did not have a 'transaction_id' on the GET /rooms/%s/event/%s response", roomID, eventID)
+	unsignedTxnId := result.Get("unsigned.transaction_id")
+	if !unsignedTxnId.Exists() {
+		t.Fatalf("Event did not have a 'unsigned.transaction_id' on the GET /rooms/%s/event/%s response", roomID, eventID)
+	}
+
+	must.EqualStr(t, unsignedTxnId.Str, txnId, fmt.Sprintf("Event did not have a 'unsigned.transaction_id' on GET /rooms/%s/event/%s response", eventID, roomID))
```
DMRobertson, line 47: I think this error message should say "had an incorrect transaction id" rather than "did not have a  transaction id"? Similarly for line 59 below

tests/csapi/txnid_test.go
```diff
@@ -22,20 +25,192 @@ func TestTxnInEvent(t *testing.T) {
...
+		},
+	}, txnId)
+
+	// When syncing, we should find the event and it should have a transaction ID on the first client.
+	c1.MustSyncUntil(t, client.SyncReq{}, mustHaveTransactionID(t, roomID, eventID, txnId))
+
+	// Create a second client, inheriting the first device ID.
+	c2 := deployment.Client(t, "hs1", "")
+	c2.UserID, c2.AccessToken, _ = c2.LoginUser(t, "alice", "password", client.WithDeviceID(c1.DeviceID))
+	c2.DeviceID = c1.DeviceID
```
DMRobertson, line 117: Is this something we should be asserting? (I.e. that logging in with a chosen device ID doesn't give you out a new device ID?) Ditto for 155 below.

DMRobertson (commented):
Other than the error messages I pointed out, I think I only have questions and comments rather than hard changes:

DMRobertson (approved):
LGTM assuming CI is happy!

PR #624, opened 2023-03-13
Add federation ACL test

S7evinK:
Test for https://github.com/matrix-org/dendrite/issues/3004

### DIFF
--- a/tests/federation_acl_test.go
+++ b/tests/federation_acl_test.go
@@ -12,6 +12,7 @@ import (
 
 // Test for https://github.com/matrix-org/dendrite/issues/3004
 func TestACLs(t *testing.T) {
+	// runtime.SkipIf(t, runtime.Dendrite) // needs https://github.com/matrix-org/dendrite/pull/3008
 	// 1. Prepare 3 or more servers. 1st will be room host, 2nd will be blocked with m.room.server_acl and 3rd server will be affected by this issue. 1st and 2nd servers don't have to be powered by dendrite.
 	deployment := Deploy(t, b.Blueprint{
 		Name: "federation_three_homeservers",
@@ -51,10 +52,6 @@ func TestACLs(t *testing.T) {
 	bob := deployment.Client(t, "hs2", "@bob:hs2")
 	charlie := deployment.Client(t, "hs3", "@charlie:hs3")
 
-	/*
-	   You can set m.room.server_acl again, to "fix" this.
-	*/
-
 	// 2. Create room on 1st server
 	roomID := alice.CreateRoom(t, map[string]interface{}{"preset": "public_chat"})
 	aliceSince := alice.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(alice.UserID, roomID))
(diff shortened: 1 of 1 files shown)


tests/federation_acl_test.go
```diff
@@ -0,0 +1,142 @@
...
+		Content: map[string]interface{}{
+			"allow":             []string{"*"},
+			"allow_ip_literals": true,
+			"deny":              []string{"hs2"},
+		},
+	})
+	// wait for the ACL to show up on hs2
+	bob.MustSyncUntil(t, client.SyncReq{Since: bobSince}, client.SyncTimelineHasEventID(roomID, eventID))
+
+	// 5. Join from 3rd (dendrite) server.
```
DMRobertson, line 92: I'm guessing these comments are cribbed from the issue report? It'd be nice to keep Dendrite homeserver-agnostic where possible.

tests/federation_acl_test.go
```diff
@@ -0,0 +1,142 @@
...
+	// wait for the sentinel event to come down sync
+	alice.MustSyncUntil(t, client.SyncReq{Since: aliceSince}, client.SyncTimelineHasEventID(sentinelRoom, sentinelEventID))
+	charlie.MustSyncUntil(t, client.SyncReq{Since: charlieSince}, client.SyncTimelineHasEventID(sentinelRoom, sentinelEventID))
+
+	// Verify with alice and charlie that we never received eventID
+	for _, user := range []*client.CSAPI{alice, charlie} {
+		syncResp, _ := user.MustSync(t, client.SyncReq{})
+
+		// we don't expect eventID (blocked) to be in the sync response
+		events := syncResp.Get(fmt.Sprintf("rooms.join.%s.timeline.events", client.GjsonEscape(roomID))).Array()
```
DMRobertson, line 125: What guarantees that this initial sync response includes the sentinel event? (Synapse caches initial sync responses, though this is [sneakily disabled in complement](https://github.com/matrix-org/synapse/blob/22cc93afe38d34c859d8863a99996e7e72ca1733/docker/complement/conf/workers-shared-extra.yaml.j2#L115-L117).)

tests/federation_acl_test.go
DMRobertson:
```suggestion
	// 5. Join from 3rd server.
```

DMRobertson (approved):
Seems sane, I've left a few nitpicks though

tests/federation_acl_test.go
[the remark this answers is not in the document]
S7evinK: Sytest is testing ACLs for servers already joined to the room, so HS1 and HS2 are present in the room, then HS2 is banned and verified that it can't do anything.

This Complement test joins HS1 and HS2, bans HS2 and HS3 joins. Since Dendrite handles most of the events as outliers when joining over federation, the ACLs which ban HS2 aren't "loaded", resulting in messages etc. from HS2 to appear on HS3.

S7evinK:
CI is failing because of
> Merge 388dc53171a1adbfbc2d417be0fbf6915e8ec446 into 69d1326f322e93be0948e28fad470737637b1fb8

not sure if this is actually the intended behavior or if we should do [this](https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit) instead.

DMRobertson:
> CI is failing because of
> 
> > Merge [388dc53](https://github.com/matrix-org/complement/commit/388dc53171a1adbfbc2d417be0fbf6915e8ec446) into [69d1326](https://github.com/matrix-org/complement/commit/69d1326f322e93be0948e28fad470737637b1fb8)
> 
> not sure if this is actually the intended behavior or if we should do [this](https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit) instead.

I think what you mean is: if we tested against the merge of https://github.com/matrix-org/dendrite/pull/3008 with dendrite main, we wouldn't have seen this error? Sounds like it'd be nice to have (since Dendrite's CI will test the PR by running against that merge commit)

DMRobertson (approved):
Thanks, LGTM!

PR #626, opened 2023-03-14
Deflake faster joins tests by leaving rooms as part of test cleanup

squahtx:
Many of the faster joins test flakes are due to the homeserver under
test failing to contact Complement homeservers after they have been
torn down. When this happens, subsequent tests can fail if they use a
Complement homeserver that happens to have the same hostname:port as one
which the homeserver under test has previously marked as offline.

Make the homeserver under test leave rooms at the end of each test, so
that it stops trying to contact the Complement homeservers in the room.

Signed-off-by: Sean Quah <seanq@matrix.org>

---

Reviewable commit by commit.

NB: you can tease out the flakes by forcing Complement to reuse ports whenever possible:
```diff
diff --git a/internal/federation/server.go b/internal/federation/server.go
index ed7974d..4c4ccbe 100644
--- a/internal/federation/server.go
+++ b/internal/federation/server.go
@@ -455,7 +455,10 @@ func (s *Server) Listen() (cancel func()) {
        var wg sync.WaitGroup
        wg.Add(1)

-       ln, err := net.Listen("tcp", ":0") //nolint
+       ln, err := net.Listen("tcp", ":63333") //nolint
+       if err != nil {
+               ln, err = net.Listen("tcp", ":63334") //nolint
+       }
        if err != nil {
                s.t.Fatalf("ListenFederationServer: net.Listen failed: %s", err)
        }
```

PR #627, opened 2023-03-14
Deflake faster joins device list tests by waiting for leave event

squahtx:
Many of the faster joins test flakes are due to the homeserver under
test failing to contact Complement homeservers after they have been
torn down. When this happens, subsequent tests can fail if they use a
Complement homeserver that happens to have the same hostname:port as one
which the homeserver under test has previously marked as offline.

Wait for the homeserver under test to finish broadcasting its leave at
the end of the device list tests.

---

Builds on #626. Ignore the first 3 commits, since they come from that PR.
Reviewable commit by commit.

### DIFF
--- a/internal/federation/server.go
+++ b/internal/federation/server.go
@@ -339,7 +339,7 @@ func (s *Server) MustCreateEvent(t *testing.T, room *ServerRoom, ev b.Event) *go
 
 // MustJoinRoom will make the server send a make_join and a send_join to join a room
 // It returns the resultant room.
-func (s *Server) MustJoinRoom(t *testing.T, deployment *docker.Deployment, remoteServer gomatrixserverlib.ServerName, roomID string, userID string) *ServerRoom {
+func (s *Server) MustJoinRoom(t *testing.T, deployment *docker.Deployment, remoteServer gomatrixserverlib.ServerName, roomID string, userID string, partialState ...bool) *ServerRoom {
 	t.Helper()
 	origin := gomatrixserverlib.ServerName(s.serverName)
 	fedClient := s.FederationClient(deployment)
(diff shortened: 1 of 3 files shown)


squahtx:
CI flaked due to the issue fixed by #628.

internal/federation/server.go
```diff
@@ -339,7 +339,7 @@ func (s *Server) MustCreateEvent(t *testing.T, room *ServerRoom, ev b.Event) *go
 
 // MustJoinRoom will make the server send a make_join and a send_join to join a room
 // It returns the resultant room.
-func (s *Server) MustJoinRoom(t *testing.T, deployment *docker.Deployment, remoteServer gomatrixserverlib.ServerName, roomID string, userID string) *ServerRoom {
+func (s *Server) MustJoinRoom(t *testing.T, deployment *docker.Deployment, remoteServer gomatrixserverlib.ServerName, roomID string, userID string, partialState ...bool) *ServerRoom {
```
DMRobertson, line 342: I guess this is go's idiom for an optional argument? Feels a bit odd to allow an arbitrary length array/slice, but I guess it's easier than updating every call site?

DMRobertson (approved):
Looks like this was painful to diagnose; thank you for persisting.

internal/federation/server.go
squahtx: Not sure actually. Functional varargs (https://stackoverflow.com/a/26326418) is the pattern [MustDoFunc uses](https://github.com/matrix-org/complement/blob/5c444e7d440feb4bde44f332fd0df8f67ddb6f3d/internal/client/client.go#L584). Here I didn't bother with them since there's only one optional argument.

PR #629, opened 2023-03-16
Fix incorrect refactor of Leave during resync faster joins test

squahtx:
We need to start expecting leave events before the test user leaves the
room, not after.

Introduced in #626.


DMRobertson (approved):
Oops, sorry I missed this!

PR #631, opened 2023-03-23
Gracefully stop containers

S7evinK:
If we're running Dendrite with coverage enabled, we need to gracefully stop the containers, as otherwise the coverage data can't be written to disk. `ContainerStop` will still kill the container if it fails to stop within 1 second.

As this will be used for every homeserver implementation, an alternative could be to check `runtime.Homeserver` and only always gracefully stop containers if it is Dendrite.

### DIFF
--- a/internal/docker/deployer.go
+++ b/internal/docker/deployer.go
@@ -147,18 +147,24 @@ func (d *Deployer) Deploy(ctx context.Context, blueprintName string) (*Deploymen
 	return dep, lastErr
 }
 
-// Destroy a deployment. This will stop/kill all running containers.
+// Destroy a deployment. This will kill all running containers.
 func (d *Deployer) Destroy(dep *Deployment, printServerLogs bool, testName string, failed bool) {
-	timeout := 1 * time.Second
 	for _, hsDep := range dep.HS {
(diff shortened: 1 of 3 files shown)


internal/docker/deployer.go
```diff
@@ -147,24 +147,18 @@ func (d *Deployer) Deploy(ctx context.Context, blueprintName string) (*Deploymen
...
-			// the logs to be flushed.
-			timeout := 1 * time.Second
-			err := d.Docker.ContainerStop(context.Background(), hsDep.ContainerID, &timeout)
-			if err != nil {
-				log.Printf("Destroy: Failed to destroy container %s : %s\n", hsDep.ContainerID, err)
-			}
-
 			printLogs(d.Docker, hsDep.ContainerID, hsDep.ContainerID)
-		} else {
-			err := d.Docker.ContainerKill(context.Background(), hsDep.ContainerID, "KILL")
```
kegsay, line 164: We still need the kill.

S7evinK: I've modified the PR to introduce a `ContainerKillFunc`, this defaults to `ContainerKill` but can be overwritten by homeserver implementations, if they need a graceful stop.

kegsay (approved):

PR #634, opened 2023-04-06
Fix CI by ensuring `gotestfmt` is on PATH, plus cleanups

DMRobertson:
See commits for details.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,6 +61,7 @@ jobs:
           go get -v github.com/gotesttools/gotestfmt/v2/cmd/gotestfmt@latest
           mkdir .gotestfmt/github -p
           cp .ci/complement_package.gotpl .gotestfmt/github/package.gotpl
+          gotestfmt -help
 
       - name: "Checkout corresponding ${{ matrix.homeserver }} branch"
         shell: bash
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ jobs:
   complement-internal:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2 # Checkout complement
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -53,27 +48,20 @@ jobs:
...
       # to do this so we can _be_ the host when running Complement so we can snaffle all the ports. If
       # we run Complement _in_ Docker then we can't -p all high numbered ports which then breaks federation
       # servers which listen on random high numbered ports.
       - name: "Install Complement Dependencies"
-        # We don't need to install Go because it is included on the Ubuntu 20.04 image:
-        # See https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md specifically GOROOT_1_17_X64
         run: |
           sudo apt-get update && sudo apt-get install -y libolm3 libolm-dev
-          go get -v github.com/gotesttools/gotestfmt/v2/cmd/gotestfmt@latest
+          go install -v github.com/gotesttools/gotestfmt/v2/cmd/gotestfmt@latest
```
clokep, line 61: Probably a version of go changing stuff?

.github/workflows/ci.yaml
```diff
@@ -53,27 +48,20 @@ jobs:
...
           mkdir .gotestfmt/github -p
           cp .ci/complement_package.gotpl .gotestfmt/github/package.gotpl
+          gotestfmt -help
```
clokep, line 64: This is just to fail fast if it doesn't exist?

clokep (approved):

.github/workflows/ci.yaml
DMRobertson: Exactly!

.github/workflows/ci.yaml
DMRobertson: I think so. I didn't dig too hard here!

PR #647, opened 2023-08-18
Fix flakey test due to incorrectly running in parallel

erikjohnston:
I think these tests were somehow interacting with each other poorly when run in parallel, leading to flakes. They are only two tests so may as well try removing the parallel bit and seeing if it helps


DMRobertson (approved):
:shrug: worth a try

PR #648, opened 2023-09-04
Fix flakey 'query remote device keys' test

erikjohnston:
i.e. `TestFederationKeyUploadQuery/Can_query_remote_device_keys_using_POST` / `Can query remote device keys using POST`

### DIFF
--- a/tests/federation_upload_keys_test.go
+++ b/tests/federation_upload_keys_test.go
@@ -80,13 +80,17 @@ func TestFederationKeyUploadQuery(t *testing.T) {
 
 	// sytest: Can query remote device keys using POST
 	t.Run("Can query remote device keys using POST", func(t *testing.T) {
+		// Device list changes only come down incremental syncs, so we do an
+		// initial sync up front.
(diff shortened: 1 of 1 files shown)


tests/federation_upload_keys_test.go
```diff
@@ -80,13 +80,17 @@ func TestFederationKeyUploadQuery(t *testing.T) {
 
 	// sytest: Can query remote device keys using POST
 	t.Run("Can query remote device keys using POST", func(t *testing.T) {
+		// Device list changes only come down incremental syncs, so we do an
+		// initial sync up front.
+		_, nextBatch := bob.MustSync(t, client.SyncReq{})
```
DMRobertson, line 85: This seems reasonable---but how did the test ever pass without this?

DMRobertson (approved):

tests/federation_upload_keys_test.go
erikjohnston: I think that things were slow enough that the device list update wouldn't make it down the first `/sync`?

PR #649, opened 2023-09-05
Fix flakey key upload test

erikjohnston:
We waited for the device list update to come down sync before calling `/keys/query`. However, the update going down sync occasionally was for the *initial* key upload, and so the call to `/keys/query` would return the old key and the test would fail.

To fix that, wait for the initial key to come down sync, and *then* run the test.

Closes https://github.com/matrix-org/synapse/issues/16020 (maybe?)

### DIFF
--- a/tests/federation_upload_keys_test.go
+++ b/tests/federation_upload_keys_test.go
@@ -22,6 +22,9 @@ func TestFederationKeyUploadQuery(t *testing.T) {
 	alice := deployment.Client(t, "hs1", "@alice:hs1")
 	bob := deployment.Client(t, "hs2", "@bob:hs2")
 
+	// Do an initial sync so that we can see the changes come down sync.
+	_, nextBatchBeforeKeyUpload := bob.MustSync(t, client.SyncReq{})
+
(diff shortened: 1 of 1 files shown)


tests/federation_upload_keys_test.go
```diff
@@ -80,9 +83,20 @@ func TestFederationKeyUploadQuery(t *testing.T) {
...
+		// notification to go down /sync.
+		nextBatch := bob.MustSyncUntil(t, client.SyncReq{Since: nextBatchBeforeKeyUpload}, func(clientUserID string, topLevelSyncJSON gjson.Result) error {
+			devicesChanged := topLevelSyncJSON.Get("device_lists.changed")
+			if devicesChanged.Exists() {
+				for _, userID := range devicesChanged.Array() {
+					if userID.Str == alice.UserID {
+						return nil
+					}
+				}
+			}
```
DMRobertson, line 97: there might be some clever way to use JSONCheckOffAllowUnwanted here but I usually find that more confusing tbh

DMRobertson (approved):
Thanks! Let's try it

PR #652, opened 2023-09-22
Add empty request bodies to /forget calls in TestRoomForget

reivilibre:
These are meant to be required but it seems like they are not in practice...

c.f. https://github.com/matrix-org/synapse/issues/16366


DMRobertson (approved):

sumnerevans (approved):

reivilibre:
(ironically it was sending `Content-Type: application/json` even though they didn't have a request body! Ideally Synapse would not accept these, c.f. https://github.com/matrix-org/synapse/issues/16393)

Issue #526, opened 2022-10-20
Split up internal/client/client.go


ShadowJonathan:
The current file is 800 lines long, but contains a bunch of (mostly-independent) functions that are hard to sort through when scrolling down the file.

I think it'll be much easier to maintain if we split it into three files;
- client.go
- func.go (`MustDoFunc` etc.)
- sync.go

PR #658, opened 2023-10-03
Client refactor

### LINKED ISSUE #381: Add client.Event
People keep adding fields to `b.Event` when they aren't read in blueprints:
 - txn ID
 - prev/auth events
 - unsigned

This makes it very surprising from an API perspective to set these fields in a blueprint and have them do nothing. Instead, we should embed them:
```go
package client

type Event struct {
	b.Event
	Unsigned map[string]interface{}
	// The events needed to authenticate this event.
	// This can be either []EventReference for room v1/v2, or []string for room v3 onwards.
	// If it is left at nil, MustCreateEvent will populate it automatically based on the room state.
	AuthEvents interface{}
	// The prev events of the event if we want to override or falsify them.
	// If it is left at nil, MustCreateEvent will populate it automatically based on the forward extremities.
	PrevEvents interface{} 
...
```

And use `client.Event` in tests.

Ref: https://github.com/matrix-org/complement/pull/380/files

kegsay:
- Remove `Func` from `MustDoFunc` and `DoFunc`. They were called that to differentiate from the old-style `Do` and `MustDo` which are no more.
- Move sync functions to a new file.
- Move registration/login functions to a new file.
- Move `/internal/client` to `/client`.
- Signal unsafe functions.
- Add more helper functions where their usage is proven in tests.
- Use `TestLike`, an interface to represent `testing.T` to allow the client to be used outside of tests e.g benchmarks.

Slightly not client related:
 - Add `federation.Event` instead of abusing `b.Event` for creating events over federation. Makes this much clearer and fixes #381

Fixes https://github.com/matrix-org/complement/issues/526 , https://github.com/matrix-org/complement/issues/381 

Partly addresses https://github.com/matrix-org/complement/issues/226


client/client.go
```diff
@@ -0,0 +1,647 @@
...
+		queryParams.Add("after", after)
+	}
+
+	return c.MustDo(t, "PUT", []string{"_matrix", "client", "v3", "pushrules", scope, kind, ruleID}, WithJSONBody(t, body), WithQueries(queryParams))
+}
+
+// Unsafe_SendEventUnsynced sends `e` into the room. This function is UNSAFE as it does not wait
+// for the event to be fully processed. This can cause flakey tests. Prefer `SendEventSynced`.
+// Returns the event ID of the sent event.
+func (c *CSAPI) Unsafe_SendEventUnsynced(t TestLike, roomID string, e b.Event) string {
```
S7evinK, line 232:
```suggestion
func (c *CSAPI) UnsafeSendEventUnsynced(t TestLike, roomID string, e b.Event) string {
```
Maybe? (same for the others)

S7evinK (approved):
Aside from comments, LGTM

tests/csapi/sync_archive_test.go
```diff
@@ -306,7 +306,8 @@ func TestOlderLeftRoomsNotInLeaveSection(t *testing.T) {
 }
 
 // sytest: We should see our own leave event, even if history_visibility is
-//  restricted (SYN-662)
+//
+//	restricted (SYN-662)
```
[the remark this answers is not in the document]
kegsay: Yeah this is `go fmt` being overenthusiastic.

client/client.go
kegsay: I wanted to keep the `_` to make it very clear that it's UNSAFE_SomeFuncName`, rather than just the function happens to have "unsafe" in the name.

kegsay: It _should_ jump out at the reader as being dodgey is my point.

client/auth.go
```diff
@@ -0,0 +1,163 @@
...
+	res := c.MustDo(t, "POST", []string{"_matrix", "client", "v3", "login"}, WithJSONBody(t, reqBody))
+
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		t.Fatalf("unable to read response body: %v", err)
+	}
+
+	userID = gjson.GetBytes(body, "user_id").Str
+	accessToken = gjson.GetBytes(body, "access_token").Str
+	deviceID = gjson.GetBytes(body, "device_id").Str
```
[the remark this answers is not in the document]
kegsay: Absolutely.

client/client.go
```diff
@@ -0,0 +1,586 @@
...
+	defer res.Body.Close()
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		t.Fatalf("MustParseJSON: reading HTTP response body returned %s", err)
+	}
+	if !gjson.ValidBytes(body) {
+		t.Fatalf("MustParseJSON: Response is not valid JSON")
+	}
+	return body
+}
```
DMRobertson: It's odd for a function called `Parse` to return a raw byte slice; I'd've expected it to return a gjson.Result.

client/client.go
```diff
@@ -0,0 +1,586 @@
...
+
+// GjsonEscape escapes . and * from the input so it can be used with gjson.Get
+func GjsonEscape(in string) string {
+	in = strings.ReplaceAll(in, ".", `\.`)
+	in = strings.ReplaceAll(in, "*", `\*`)
+	return in
+}
```
DMRobertson, line 601: Is this actually needed? This gets used to escape user IDs, room IDs and event IDs. Are `.` and `*` legit characters for these IDs?

client/client.go
```diff
@@ -0,0 +1,586 @@
...
+
+func loopArray(object gjson.Result, key string, check func(gjson.Result) bool) error {
```
DMRobertson, line 603: Pants name. Maybe checkArrayElements?

client/auth.go
```diff
@@ -20,7 +18,7 @@ func WithDeviceID(deviceID string) LoginOpt {
 }
 
 // LoginUser will log in to a homeserver and create a new device on an existing user.
-func (c *CSAPI) LoginUser(t *testing.T, localpart, password string, opts ...LoginOpt) (userID, accessToken, deviceID string) {
+func (c *CSAPI) LoginUser(t TestLike, localpart, password string, opts ...LoginOpt) (userID, accessToken, deviceID string) {
```
DMRobertson, line 25: Does [testing.TB](https://pkg.go.dev/testing#TB) not suffice?

DMRobertson (commented):
(otherwise, no objections)

client/client.go
kegsay: They aren't typically seen in those ID types. It's also used to pull out JSON keys with `m.` in them which is much more frequently seen e.g `GjsonEscape("m.joined_member_count")`

client/client.go
kegsay: Copypastaed, but yes this is a better name. Will rename.

client/auth.go
kegsay: That has too many useless functions we don't use like `Setenv(key, value [string](https://pkg.go.dev/builtin#string))`.

client/client.go
kegsay: This isn't something I'm going to fix now. It is a problem though, as not all functions compose nicely. There are cases where you have:
 - `*http.Response`
 - `[]byte`
 - `io.Reader`
 - `gjson.Result`

There's reasons for this though:
 - Responses for when you care about headers and status codes.
 - `[]byte` for processing response bodies which may not be JSON.
 - `io.Reader` for when you need to re-consume the HTTP response body.
 - `gjson.Result` for when you're messing with JSON.

PR #659, opened 2023-10-04
Move /internal/b to /b

kegsay:
The API is unchanged and is suitable to be made public.


DMRobertson:
If it's going to be public can we call it `blueprints` rather than `b` or similar please?

S7evinK (approved):

kegsay:
No, because common usage would then be long and stutter: `blueprints.BlueprintOneToOneRoom`. Having `blueprints.OneToOneRoom` isn't much better, because the fundamental `type Blueprint struct` then becomes `blueprints.Blueprint`. `b` is short and unique, and doesn't hurt readability in tests imo.

kegsay:
dmr sez:

> I think part of the cause of https://github.com/matrix-org/complement/issues/381 is that it's not clear what b.Event means hence the suggestion, don't particularly care though

Issue #546, opened 2022-10-31
`should` counterpart to `must`


ShadowJonathan:
This came up while thinking of a solution for [this comment](https://github.com/matrix-org/complement/pull/535#discussion_r1009458884), I think it would help if we have a `should` package that mirrors the current `must` package, but omit the `*testing.T` input, and instead have an `error` output.

This'd allow to use the matching inside `client.SyncCheckOpt`, JSONCheckOff (and friends) and other such conditions, where tentative checking ("should match") needs to be done instead of harsh ("must match") checking.

Code between the modules dont need to be duplicated, as `must` code can become `should` code, with returning the error instead of `t.Fatalf`-ing it, and `must` only taking functions from `should` and aborting on failure when its non-nil.

kegsay:
This has come up a few times. See https://github.com/matrix-org/complement/pull/521#discussion_r997228161

The matchers themselves are doing the right thing and return `error` types. The problem is that they are typically paired with `must` functions, which as the package suggests, is fatal if there's a problem. It may be enough to just do as you describe:
 - remove t.Fatal/Error on existing `must` functions
 - rename package to `should`
 - add new `must` package which just proxies to `should` and fails if there is an error.

PR #663, opened 2023-10-05
Add should package

kegsay:
The `should` package is a more lenient form of the `must` package, in that it does not fail the test if the assertion fails. Instead, these functions return an error, which the test can then handle.

This is particularly useful in cases where a test needs to retry a certain request. Previously, you had to hand-roll the checks on the response because using `must.MatchResponse` would fail the test, even though it might have been okay as the next retry may pass the matchers.

Fixes https://github.com/matrix-org/complement/issues/546

Tests need to be revisited to see if this can be used in more places.

An example of this. Before:
```go
res = alice.DoFunc(
	t,
	"GET",
	[]string{"_matrix", "client", "v3", "rooms", roomID, "aliases"},
	client.WithRetryUntil(
		1*time.Second,
		func(res *http.Response) bool {
			if res.StatusCode != 200 {
				return false
			}
			eventResBody := client.ParseJSON(t, res)
			matcher := match.JSONKeyEqual("aliases", []interface{}{roomAlias})
			err := matcher(eventResBody)
			if err != nil {
				t.Log(err)
				return false
			}
			return true
		},
	),
)
```
After:
```go
res = alice.Do(
	t,
	"GET",
	[]string{"_matrix", "client", "v3", "rooms", roomID, "aliases"},
	client.WithRetryUntil(
		1*time.Second,
		func(res *http.Response) bool {
			_, err := should.MatchResponse(res, match.HTTPResponse{
				StatusCode: 200,
				JSON: []match.JSON{
					match.JSONKeyEqual("aliases", []interface{}{roomAlias}),
				},
			})
			if err != nil {
				t.Log(err)
				return false
			}
			return true
		},
	),
)
```

### DIFF
--- a/must/must.go
+++ b/must/must.go
@@ -1,29 +1,27 @@
-// Package must contains assertions for tests
+// Package must contains assertions for tests, which fail the test if the assertion fails.
 package must
 
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
 	"io"
 	"net/http"
 	"strings"
 	"testing"
 
 	"github.com/tidwall/gjson"
-	"golang.org/x/exp/slices"
 
 	"github.com/matrix-org/gomatrixserverlib/fclient"
 
-	"github.com/matrix-org/complement/client"
 	"github.com/matrix-org/complement/match"
+	"github.com/matrix-org/complement/should"
 )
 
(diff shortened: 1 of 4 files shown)


DMRobertson (approved):
Seems very sensible.

PR #664, opened 2023-10-06
Test content disposition against a content type with a parameter.

clokep:
See https://github.com/matrix-org/synapse/pull/16440

### DIFF
--- a/tests/media_filename_test.go
+++ b/tests/media_filename_test.go
@@ -137,6 +137,25 @@ func TestMediaFilenames(t *testing.T) {
 				}
 			})
 
+			t.Run("Will serve safe media types with parameters as inline", func(t *testing.T) {
+				if runtime.Homeserver != runtime.Synapse {
+					// We need to check that this security behaviour is being correctly run in
+					// Synapse, but since this is not part of the Matrix spec we do not assume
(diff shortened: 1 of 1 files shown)


tests/media_filename_test.go
```diff
@@ -137,6 +137,25 @@ func TestMediaFilenames(t *testing.T) {
 				}
 			})
 
+			t.Run("Will serve safe media types with parameters as inline", func(t *testing.T) {
+				if runtime.Homeserver != runtime.Synapse {
+					// We need to check that this security behaviour is being correctly run in
+					// Synapse, but since this is not part of the Matrix spec we do not assume
+					// other homeservers are doing so.
+					t.Skip("Skipping test of Content-Disposition header requirements on non-Synapse homeserver")
```
kegsay, line 145: So I guess the reason you're not using `runtime.SkipIf` is because you want to _only_ run this test on synapse, rather than skip if it is dendrite/conduit?

kegsay (approved):

tests/media_filename_test.go
clokep: It is what the other tests in this file used and I copied & pasted the test to bootstrap. 😄

PR #666, opened 2023-10-10
Remove msc build tags. Add msc packages.

kegsay:
This adds more parallelism as a side-effect.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,7 +48,7 @@ jobs:
 
           - homeserver: Dendrite
             tags: dendrite_blacklist
-            packages: ./tests/msc2836
+            packages: ""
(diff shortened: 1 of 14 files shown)


DMRobertson (commented):
Is this going to need changes to https://github.com/matrix-org/synapse/blob/6d81aec09febe86532235141e84c4ea0b3f56049/scripts-dev/complement.sh#L217? Are there other places in Synapse we'll need to change? Similarly for dendrite

kegsay:
Yes.

DMRobertson (approved):
LGTM but it'd be very handy if you can do the Synapse+Dendrite changes too. (Though FWIW Synapse's develop is failing on particular complement test RN)

devonh (approved):

PR #669, opened 2023-10-12
Standardise entry points for Complement tests

kegsay:
Every test package that wants to use Complement needs some bootstrapping boilerplate code to do things like read the env vars and clean up old containers in case of an ungraceful shutdown. There is a convenient `TestMain` function which is part of standard Go tests which we can use for this.

We also had a helper function called `Deploy` which would deploy a set of containers and return the `docker.Deployment` for tests to use. This helper function is actually boilerplate as well, but has still historically been defined for each test package unnecessarily.

This PR cleans up these entry points in the following way:
 - Always use the helper functions when writing entry points. Previously, only MSC packages did this. This was just an omission from #666 
 - Moves entry point functions from the `helpers` package to the top-level `complement` package. Why: this signals to test writers where to look when they want to hook into Complement machinery as it serves as the entry point functions. This ends up reading very nicely as `deployment := complement.Deploy(t, b.BlueprintAlice)`.
 - Remove the useless proxy function `Deploy(...)` which just called the previous line.

This now means that out-of-repo tests (which need to call `somepackage.Deploy`) and in-repo tests look identical, and bootstrapping boilerplate is as small as possible:
```
package tests

import (
	"testing"

	"github.com/matrix-org/complement"
)

func TestMain(m *testing.M) {
	complement.TestMain(m, "package-name-goes-here")
}
```

### DIFF
--- a/test_main.go
+++ b/test_main.go
@@ -1,4 +1,4 @@
-package helpers
+package complement
 
 import (
 	"fmt"
--- a/test_package.go
+++ b/test_package.go
@@ -1,4 +1,4 @@
-package helpers
+package complement
 
 import (
 	"context"
--- a/tests/csapi/account_change_password_pushers_test.go
+++ b/tests/csapi/account_change_password_pushers_test.go
@@ -6,8 +6,9 @@ package csapi_tests
 import (
 	"testing"
 
-	"github.com/matrix-org/complement/client"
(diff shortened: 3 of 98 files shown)


DMRobertson (approved):
Thanks for the clean commit history. Seems sane to me.

PR #672, opened 2023-10-12
Add a complement.Deployment interface

kegsay:
The `docker` package is internal, meaning that it's hard for out-of-repo tests to make use of the `*docker.Deployment`. Instead, hide the concrete impl behind an interface `complement.Deployment`, which also conveniently only exposes the functions used by real tests. Pepper this API with TODOs. Add a `FederationDeployment` which is a subset of `complement.Deployment` required by the complement-fed-server.

This is the first step to enabling dirty runs. Now that the deployment API is defined, this opens up the ability to:
 - specify a general "give me a CSAPI" function with functional options for passwords/etc, which will replace `NewUser`, `LoginUser`, `RegisterUser`, `Client`). This function will _not allow the full user ID to be specified_, so dirty runs can add a prefix/suffix to the localpart.
 - add `complement.DirtyDeploy` which can have the newly desired function signature of `deployment := complement.Deploy(t, numHomeservers)`. Swap some tests over to this to get a feel for what works/doesn't work. This means tests _will not be able to specify blueprints_. This is fine because the new "give me a CSAPI" function will handle the typical `alice := deployment.Client(t, "hs1", "@alice:hs1")` calls.
 - replace `complement.Deploy` with `complement.DirtyDeploy` (and remove `Dirty`), updating all tests to the new format.

At that point, containers are still made once for each test, but all tests are now ignorant of the user ID / blueprint, meaning we can then swap out the concrete `Deployment` impl for one which re-uses containers, and add an env var to control it e.g `COMPLEMENT_DIRTY=1`.

Long term, I want dirty runs to be enabled _by default_, to encourage test writers to write isolated tests which don't pollute other tests.

### DIFF
--- a/internal/federation/server.go
+++ b/internal/federation/server.go
@@ -30,8 +30,13 @@ import (
 	"github.com/matrix-org/util"
 
 	"github.com/matrix-org/complement/internal/config"
-	"github.com/matrix-org/complement/internal/docker"
 )
+
+// Subset of Deployment used in federation
+type FederationDeployment interface {
+	GetConfig() *config.Complement
+	RoundTripper() http.RoundTripper
+}
 
 // Server represents a federation server
 type Server struct {
@@ -57,7 +62,7 @@ type Server struct {
 }
 
 // NewServer creates a new federation server with configured options.
-func NewServer(t *testing.T, deployment *docker.Deployment, opts ...func(*Server)) *Server {
+func NewServer(t *testing.T, deployment FederationDeployment, opts ...func(*Server)) *Server {
 	// generate signing key
 	pub, priv, err := ed25519.GenerateKey(nil)
 	if err != nil {
@@ -71,7 +76,7 @@ func NewServer(t *testing.T, deployment *docker.Deployment, opts ...func(*Server
 		mux:   mux.NewRouter(),
 		// The server name will be updated when the caller calls Listen() to include the port number
 		// of the HTTP server e.g "host.docker.internal:56353"
-		serverName:                  deployment.Config.HostnameRunningComplement,
+		serverName:                  deployment.GetConfig().HostnameRunningComplement,
 		rooms:                       make(map[string]*ServerRoom),
 		aliases:                     make(map[string]string),
 		UnexpectedRequestsAreErrors: true,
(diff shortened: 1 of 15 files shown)


internal/federation/server.go
```diff
@@ -30,9 +30,14 @@ import (
...
 
 	"github.com/matrix-org/complement/internal/config"
-	"github.com/matrix-org/complement/internal/docker"
 )
 
+// Subset of Deployment used in federation
+type FederationDeployment interface {
+	GetConfig() *config.Complement
+	RoundTripper() http.RoundTripper
+}
```
DMRobertson, line 39: So a FederationDeployment is a Deployment, but not the other way round?

I'm not sure that makes much sense, since one could get two regular Deployments to federate with each other?

Is the point that internal/federation represents a complement-controlled mock homeserver?

DMRobertson: generally speaking if I have a universe of objects Foo and some adjective or noun X, I would expect an XFoo to be a more specific interface rather than less specific. (Unless X is `Base` or `Abstract`, I suppose.)

DMRobertson (commented):
Switching out the interfaces seems sane, even if I don't really follow what the interfaces _represent_. Might want a second opinion here though.

internal/federation/server.go
kegsay: `Deployment` is a collection of running homeservers which are federating with each other. There can just be 1 homeserver in which case it isn't federating. It exposes a bunch of functions for interacting with these servers.

`FederationDeployment` is a subset of `Deployment` functions which are only relevant for the internal complement server. This is mostly out of neccessity rather than anything else, as if I passed in a `Deployment` I'd get a circular reference loop. In essence, `FederationDeployment` is just the bits of `Deployment` this file needs to work correctly.

test_package.go
```diff
@@ -4,16 +4,44 @@ import (
...
 	"time"
 
 	"github.com/matrix-org/complement/b"
+	"github.com/matrix-org/complement/client"
 	"github.com/matrix-org/complement/internal/config"
 	"github.com/matrix-org/complement/internal/docker"
 	"github.com/sirupsen/logrus"
 )
 
+type Deployment interface {
```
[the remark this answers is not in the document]
kegsay: homeserver_s_ but yes. Will tweak.

PR #673, opened 2023-10-13
Refactor Deployment Register/Login functions

kegsay:
Define just 2 functions which can be used universally. By having this as an interface, this also means we keep the door open for matrix-authentication-service to swap out the impl for one which registers with MAS.

The `Login` function requires an _existing_ CSAPI client to work. This is by design to force tests to not hardcode "login as localpart `alice`".

The `Register` function does not do this, and exposes a literal `Localpart` field to set. This will change to a prefix/suffix, but is more involved e.g a test may call `Register("chris")` then later on assert `@chris:hs1` is present somewhere, so all tests need to be sanity checked for this before swapping to prefixes/suffixes). Hence this will be up next in another PR.

### DIFF
--- a/helpers/clientopts.go
+++ b/helpers/clientopts.go
@@ -1,0 +1,12 @@
+package helpers
+
+type RegistrationOpts struct {
+	Localpart string // default '' (don't care)
+	DeviceID  string // default '' (generate new)
+	Password  string // default 'complement_meets_min_password_requirement'
+	IsAdmin   bool   // default false
+}
+
+type LoginOpts struct {
+	Password string // default 'complement_meets_min_password_requirement'
+}
--- a/internal/docker/deployment.go
+++ b/internal/docker/deployment.go
@@ -85,7 +85,10 @@ func (d *Deployment) Register(t *testing.T, hsName string, opts helpers.Registra
 	if password == "" {
 		password = "complement_meets_min_password_req"
 	}
-	localpart := fmt.Sprintf("user-%v-%v", d.localpartCounter.Add(1), opts.Localpart)
+	localpart := opts.Localpart
+	if localpart == "" {
+		localpart = fmt.Sprintf("user-%v", d.localpartCounter.Add(1))
+	}
 	var userID, accessToken, deviceID string
 	if opts.IsAdmin {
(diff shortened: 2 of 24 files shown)


DMRobertson:
> By having this as an interface, this also means we keep the door open for matrix-authentication-service to swap out the impl for one which registers with MAS.

Nice!

helpers/clientopts.go
```diff
@@ -0,0 +1,12 @@
+package helpers
+
+type RegistrationOpts struct {
+	Localpart string // default '' (don't care)
+	DeviceID  string // default '' (generate new)
+	Password  string // default 'complement_meets_min_password_requirement'
+	IsAdmin   bool   // default false
```
DMRobertson, line 7: Aside: do any of the tests actually use the adminness? It's not a specced property of Matrix as much as a backdoor for Synapse, I thought?

internal/docker/deployment.go
```diff
@@ -60,31 +65,13 @@ func (d *Deployment) RoundTripper() http.RoundTripper {
 	return &RoundTripper{Deployment: d}
 }
 
-// Client returns a CSAPI client targeting the given hsName, using the access token for the given userID.
-// Fails the test if the hsName is not found. Returns an unauthenticated client if userID is "", fails the test
-// if the userID is otherwise not found.
-func (d *Deployment) Client(t *testing.T, hsName, userID string) *client.CSAPI {
-	t.Helper()
+func (d *Deployment) Register(t *testing.T, hsName string, opts helpers.RegistrationOpts) *client.CSAPI {
```
DMRobertson, line 68: When it comes to writing individual tests, I'd like to be able to still do something like `RegisterNamed(t, "alice") like we do in SS. But I guess that can just be a wrapper on top of this interface?

DMRobertson (approved):

helpers/clientopts.go
kegsay: Yes tests use it. It _is_ a specced property in Matrix all over the place, just no means to make one via the login/register APIs. See https://spec.matrix.org/latest/client-server-api/#get_matrixclientv3adminwhoisuserid for the most obvious example.

internal/docker/deployment.go
kegsay: Yes, see next PR.

PR #674, opened 2023-10-13
Replace Localpart with LocalpartSuffix 

kegsay:
Based on https://github.com/matrix-org/complement/pull/673

This means clients can't register specific localparts, which is good for dirty runs.

TODO: Delete `Deployment.Client` and swap to `Deployment.Register` for standard alice/bob runs.

### DIFF
--- a/helpers/clientopts.go
+++ b/helpers/clientopts.go
@@ -1,10 +1,10 @@
 package helpers
 
 type RegistrationOpts struct {
-	Localpart string // default '' (don't care)
-	DeviceID  string // default '' (generate new)
-	Password  string // default 'complement_meets_min_password_requirement'
-	IsAdmin   bool   // default false
+	LocalpartSuffix string // default '' (don't care)
+	DeviceID        string // default '' (generate new)
+	Password        string // default 'complement_meets_min_password_requirement'
+	IsAdmin         bool   // default false
 }
 
 type LoginOpts struct {
--- a/internal/docker/deployment.go
+++ b/internal/docker/deployment.go
@@ -85,10 +85,7 @@ func (d *Deployment) Register(t *testing.T, hsName string, opts helpers.Registra
 	if password == "" {
 		password = "complement_meets_min_password_req"
 	}
-	localpart := opts.Localpart
-	if localpart == "" {
-		localpart = fmt.Sprintf("user-%v", d.localpartCounter.Add(1))
-	}
+	localpart := fmt.Sprintf("user-%v-%v", d.localpartCounter.Add(1), opts.LocalpartSuffix)
 	var userID, accessToken, deviceID string
 	if opts.IsAdmin {
 		userID, accessToken, deviceID = client.RegisterSharedSecret(t, localpart, password, opts.IsAdmin)
--- a/tests/csapi/device_lists_test.go
+++ b/tests/csapi/device_lists_test.go
@@ -2,7 +2,6 @@ package csapi_tests
 
 import (
 	"fmt"
-	"sync/atomic"
 	"testing"
 
 	"github.com/matrix-org/complement"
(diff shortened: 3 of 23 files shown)


helpers/clientopts.go
```diff
@@ -1,10 +1,10 @@
 package helpers
 
 type RegistrationOpts struct {
-	Localpart string // default '' (don't care)
-	DeviceID  string // default '' (generate new)
-	Password  string // default 'complement_meets_min_password_requirement'
-	IsAdmin   bool   // default false
+	LocalpartSuffix string // default '' (don't care)
```
DMRobertson, line 4: Can we make this a prefix so that I can grep for `@alice` instead of `@[^:]+alice`?

DMRobertson: c.f. https://github.com/matrix-org/sliding-sync/blob/57d8c1deba918597f558b013cbc94432ed14d649/tests-e2e/main_test.go#L214

internal/docker/deployment.go
```diff
@@ -85,10 +85,7 @@ func (d *Deployment) Register(t *testing.T, hsName string, opts helpers.Registra
 	if password == "" {
 		password = "complement_meets_min_password_req"
 	}
-	localpart := opts.Localpart
-	if localpart == "" {
-		localpart = fmt.Sprintf("user-%v", d.localpartCounter.Add(1))
-	}
+	localpart := fmt.Sprintf("user-%v-%v", d.localpartCounter.Add(1), opts.LocalpartSuffix)
```
DMRobertson, line 88: If there's no suffix, the localpart is gonna end with `-` which feels a bit weird. Though that is completely fine.

tests/csapi/device_lists_test.go
```diff
@@ -21,13 +20,6 @@ import (
 //  1. `/sync`'s `device_lists.changed/left` contain the correct user IDs.
 //  2. `/keys/query` returns the correct information after device list updates.
 func TestDeviceListUpdates(t *testing.T) {
-	var localpartIndex int64 = 0
-	// generateLocalpart generates a unique localpart based on the given name.
-	generateLocalpart := func(localpart string) string {
-		index := atomic.AddInt64(&localpartIndex, 1)
-		return fmt.Sprintf("%s%d", localpart, index)
-	}
-
```
DMRobertson, line 30: This feels like a win!

tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -376,7 +376,7 @@ func TestPartialStateJoin(t *testing.T) {
 	// resync completes. This test does exactly that.
 	t.Run("EagerLongPollingSyncWokenWhenResyncCompletes", func(t *testing.T) {
 		alice := deployment.Register(t, "hs1", helpers.RegistrationOpts{
-			Localpart: "t1alice_long_polling",
+			LocalpartSuffix: "t1alice_long_polling",
```
DMRobertson, line 379: Aside: can you remember why we ended up making all these different users in the faster joins tests?

tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -4080,8 +4080,8 @@ func TestPartialStateJoin(t *testing.T) {
 		}
 		t.Log("Alice begins a partial join to a room")
 		alice := deployment.Register(t, "hs1", helpers.RegistrationOpts{
-			Localpart: "t46alice",
-			IsAdmin:   true,
+			LocalpartSuffix: "t46alice",
+			IsAdmin:         true,
```
DMRobertson, line 4084: Would be very surprised if alice actually needs Admin here, similar for terry above. Te `t.Log()` makes it look like I wrote this, so maybe I was completely off my rocker here.

DMRobertson (changes requested):
I think a prefix would be friendlier, but I have no other objections

helpers/clientopts.go
kegsay: I guess. It's just that there are restrictions on valid characters for `localpart[0]` so it adds arbitrary restrictions on test writers.

internal/docker/deployment.go
kegsay: Yeah I know, but shrug?

tests/msc3902/federation_room_join_partial_state_test.go
kegsay: No clue. Probably because they were parallel at some point, using the same HS deployment?

tests/msc3902/federation_room_join_partial_state_test.go
kegsay: Agreed, this surprised me but also I wasn't about to touch it.

tests/msc3902/federation_room_join_partial_state_test.go
DMRobertson: Sounds plausible. I vaguely remember that we added them to make things less flakey.

DMRobertson (approved):

PR #676, opened 2023-10-13
Remove Deployment.Client and change Deploy

kegsay:
- `Deployment.Client` was used to get pre-registered clients. Now we want tests to register new users for each test, for dirty runs. So swap for `Deployment.Register` everywhere.
- `Deploy` was used to deploy a blueprint. We don't want this to enable dirty runs. So replace it with the number of servers you need e.g `Deploy(t, 2)`.

### DIFF
--- a/internal/docker/deployment.go
+++ b/internal/docker/deployment.go
@@ -36,8 +36,10 @@ type HomeserverDeployment struct {
 	accessTokensMutex   sync.RWMutex
 	ApplicationServices map[string]string // e.g { "my-as-id": "id: xxx\nas_token: xxx ..."} }
 	DeviceIDs           map[string]string // e.g { "@alice:hs1": "myDeviceID" }
-	CSAPIClients        []*client.CSAPI
-	CSAPIClientsMutex   sync.Mutex
+
+	// track all clients so if Restart() is called we can repoint to the new high-numbered port
+	CSAPIClients      []*client.CSAPI
+	CSAPIClientsMutex sync.Mutex
 }
 
 // Updates the client and federation base URLs of the homeserver deployment.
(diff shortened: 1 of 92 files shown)


test_package.go
```diff
@@ -82,9 +83,30 @@ func (tp *TestPackage) Cleanup() {
...
+	timeStartDeploy := time.Now()
+	dep, err := d.Deploy(context.Background(), blueprint.Name)
+	if err != nil {
+		t.Fatalf("OldDeploy: Deploy returned error %s", err)
+	}
+	t.Logf("OldDeploy times: %v blueprints, %v containers", timeStartDeploy.Sub(timeStartBlueprint), time.Since(timeStartDeploy))
+	return dep
+}
+
+func (tp *TestPackage) Deploy(t *testing.T, numServers int) Deployment {
```
sandhose, line 106: If I got that correctly, the new Deploy only cares about spawning N servers, and does not preconfigure users on it?

internal/docker/deployment.go
```diff
@@ -36,8 +36,10 @@ type HomeserverDeployment struct {
 	accessTokensMutex   sync.RWMutex
 	ApplicationServices map[string]string // e.g { "my-as-id": "id: xxx\nas_token: xxx ..."} }
 	DeviceIDs           map[string]string // e.g { "@alice:hs1": "myDeviceID" }
-	CSAPIClients        []*client.CSAPI
-	CSAPIClientsMutex   sync.Mutex
+
+	// track all clients so if Restart() is called we can repoint to the new high-numbered port
```
sandhose, line 40: I'm failing to understand this comment, but maybe I'll get it once I'm further in the review

test_package.go
```diff
@@ -19,15 +19,16 @@ import (
...
 // Deployment provides a way for tests to interact with a set of homeservers.
 type Deployment interface {
-	// Client returns a CSAPI client targeting the given hsName, using the access token for the given userID.
-	// Fails the test if the hsName is not found. Returns an unauthenticated client if userID is "", fails the test
-	// if the userID is otherwise not found.
-	Client(t *testing.T, serverName, userID string) *client.CSAPI
+	// UnauthenticatedClient returns a blank CSAPI client.
+	UnauthenticatedClient(t *testing.T, serverName string) *client.CSAPI
 	// Register a new user on the given server.
 	Register(t *testing.T, hsName string, opts helpers.RegistrationOpts) *client.CSAPI
```
sandhose, line 25: I know it's not part of this PR, but looking at the usage, why keep the `hsName` here, instead of using an index?

Also, since most of the times, the registration options are not set, can't we have an API with variadic args? So that `d.Register(t, "hs1")` works, and you can still do `d.Register(t, "hs1", WithDeviceID("foo"), WithLocalPartSuffix("suffix"))`

sandhose (approved):
As someone who was very confused when writing complements tests why blueprints were a thing & why some test would use the builtin users and some not, this makes me happy

I've left a few comments here and there, but overall LGTM

test_package.go
kegsay: Correct.

test_package.go
kegsay: I chose to keep `hsName` mostly to try to keep things not too dissimilar to the previous API shape. It'll likely change again in another PR.

Agreed variadic functional args would be clearer, I'll do it on another PR.

PR #679, opened 2023-10-17
Fix param name in (Must)CreateRoom

DMRobertson:
the given map is the JSON body of the /createRoom request, NOT the "creation content" that the spec speaks of (i.e. the content of the m.room.create event.)


clokep (approved):
I think the synapse failure is a flake?

PR #681, opened 2023-10-18
Add regression test for matrix-org/synapse#16463

erikjohnston:
c.f.  https://github.com/matrix-org/synapse/issues/16463 and https://github.com/matrix-org/synapse/pull/16485

### DIFF
--- a/tests/csapi/sync_test.go
+++ b/tests/csapi/sync_test.go
@@ -507,11 +507,11 @@ func TestSyncTimelineGap(t *testing.T) {
 			events := section.Get("events").Array()
 			for _, ev := range events {
 				if limited {
-					if ev.Get("Sender").Str == alice.UserID {
+					if ev.Get("sender").Str == alice.UserID {
 						t.Fatalf("Got message from alice after limited flag")
 					}
 				} else {
-					if ev.Get("Sender").Str == charlie {
+					if ev.Get("sender").Str == charlie {
 						t.Fatalf("Got message from remote without limited flag being set")
 					}
 				}
--- a/tests/csapi/sync_test.go
+++ b/tests/csapi/sync_test.go
@@ -384,7 +384,7 @@ func TestSync(t *testing.T) {
 // room. Charlie sends 50+ messages into the room without sending to Alice's
 // server. Charlie then sends one more which get sent to Alice.
 //
-// Alice should observer that she receives some (though not all) of charlie's
+// Alice should observe that she receives some (though not all) of charlie's
 // events, with the `limited` flag set.
 func TestSyncTimelineGap(t *testing.T) {
 	runtime.SkipIf(t, runtime.Dendrite)
@@ -458,10 +458,12 @@ func TestSyncTimelineGap(t *testing.T) {
 
 	// We now test two different modes of /sync work. The first is when we are
 	// syncing when the server receives the `lastEvent` (and so, at least
-	// Synapse, will start sending down some events immediately).
+	// Synapse, will start sending down some events immediately). In this mode
+	// we may see alice's message, but charlie's messages should set the limited
+	// flag.
 	//
-	// The second mode is when we sync *after* all the events have finished
-	// being persisted, and so we get everything in one chunk.
+	// The second mode is when we incremental sync *after* all the events have
+	// finished being persisted, and so we get only charlie's messages.
 	t.Run("incremental", func(t *testing.T) {
 		timelineSequence := make([]gjson.Result, 0)
 
@@ -524,7 +526,8 @@ func TestSyncTimelineGap(t *testing.T) {
 	})
 
 	t.Run("full", func(t *testing.T) {
-		// Wait until we see `lastEvent` come down sync
+		// Wait until we see `lastEvent` come down sync implying that all events have been persisted
+		// by alice's homeserver.
 		alice.MustSyncUntil(t, client.SyncReq{}, client.SyncTimelineHasEventID(roomID, lastEvent.EventID()))
 
 		// Now an incremental sync from before should return a limited batch for
--- a/tests/csapi/sync_test.go
+++ b/tests/csapi/sync_test.go
@@ -387,7 +387,6 @@ func TestSync(t *testing.T) {
 // Alice should observe that she receives some (though not all) of charlie's
 // events, with the `limited` flag set.
 func TestSyncTimelineGap(t *testing.T) {
-	runtime.SkipIf(t, runtime.Dendrite)
 	deployment := complement.Deploy(t, 1)
 	defer deployment.Destroy(t)
 	alice := deployment.Register(t, "hs1", helpers.RegistrationOpts{})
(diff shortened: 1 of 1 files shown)


tests/csapi/sync_test.go
```diff
@@ -374,6 +377,188 @@ func TestSync(t *testing.T) {
...
+				return fmt.Errorf("Did not find lastEvent (%s) in timeline batch: (%s)", lastEvent.EventID(), roomResult.Raw)
+			}
+
+			return nil
+		})
+
+		t.Logf("Got timeline sequence: %s", timelineSequence)
+
+		// Check that we only see Alice's message from before the gap *before*
+		// we seen any limited batches, and vice versa for Charlie's messages.
```
clokep, line 503: I'm not following this comment. Is "seen" supposed to be "see"?

tests/csapi/sync_test.go
```diff
@@ -374,6 +377,188 @@ func TestSync(t *testing.T) {
...
+	// Synapse, will start sending down some events immediately).
+	//
+	// The second mode is when we sync *after* all the events have finished
+	// being persisted, and so we get everything in one chunk.
+	t.Run("incremental", func(t *testing.T) {
+		timelineSequence := make([]gjson.Result, 0)
+
+		t.Logf("Doing incremental syncs from %s", nextBatch)
+
+		// This just reads all timeline batches into `timelineSequence` until we see `lastEvent` come down
```
clokep, line 470: So `timelineSequence` is an array of arrays? We're not flattening into it, correct?

tests/csapi/sync_test.go
```diff
@@ -374,6 +377,188 @@ func TestSync(t *testing.T) {
...
+		}
+
+		if !limited {
+			t.Fatalf("No timeline batch for the room was limited")
+		}
+	})
+
+	t.Run("full", func(t *testing.T) {
+		// Wait until we see `lastEvent` come down sync
+		alice.MustSyncUntil(t, client.SyncReq{}, client.SyncTimelineHasEventID(roomID, lastEvent.EventID()))
```
clokep, line 528:
```suggestion
		// Wait until we see `lastEvent` come down sync implying that all events have been persisted
		// by alice's homeserver.
		alice.MustSyncUntil(t, client.SyncReq{}, client.SyncTimelineHasEventID(roomID, lastEvent.EventID()))
```

tests/csapi/sync_test.go
```diff
@@ -561,3 +746,105 @@ func usersInPresenceEvents(t *testing.T, presence gjson.Result, users []string)
...
+func eventIDsFromEvents(he []gomatrixserverlib.PDU) []string {
+	eventIDs := make([]string, len(he))
+	for i := range he {
+		eventIDs[i] = he[i].EventID()
+	}
+	return eventIDs
+}
+
+// Helper method to respond to federation APIs associated with trying to get missing events.
+func respondToGetMissingEventsEndpoints(t *testing.T, srv *federation.Server, room *federation.ServerRoom, missingEvents []gomatrixserverlib.PDU) {
```
clokep, line 759: This feels boilerplate-y, do we do this in other tests a lot?

tests/csapi/sync_test.go
```diff
@@ -374,6 +377,188 @@ func TestSync(t *testing.T) {
...
+	respondToGetMissingEventsEndpoints(t, srv, room, missingEvents)
+
+	srv.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{lastEvent.JSON()}, nil)
+
+	// We now test two different modes of /sync work. The first is when we are
+	// syncing when the server receives the `lastEvent` (and so, at least
+	// Synapse, will start sending down some events immediately).
+	//
+	// The second mode is when we sync *after* all the events have finished
+	// being persisted, and so we get everything in one chunk.
```
clokep, line 464: I'm only kind of understanding the difference here; incremental will have *both* alice and charlie's events in the response, but will be limited? While after full it will be just charlie's?

tests/csapi/sync_test.go
erikjohnston: Yes, it is mean to be "see". There are two valid scenarios: 1) we sync and see Alice's message, then we sync again and see charlie's messages with `limited` set to true, or 2) we just see charlie's messages with limited set to true

tests/csapi/sync_test.go
erikjohnston: It's an array of the timeline chunks, i.e. `[{"events": [..], "limited": false}, ...]`

tests/csapi/sync_test.go
erikjohnston: We don't do it very often, and each is fairly unique AFAICT

tests/csapi/sync_test.go
erikjohnston: Yes indeed. It's perfectly valid when actively syncing for us to see Alice's message, *then* a limited batch with one or more of charlie's messages. However, we also want to check the case of you syncing only after all the messages come in

clokep (approved):
I think this is OK, modulo it being really confusing and the comments could possibly be expanded a bit more. I made some suggestions, but not sure they're helpful.

tests/csapi/sync_test.go
```diff
@@ -374,6 +377,188 @@ func TestSync(t *testing.T) {
...
+// https://github.com/matrix-org/synapse/issues/16463
+//
+// We test this by having a local user (alice) and remote user (charlie) in a
+// room. Charlie sends 50+ messages into the room without sending to Alice's
+// server. Charlie then sends one more which get sent to Alice.
+//
+// Alice should observer that she receives some (though not all) of charlie's
+// events, with the `limited` flag set.
+func TestSyncTimelineGap(t *testing.T) {
+	runtime.SkipIf(t, runtime.Dendrite)
```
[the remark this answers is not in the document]
erikjohnston: Err, good Q. I can try removing it and seeing what happens

tests/csapi/sync_test.go
```diff
@@ -374,6 +377,191 @@ func TestSync(t *testing.T) {
...
+// https://github.com/matrix-org/synapse/issues/16463
+//
+// We test this by having a local user (alice) and remote user (charlie) in a
+// room. Charlie sends 50+ messages into the room without sending to Alice's
+// server. Charlie then sends one more which get sent to Alice.
+//
+// Alice should observe that she receives some (though not all) of charlie's
+// events, with the `limited` flag set.
+func TestSyncTimelineGap(t *testing.T) {
+	runtime.SkipIf(t, runtime.Dendrite)
```
erikjohnston, line 390:
```suggestion
```

tests/csapi/sync_test.go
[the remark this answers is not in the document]
clokep: Looks like it fails! Which brings me back to a question I asked somewhere else -- does this require spec clarifications?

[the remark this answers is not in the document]
erikjohnston: Yup, looks like it. The current spec says this about `limited`:

> True if the number of events returned was limited by the limit on the filter.

PR #690, opened 2023-11-07
WIP: Add regression test

kegsay:


DMRobertson:
To spell this out, the spec [says](https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientv3sync) that `state` is

> Updates to the state, between the time indicated by the since parameter, and the start of the timeline (or all state up to the start of the timeline, if since is not given, or full_state is true).

This means that the `state` block is a subset of the room state before the first timeline entry.

[On `m.room.create`:](https://spec.matrix.org/v1.8/client-server-api/#mroomcreate)

> This is the first event in a room and cannot be changed. It acts as the root of all other events.

This means that the state before the create event is empty.

The invariant to enforce is thus

> If m.room.create appears in a `timeline` block, it must be the first entry in the timeline and there must be no `state` block.

clokep:
Clearing review flags since this is a draft? Shout if that was the wrong thing to do and you wanted some sort of initial review.

PR #694, opened 2023-11-23
Add to-device event tests over federation; incl. connectivity tests

kegsay:
The connectivity tests form part of complement-crypto's test suite, specifically:

> If a server cannot send device list updates over federation, it retries.

### DIFF
--- a/tests/federation_to_device_test.go
+++ b/tests/federation_to_device_test.go
@@ -1,0 +1,95 @@
+package tests
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/matrix-org/complement"
+	"github.com/matrix-org/complement/client"
+	"github.com/matrix-org/complement/helpers"
+	"github.com/tidwall/gjson"
+)
+
+// Test that to-device messages can go from one homeserver to another.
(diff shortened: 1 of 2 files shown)


S7evinK (approved):

tests/federation_to_device_test.go
```diff
@@ -0,0 +1,127 @@
...
+				return reflect.DeepEqual(evContent, content)
+			}
+			// just in case the server returns 200 OK before flushing to disk, give it a grace period.
+			// This is too nice of us given in the real world no grace is provided..
+			time.Sleep(time.Second)
+
+			tc.makeReachable(t)
+
+			// servers may need to be poked with another to-device msg. This isn't great.
+			// See https://github.com/matrix-org/synapse/issues/16680
```
S7evinK, line 105:
```suggestion
			// servers may need to be poked with another to-device msg. This isn't great.
			// See https://github.com/matrix-org/synapse/issues/16680
			// bob has a sync timeout of 30s set, so if the test has not yet passed, we are kicking the server
			// after 10s to.
```

PR #699, opened 2024-01-05
Add test to check timeline ordering semantics

kegsay:
Based on random conversations, we want to make sure that clients cannot accidentally miss being told about events if an earlier event (earlier in the DAG) arrives interleaved with newer events.

### DIFF
--- a/tests/federation_room_send_test.go
+++ b/tests/federation_room_send_test.go
@@ -1,15 +1,23 @@
 package tests
 
 import (
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"slices"
 	"testing"
 	"time"
 
 	"github.com/matrix-org/complement"
 	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/tidwall/gjson"
 
 	"github.com/matrix-org/complement/b"
-	"github.com/matrix-org/complement/helpers"
+	"github.com/matrix-org/complement/client"
 	"github.com/matrix-org/complement/federation"
+	"github.com/matrix-org/complement/helpers"
+	"github.com/matrix-org/complement/match"
+	"github.com/matrix-org/complement/must"
 )
 
 // TODO:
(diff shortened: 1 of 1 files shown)


tests/federation_room_send_test.go
```diff
@@ -68,3 +77,146 @@ func TestOutboundFederationSend(t *testing.T) {
...
+// Test ordering behaviour between /sync and /messages, when unexpected earlier federation events are injected.
+// This aims to ensure that clients cannot miss events in the following case:
+//
+//	consider two homeservers in the same room starting at some event 0. Then:
+//	  - network partition the servers
+//	  - HS1 sends 1, HS2 sends 1'
+//	  - a local user on HS1 sends 2,3,4
+//	  - network partition is fixed. HS2 sends 1' to HS1.
+//	  - a local user on HS1 sends 5,6,7
+//	  - At this point, from HS1's pov, the streaming ordering is [1,2,3,4,1',5,6,7] and the topological order is [1 | 1', 2,3,4,5,6,7]
```
erikjohnston, line 90:
```suggestion
//	  - At this point, from HS1's pov, the streaming ordering is [1,2,3,4,1',5,6,7] and the topological order is [1 | 1', 2,3,4,5,6,7]
```

That is *Synapse's* choice of topological ordering. Valid ones are basically `[(1, 2, 3, 4) | 1', 5, 6, 7]` (i.e. `1'` could appear anywhere before/after/between `1, 2, 3, 4`).

tests/federation_room_send_test.go
```diff
@@ -68,3 +77,146 @@ func TestOutboundFederationSend(t *testing.T) {
...
+//	  - client requests timeline limit = 4 => [1',5,6,7] is returned because /sync is stream ordering. Everything is fine so far.
+//	  - using the prev_batch token here in /messages SHOULD return 4,3,2,1, to ensure clients cannot missing 4,3,2.
```
erikjohnston, line 92: And potentially `1'`, depending, e.g. if  you do `/messages?limit=1` repeatedly `1'` would turn up again currently.

erikjohnston (approved):

erikjohnston:
As requested, I'm going to quickly lay out how Synapse implements this currently.

There are two orderings Synapse uses, called "stream ordering" and "topological ordering" (which is a specific topological ordering). Once an event is persisted these orderings do not change.

There are also two types of tokens: stream tokens and topological tokens. These correspond with segmenting events based on the stream ordering and the topological ordering respectively.

The two requests that happen are:
1. `/sync` which returns events in stream ordering and includes stream tokens for the start and end of the included set of events. The events returned are the latest N events based on stream ordering in the room.
2. `/messages` which returns events in topological ordering and includes topological tokens for the start and end of the included set of events. You can specify either stream or topological tokens for the upper and lower bounds of requested events. The events returned are the latest N events based on topological ordering in the room, filtered by the tokens specified.

The question is: if we use a stream token from the start of the timeline in `/sync` to paginate via `/messages`, do we see all events?

Let's say we do a `/sync` and get a `prev_batch` stream token of `s`. When we do `/messages` this will return the set of events ordered by topological ordering less than `s`, greatest first. i.e. there is a topological token `t` (corresponding to the first event `/messages` returns) for which any event before `t` will be returned by `/messages` (except maybe if the event is after `s`).

Given an arbitrary event `E` in the room, what cases are there and will it turn up in subsequent `/sync` or `/messages`?

1. If `E` is after `s` then it will come down `/sync`, and we're done.
2. If `E` is before `t` then it will come down `/messages` and we're done.
3. By construction of `t` there cannot be an event that is after `t` but before `s`, as `t` is defined to be a token that returns all events before `s`. i.e., `E` must be in one of the two cases above.

This is true for the case where we receive `E` and *then* do the the pagination. What about the case where we have already started paginating from a previous `/sync` request? For non-backfilled events then `E` must by definition of stream ordering happen after `s`, and so come down `/sync` if there are no gaps. For the gappy sync case the client will have to paginate, but then the same rationale as above applies.

PR #704, opened 2024-01-24
Add device list update regression test

kegsay:
For https://github.com/element-hq/synapse/issues/11374


kegsay:
Need branch name to match.  See https://github.com/matrix-org/complement/pull/706

PR #705, opened 2024-01-26
Annotate `client.CSAPI` with `.Password`

kegsay:
This field isn't really used anywhere, but it is extremely useful to be tagged with the client itself when you want to derive a new client from an existing one. Mostly helpful for Complement Crypto currently.

### Pull Request Checklist

### DIFF
--- a/internal/docker/deployment.go
+++ b/internal/docker/deployment.go
@@ -99,6 +99,7 @@ func (d *Deployment) Register(t ct.TestLike, hsName string, opts helpers.Registr
 		Client:           client.NewLoggedClient(t, hsName, nil),
 		SyncUntilTimeout: 5 * time.Second,
 		Debug:            d.Deployer.debugLogging,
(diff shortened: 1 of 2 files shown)


internal/docker/deployment.go
```diff
@@ -99,6 +99,7 @@ func (d *Deployment) Register(t ct.TestLike, hsName string, opts helpers.Registr
 		Client:           client.NewLoggedClient(t, hsName, nil),
 		SyncUntilTimeout: 5 * time.Second,
 		Debug:            d.Deployer.debugLogging,
+		Password:         opts.Password,
```
S7evinK, line 102: The password, if empty and using the default, is not written back to the client [below](https://github.com/matrix-org/complement/blob/d690a2ff37d1a62728172ddca90229493236503c/internal/docker/deployment.go#L108-L111).

PR #707, opened 2024-01-30
Run the post script _before_ killing the containers

kegsay:
It's not very useful having the container ID otherwise..


reivilibre (approved):

S7evinK:
Maybe should have added the reason why it is like this here as well when @kegsay asked for a review:
The current approach makes sure that e.g. Dendrite is shutdown cleanly _before_ attempting to pull out coverage data, which is only flushed out to disk on a clean shutdown (that's also why `complementRuntime.ContainerKillFunc` is a thing). Moving this up makes this now unusable for this use case.

I understand why it may be moved up, but this breaks coverage reporting for Dendrite.

From my perspective, we need both use cases.

PR #710, opened 2024-02-16
Regression test for https://github.com/element-hq/synapse/issues/16928

kegsay:
And some drive-by logging improvements, along with a new `ServerRoom.WaiterForEvent(eventID)` function which makes waiting for an event to arrive on the Complement server much much easier to write in code.


richvdh:
For linking: https://github.com/element-hq/synapse/issues/16928

erikjohnston (approved):

PR #711, opened 2024-02-16
Refactor the match.CheckOff API to use functional options

kegsay:
This means you don't need to tack on `, nil` all the time, and can optionally specify when you want to allow unwanted items, obviating the need for `JSONCheckOffAllowUnwanted`. This composes better and supports adding additional functionality e.g allowing duplicate items.

### DIFF
--- a/match/json.go
+++ b/match/json.go
@@ -113,7 +113,69 @@ func JSONKeyArrayOfSize(wantKey string, wantSize int) JSON {
 	}
 }
 
-func jsonCheckOffInternal(wantKey string, wantItems []interface{}, allowUnwantedItems bool, mapper func(gjson.Result) interface{}, fn func(interface{}, gjson.Result) error) JSON {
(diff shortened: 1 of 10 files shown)


match/json.go
```diff
@@ -128,11 +190,14 @@ func jsonCheckOffInternal(wantKey string, wantItems []interface{}, allowUnwanted
 			if res.IsArray() {
 				itemRes = val
 			}
-			// convert it to something we can check off
-			item := mapper(itemRes)
-			if item == nil {
-				err = fmt.Errorf("JSONCheckOff(%s): mapper function mapped %v to nil", wantKey, itemRes.Raw)
-				return false
+			var item interface{} = itemRes
```
devonh, line 193: Nifty use of `interface` here

devonh (approved):
Everything checks out. Definitely makes the api a little easier to understand.

PR #712, opened 2024-02-19
Add a regression test for a /sync bug

richvdh:
Ensure that `state` is calculated correctly for archived rooms, even when there are no events returned in the timeline.

cf https://github.com/element-hq/synapse/pull/16932

### DIFF
--- a/tests/csapi/sync_archive_test.go
+++ b/tests/csapi/sync_archive_test.go
@@ -1,9 +1,8 @@
 package csapi_tests
 
 import (
-	"testing"
-
 	"github.com/tidwall/gjson"
+	"testing"
 
 	"github.com/matrix-org/complement"
 	"github.com/matrix-org/complement/b"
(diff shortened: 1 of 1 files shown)


tests/csapi/sync_archive_test.go
```diff
@@ -190,43 +190,106 @@ func TestArchivedRoomsHistory(t *testing.T) {
 		},
 	})
 
-	exhaustiveResCheck := func(result gjson.Result, origin string) {
-		roomRes := result.Get("rooms.leave." + client.GjsonEscape(roomID))
+	t.Run("timeline has events", func(t *testing.T) {
```
erikjohnston, line 193: Can we comment on what each of these subtests are testing here please? (I know this one toy didn't add, but since you're here...)

richvdh: Have added a couple of comments - are they clear enough?

erikjohnston (approved):

PR #716, opened 2024-03-05
Add test for shared history visibility over federation

kegsay:
Currently fails on Synapse.

### DIFF
--- a/tests/federation_room_send_test.go
+++ b/tests/federation_room_send_test.go
@@ -249,3 +249,51 @@ func TestNetworkPartitionOrdering(t *testing.T) {
 		},
 	})
 }
+
+// Tests that /sync v2 returns the correct timeline for the following situation _over federation_:
+//   - Alice and Bob in a room.
+//   - Bob leaves the room.
+//   - Alice sends a message.
+//   - Bob joins the room.
+//   - The next sync response should include a timeline which ends with [BOB_LEAVE, ALICE_MSG, BOB_JOIN].
+//
+// In practice, it seems Synapse can respond with [BOB_LEAVE, BOB_JOIN], despite the room being shared history visibility.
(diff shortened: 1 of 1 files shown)


tests/federation_room_send_test.go
```diff
@@ -249,3 +249,51 @@ func TestNetworkPartitionOrdering(t *testing.T) {
 		},
 	})
 }
+
+// Tests that /sync v2 returns the correct timeline for the following situation _over federation_:
+//   - Alice and Bob in a room.
+//   - Bob leaves the room.
+//   - Alice sends a message.
+//   - Bob joins the room.
+//   - The next sync response should include a timeline which ends with [BOB_LEAVE, ALICE_MSG, BOB_JOIN].
```
erikjohnston, line 258: I think it would be perfectly valid to return just `BOB_JOIN` here, which I thought we did

tests/federation_room_send_test.go
```diff
@@ -249,3 +249,51 @@ func TestNetworkPartitionOrdering(t *testing.T) {
...
+	_, nextBatch = bob.MustSync(t, client.SyncReq{Since: nextBatch})
+
+	// bob rejoins and syncs, he should see this message
+	bob.MustJoinRoom(t, roomID, []string{"hs1"})
+	syncResponse, _ := bob.MustSync(t, client.SyncReq{Since: nextBatch})
+	timeline := syncResponse.Get(fmt.Sprintf("rooms.join.%s.timeline.events", client.GjsonEscape(roomID)))
+	t.Logf("timeline => %v", timeline.Raw)
+	timelineArray := timeline.Array()
+	must.NotEqual(t, len(timelineArray), 0, "no timeline for room "+roomID)
+	must.Equal(t, len(timelineArray) >= 3, true, "need at least 3 timeline events for the room")
```
erikjohnston, line 292: This doesn't print the number of events there are, which makes seeing how Synapse is failing hard

PR #719, opened 2024-04-10
Regressiong test for element-hq/synapse#16940

kegsay:


erikjohnston (approved):
LGTM, though we can't merge it.

PR #725, opened 2024-06-07
Stable support for MSC4115-style membership properties

richvdh:
MSC4115 has now completed FCP, so we can move it to the main test suite and use stable properties.

Requires https://github.com/element-hq/synapse/pull/17282 to land before it will pass CI.


richvdh:
> Requires https://github.com/element-hq/synapse/pull/17282 to land before it will pass CI.

Oh, no it doesn't, because complement's CI checks out the matching branch of Synapse.

tests/csapi/membership_on_events_test.go
```diff
@@ -10,11 +10,11 @@ import (
 	"github.com/matrix-org/complement/helpers"
 )
 
-// MSC4115: membership information on events
+// Membership information on events served to clients, as specified in MSC4115.
 //
-// Alice sends one message before bob joins, then one after. Bob reads both messages, and checks the membership state
+// Alice sends one message before Vob joins, then one after. Bob reads both messages, and checks the membership state
```
reivilibre, line 15: Vob? :D

reivilibre (approved):
fine otherwise :D

tests/csapi/membership_on_events_test.go
richvdh: :facepalm: 
```suggestion
// Alice sends one message before Bob joins, then one after. Bob reads both messages, and checks the membership state
```

PR #728, opened 2024-07-02
Update thumbnail tests to include _matrix/client/v1/media/thumbnail endpoint

H-Shay:


tests/media_thumbnail_test.go
```diff
@@ -29,7 +30,10 @@ func TestLocalPngThumbnail(t *testing.T) {
 
 	uri := alice.UploadContent(t, data.LargePng, fileName, contentType)
 
-	fetchAndValidateThumbnail(t, alice, uri)
+	fetchAndValidateThumbnail(t, alice, uri, false)
```
reivilibre, line 33: There's nothing wrong with this, but for this and the Remote test, I think it'd be better if the old and new endpoints were separate tests.

Supposing you had a defect in one of the versions of endpoints, you'd want the test results to show which version was working fine and which version wasn't.

Equally, if a new cutting-edge homeserver Syndruit wants to skip implementing the legacy endpoints, they may want to blacklist the tests just for the old version.

I asked in the Complement room and it seems doing this as subtests would work for this and it looks fairly easy to do: seems like you just wrap each part of the test like this:

```suggestion
    t.Run("unauthenticated", func(t *testing.T) {
	    fetchAndValidateThumbnail(t, alice, uri, false)
	})
```

& similar for the other ones (The string argument to `t.Run` is a subtest name)

reivilibre (approved):
looks like an updated test :D

if you have any trouble trying to apply the subtest idea then let me know, don't want to make things unnecessarily hard, it just seemed like a clean thing to do

H-Shay:
I've added subtests (thanks for the example!) @reivilibre do you think you could merge this for me if the changes are satisfying?

PR #731, opened 2024-07-27
Handle authed media endpoint

S7evinK:
This should fix issues where Dendrite attempts to call the authed media endpoint first and only falling back to the legacy media endpoint, as seen [here](https://github.com/matrix-org/dendrite/actions/runs/10125662955/job/28001444870?pr=3397#step:7:54).


S7evinK:
Dendrite failing is expected, as it needs https://github.com/matrix-org/dendrite/pull/3397 first.

PR #732, opened 2024-08-06
Add a basic test fetching thumbnail from `_matrix/federation/v1/media/thumbnail`

H-Shay:
As the title states. https://github.com/element-hq/synapse/issues/17518 was missed because Synapse does not request thumbnails via this endpoint, rather fetches the media over the `/download` endpoint and then generates the thumbnails. This PR adds a test to ensure the basic functionality of the `/thumbnail` endpoint.

### DIFF
--- a/tests/media_thumbnail_test.go
+++ b/tests/media_thumbnail_test.go
@@ -132,6 +132,7 @@ func TestFederationThumbnail(t *testing.T) {
 		t.Fatalf("failed to parse multipart response: %s", err)
 	}
 
+	foundImage := false
 	reader := multipart.NewReader(resp.Body, params["boundary"])
 	for {
 		p, err := reader.NextPart()
@@ -151,7 +152,11 @@ func TestFederationThumbnail(t *testing.T) {
 			if !bytes.Equal(imageBody, body) {
 				t.Fatalf("body does not match uploaded file")
 			}
+			foundImage = true
 		}
+	}
+	if !foundImage {
+		t.Fatalf("No image was found in response.")
 	}
 }
 
--- a/tests/media_thumbnail_test.go
+++ b/tests/media_thumbnail_test.go
@@ -14,9 +14,10 @@ import (
 	"image/jpeg"
 	"image/png"
 	"io"
+	"mime"
+	"mime/multipart"
 	"net/http"
 	"net/url"
-	"strconv"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


H-Shay:
(note that Synapse failure would be expected since the test can't pull in my fork)

tests/media_thumbnail_test.go
```diff
@@ -68,6 +72,60 @@ func TestRemotePngThumbnail(t *testing.T) {
...
+	mimeType := strings.Split(resType, ";")[0]
+
+	if mimeType != "multipart/mixed" {
+		t.Fatalf("thumbnail request did not return multipart/mixed response, returned %s instead", mimeType)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	length := len(body)
+
+	contentLength, err := strconv.Atoi(resp.Header.Get("Content-Length"))
```
erikjohnston, line 121: Can we instead compare the received boddy to `test.png`? I don't think we necessarily should mandate that servers set `Content-Length` in the black box test suite.

H-Shay: I think it probably makes sense to move the content-length check to the existing Synapse-specific tests for this endpoint: https://github.com/element-hq/synapse/blob/70b0e386032ed2f3ecf25fcf9a4b6c31335ffdc4/tests/federation/test_federation_media.py#L152

Xiretza: Complement should definitely be checking `Content-Length` for all requests on all implementations - maybe it already is and was just missing a call to the federation thumbnail endpoint, I don't know.

erikjohnston: To expand: HTTP servers can either respond with `Content-Length` *or* chunked transfer encoding, and I don't think we should be mandating one or the other. We could do this by explicitly checking for the two cases, but what we actually care about is that the file was correctly transmitted (and it won't be if the `Content-Length` is wrong, unless golang HTTP client does something non-HTTP-spec compliant).

tests/media_thumbnail_test.go
```diff
@@ -68,6 +73,88 @@ func TestRemotePngThumbnail(t *testing.T) {
...
+		if partContentType == contentType {
+			imageBody, err := io.ReadAll(p)
+			if err != nil {
+				t.Fatalf("failed to read multipart part %s: %s", partContentType, err)
+			}
+			if !bytes.Equal(imageBody, body) {
+				t.Fatalf("body does not match uploaded file")
+			}
+		}
+	}
```
erikjohnston, line 155: Can we add a check after the loop to make sure that an image was actually included in the response, mostly to make it clear that we are *actually* checking the image. Just something like `have_checked_image = true; ...; assert have_changed_image`

erikjohnston (approved):
Otherwise LGTM

tests/media_thumbnail_test.go
H-Shay: Right this is sorted - if you're happy with the change can you hit the merge button for me (as I no longer posses the power). Thanks so much!

Issue #729, opened 2024-07-12
Allow tests that require libolm to be opted out of


kegsay:
So:
 - A) we don't need to force people to install libolm when they don't run crypto tests
 - B) we can import different versions of mautrix without causing version incompatibilities

PR #738, opened 2024-09-24
Remove libolm dependency

kegsay:
Fixes https://github.com/matrix-org/complement/issues/729 - having libolm is a liability for building on top of and pointless given it is deprecated.

This is a breaking change because it removes blueprint functionality. Specifically, it removes:
 - the field `OneTimeKeys` from the `User` struct.
 - the performance blueprint `BlueprintPerfE2EERoom`.

It does not break or remove any tests.


S7evinK (approved):
LGTM.
Probably makes sense to do the same in Dendrite as well at some point.

PR #743, opened 2024-10-31
Add test for transactions containing a V1 event into a V10 room

morguldir:
### Pull Request Checklist

### DIFF
--- a/tests/federation_room_send_test.go
+++ b/tests/federation_room_send_test.go
@@ -1,6 +1,7 @@
 package tests
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/url"
@@ -9,11 +10,14 @@ import (
 
 	"github.com/matrix-org/complement"
 	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/gomatrixserverlib/spec"
 	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
 	"golang.org/x/exp/slices"
 
 	"github.com/matrix-org/complement/b"
 	"github.com/matrix-org/complement/client"
+	"github.com/matrix-org/complement/ct"
 	"github.com/matrix-org/complement/federation"
 	"github.com/matrix-org/complement/helpers"
 	"github.com/matrix-org/complement/match"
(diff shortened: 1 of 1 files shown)


tests/federation_room_send_test.go
```diff
@@ -249,3 +253,70 @@ func TestNetworkPartitionOrdering(t *testing.T) {
...
+
+	if err != nil {
+		ct.Fatalf(t, "Invalid PDU in transaction made the request fail")
+	}
+
+	for eventID, e := range resp.PDUs {
+		if eventID != "$meow" && eventID != event.EventID() {
+			ct.Fatalf(t, "Server responded with bogus event ID %s", eventID)
+		} else if e.Error == "" {
+			ct.Fatalf(t, "Server accepted event sent into a V10 room using V1 format")
```
S7evinK, line 312: At least for Dendrite, this is fine. When it receives a transaction, [it extracts the room ID, asks the DB about the version and then executes ](https://github.com/element-hq/dendrite/blob/7cc7ebb46f1f06d08747199542a164f0400e0b5e/internal/transactionrequest.go#L106-L135) `NewEventFromUntrustedJSON` ([link](https://github.com/element-hq/dendrite/blob/7cc7ebb46f1f06d08747199542a164f0400e0b5e/internal/transactionrequest.go#L106-L135)), which for a `V10` room (not specified in the test) runs `newEventFromUntrustedJSONV2` ([link](https://github.com/matrix-org/gomatrixserverlib/blob/dbd5f31fefc031633c3418165e4ef6d343e03999/eventV2.go#L122-L138)), which strips out the `event_id`.

tests/federation_room_send_test.go
```diff
@@ -249,3 +253,70 @@ func TestNetworkPartitionOrdering(t *testing.T) {
...
+		federation.HandleTransactionRequests(nil, nil),
+	)
+
+	cancel := srv.Listen()
+	defer cancel()
+
+	nexy := srv.UserID("nexy")
+
+	room := alice.MustCreateRoom(t, map[string]interface{}{
+		"preset": "public_chat",
```
S7evinK, line 275: As the log below wants room version 10, specify it, as it otherwise defaults to the server implementations default (which might be unknown).
```suggestion
		"preset": "public_chat",
		"room_version": "10",
```

tests/federation_room_send_test.go
```diff
@@ -249,3 +253,70 @@ func TestNetworkPartitionOrdering(t *testing.T) {
...
+	})
+	server_room := srv.MustJoinRoom(t, deployment, "hs1", room, nexy)
+
+
+	event := srv.MustCreateEvent(t, server_room, federation.Event{
+		Sender: nexy,
+		Type: "m.room.message",
+		Content: map[string]interface{}{"body": "hello i am nexy and i somehow broke the room into a v1 room"},
+	})
+	//event.EventID()
```
S7evinK, line 285:
```suggestion
```

tests/federation_room_send_test.go
```diff
@@ -249,3 +253,70 @@ func TestNetworkPartitionOrdering(t *testing.T) {
...
+	data, err := json.Marshal(event)
+	if err != nil {
+		t.Log("invalid json bwuh", err)
+		t.FailNow()
+	}
+	data, err = sjson.SetBytes(data, "event_id", event.EventID())
```
S7evinK, line 291: Missing error check.

S7evinK (changes requested):
See comments below.

tests/federation_room_send_test.go
morguldir: Added a bogus prev_event too, now it will get the INFO log about not including the event id in the error dict instead

tests/federation_room_send_test.go
```diff
@@ -249,3 +253,70 @@ func TestNetworkPartitionOrdering(t *testing.T) {
...
+		PDUs: []json.RawMessage{data},
+		EDUs: nil,
+	})
+
+	if err != nil {
+		ct.Fatalf(t, "Invalid PDU in transaction made the request fail")
+	}
+
+	for eventID, e := range resp.PDUs {
+		if eventID != "$meow" && eventID != event.EventID() {
```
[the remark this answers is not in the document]
morguldir: Oops, i think that was the bad event id i used originally, but ended up just calculating it instead, fixed

PR #744, opened 2024-11-07
Add test for the order in which OTKs are issued

richvdh:
Per MSC4225, they should be issued in the same order they were uploaded.

### DIFF
--- a/tests/csapi/upload_keys_test.go
+++ b/tests/csapi/upload_keys_test.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/tidwall/gjson"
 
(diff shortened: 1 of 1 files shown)


tests/csapi/upload_keys_test.go
anoadragon453, line 1: As MSC4225 is not accepted yet, this test should be in a separate file with an [MSC build tag](https://github.com/matrix-org/complement?tab=readme-ov-file#build-tags) to prevent it from being run on homeservers that haven't yet implemented support for this MSC.

anoadragon453 (changes requested):
The code LGTM, just one small query.

tests/csapi/upload_keys_test.go
richvdh: can we just wait for MSC4225 to land?

tests/csapi/upload_keys_test.go
```diff
@@ -174,6 +175,45 @@ func TestUploadKey(t *testing.T) {
...
+	defer deployment.Destroy(t)
+	alice := deployment.Register(t, "hs1", helpers.RegistrationOpts{})
+	_, oneTimeKeys := alice.MustGenerateOneTimeKeys(t, 2)
+
+	// first upload key 1, sleep a bit, then upload key 0.
+	otk1 := map[string]interface{}{"signed_curve25519:1": oneTimeKeys["signed_curve25519:1"]}
+	alice.MustDo(t, "POST", []string{"_matrix", "client", "v3", "keys", "upload"},
+		client.WithJSONBody(t, map[string]interface{}{"one_time_keys": otk1}))
+
+	time.Sleep(1 * time.Second)
```
kegsay, line 190: Why do we need to sleep a bit? We process OTKs asyncly in synapse?

kegsay (approved):
MSC landed so shipit

tests/csapi/upload_keys_test.go
richvdh: oh, this was to make sure there was a difference in timestamp between the two upload requests. I should probably comment to that effect.


```suggestion
    // Ensure that there is a difference in timestamp between the two upload requests.
	time.Sleep(1 * time.Second)
```

PR #747, opened 2024-11-19
Skip tests requiring unauthenticated media on Synapse

reivilibre:
These will be expected to fail when Synapse enforces authenticated media by default in
https://github.com/element-hq/synapse/pull/17889

### Pull Request Checklist

Signed-off-by: Olivier 'reivilibre' <oliverw@element.io>

### DIFF
--- a/tests/media_filename_test.go
+++ b/tests/media_filename_test.go
@@ -41,6 +41,7 @@ func TestMediaFilenames(t *testing.T) {
 				var filename = filename
 
 				t.Run(fmt.Sprintf("Can download file '%s'", filename), func(t *testing.T) {
+					runtime.SkipIf(t, runtime.Synapse)
 					t.Parallel()
 
 					mxcUri := alice.UploadContent(t, data.MatrixPng, filename, "image/png")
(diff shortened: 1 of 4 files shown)


tests/media_filename_test.go
```diff
@@ -41,6 +41,7 @@ func TestMediaFilenames(t *testing.T) {
 				var filename = filename
 
 				t.Run(fmt.Sprintf("Can download file '%s'", filename), func(t *testing.T) {
+					runtime.SkipIf(t, runtime.Synapse)
```
erikjohnston, line 44: I feel like we would want to keep these tests but change them to use authentication?

I think I'd prefer to disable authenticated media when testing with complement rather than not test these paths?

reivilibre: These tests already exist with authentication, see e.g. L56 for this one

erikjohnston: Ah right, fair enough

erikjohnston: Can you add a comment or something like "Synapse doesn't support unauthenticated media anymore"?

erikjohnston (approved):

PR #749, opened 2024-11-21
Test for passing through `unsigned` data in `/keys/query`

richvdh:
We'd like a mechanism by which a client can add "unsigned" data to their device keys, and have it be accessible by other clients involved in E2EE discussions.

An MSC for this behaviour is pending, but will be matrix-org/matrix-spec-proposals#4229.

This PR adds a test for that functionality. It tests functionality added in https://github.com/element-hq/synapse/pull/17951.

Part of element-hq/element-meta#2467.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,9 +42,9 @@ jobs:
           - homeserver: Synapse
             repo: element-hq/synapse
             tags: synapse_blacklist
-            packages: |-
-              ./tests/msc3874
-              ./tests/msc3902
+            packages: \
+              ./tests/msc3874 \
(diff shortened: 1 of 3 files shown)


S7evinK:
Note that this newly added test was not executed, as it is missing in the `packages` variable in [CI](https://github.com/matrix-org/complement/blob/9e9f6d36b105a6aa37491a3b43ba6d0ed28feb2c/.github/workflows/ci.yaml#L45)

richvdh:
sooo, we're putting this work (https://github.com/element-hq/element-meta/issues/2467) on indefinite hold; I'm therefore closing this PR. If we come back to it, we'll reopen.

PR #750, opened 2024-12-17
Support functional options in TestMain

kegsay:
Rather than having `TestMainWithCleanup` and `TestMain`, use functional options. Add support for a custom deployment mechanism at the same time.

This is a breaking change for users who use `TestMainWithCleanup(m, namespace, cleanupFn)` who now need to do `TestMain(m, namespace, complement.WithCleanup(cleanupFn))`.

### DIFF
--- a/test_main.go
+++ b/test_main.go
@@ -16,7 +16,7 @@ var (
 
 type complementOpts struct {
 	cleanup          func()
-	customDeplyoment func(numServers int) Deployment
+	customDeployment func(numServers int) Deployment
 }
 type opt func(*complementOpts)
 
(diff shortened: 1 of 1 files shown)


anoadragon453 (approved):
The code changes look good to me, though I'd like to see an implementation first (public or private) before merging.

test_main.go
```diff
@@ -9,7 +9,33 @@ import (
...
 
-var testPackage *TestPackage
+var (
+	testPackage    *TestPackage
+	customDeployer func(numServers int) Deployment
+)
+
+type complementOpts struct {
+	cleanup          func()
+	customDeplyoment func(numServers int) Deployment
```
S7evinK, line 19: Minor typo:
```suggestion
	customDeployment func(numServers int) Deployment
```

kegsay: Fixed, thanks!

gaelgatelement:
Can we merge this now ? I've tested it successfully.

anoadragon453:
I've checked the private implementation and it looks to work. Thanks!

PR #752, opened 2024-12-31
Make restricted room join tests clearer

morguldir:
Firstly this PR makes the tests work on more implementations, for example ruma and GMS won't really understand `join_authorised_via_users_server` not looking like a user id, similar to what https://github.com/matrix-org/complement/pull/224 tried to do

Making it a user id that isn't correct still makes sure that the server can handle clients updating the member event directly with /myroomnick or similar

Some errors were also silently being ignored, which made it harder to tell which part of the test was actually failing

### Pull Request Checklist

### DIFF
--- a/tests/restricted_rooms_test.go
+++ b/tests/restricted_rooms_test.go
@@ -99,8 +99,6 @@ func checkRestrictedRoom(t *testing.T, alice *client.CSAPI, bob *client.CSAPI, a
 				Content: map[string]interface{}{
 					"membership":  "join",
 					"displayname": "Bobby",
-					// This should be ignored since this is a join -> join transition.
-					"join_authorised_via_users_server": "unused",
 				},
 			},
 		)
--- a/tests/restricted_rooms_test.go
+++ b/tests/restricted_rooms_test.go
@@ -99,6 +99,8 @@ func checkRestrictedRoom(t *testing.T, alice *client.CSAPI, bob *client.CSAPI, a
 				Content: map[string]interface{}{
 					"membership":  "join",
 					"displayname": "Bobby",
+					// This should be ignored by the server since this is a join -> join transition
+					"join_authorised_via_users_server": "@unused:unused.local",
 				},
 			},
 		)
--- a/tests/restricted_rooms_test.go
+++ b/tests/restricted_rooms_test.go
@@ -64,7 +64,7 @@ func checkRestrictedRoom(t *testing.T, alice *client.CSAPI, bob *client.CSAPI, a
 
 	t.Run("Join should succeed when joined to allowed room", func(t *testing.T) {
 		// Join the allowed room.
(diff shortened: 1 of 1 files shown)


tests/restricted_rooms_test.go
```diff
@@ -99,8 +99,8 @@ func checkRestrictedRoom(t *testing.T, alice *client.CSAPI, bob *client.CSAPI, a
 				Content: map[string]interface{}{
 					"membership":  "join",
 					"displayname": "Bobby",
-					// This should be ignored since this is a join -> join transition.
-					"join_authorised_via_users_server": "unused",
+					// This should be ignored by the server since this is a join -> join transition
+					"join_authorised_via_users_server": "@unused:unused.local",
```
clokep, line 103: What happens if a ruma-based server gets garbage here then? Does it still work?

morguldir: It can create the event then, but it won’t really work properly unless the server removes the field when handling the PUT like synapse does

So when verifying the signature of a federation event actually looking like that it won’t work still

This part i think is intended though, although seemingly synapse changed that behaviour at some point too 🧐

clokep: I'm not really following your comment, sorry. Shouldn't the field be removed no matter what? So what isn't working for ruma?

morguldir: Sorry yeah it should, but if another server forgets to remove it, then ruma still tries to check the signature, but synapse seems to still allow the event

tests/restricted_rooms_test.go
```diff
@@ -114,6 +114,15 @@ func checkRestrictedRoom(t *testing.T, alice *client.CSAPI, bob *client.CSAPI, a
...
 		// attempt to join the room.
+		alice.MustSyncUntil(t, client.SyncReq{}, client.SyncTimelineHas(
+			room, func(ev gjson.Result) bool {
+				if ev.Get("type").Str != "m.room.member" || ev.Get("sender").Str != bob.UserID {
+					return false
+				}
+
+				return ev.Get("content").Get("membership").Str == "leave"
+			}))
+
```
kegsay, line 125: This is a duplicate of the stuff underneath? Why add this?

morguldir: It makes sure the leave worked in both rooms now

tests/restricted_rooms_test.go
anoadragon453: Irrespective of the behaviour of Synapse and Ruma, this field should either be set to a valid structure for this field (a Matrix ID), or it should explicitly call out that this field is set to an invalid value on purpose.

As the latter does not appear the case, I agree with this change. However I'm not opposed to another test that checks homeservers correctly handle invalid content in this field.

anoadragon453 (changes requested):
Minor request, otherwise LGTM!

PR #757, opened 2025-01-18
Add remote invite -> join/reject tests for rooms that a homeserver is already participating in

MadLittleMods:
Add remote invite -> join/reject tests for rooms that a homeserver is already participating in

Regression tests for https://github.com/element-hq/synapse/pull/18075

### Pull Request Checklist


kegsay (approved):
Thorough documentation, but you can probably do better tests here. The current ones are end-to-end with two HS impls, so you cannot engineer the race condition. If you used a Complement HS you could control when the invite comes in via `/send`.

MadLittleMods:
Good shout @kegsay! I'll continue on an engineered solution to go alongside these :fast_forward:

PR #765, opened 2025-02-27
Open up more ServerRoom functions

kegsay:
This allows more experimentation because it allows Complement to make ServerRoom structs for unknown room versions.


devonh (approved):
Don't see any reason not to expose these.

PR #766, opened 2025-02-28
Add ServerRoomImpl to allow tests to reuse helpers with custom code

kegsay:
With the addition of https://github.com/matrix-org/complement/pull/765 it's now possible for tests to add a range of good/bad/broken rooms to Complement's internal federation server. However, these rooms cannot reuse any of the existing helper functions. To aid this, logic which was once part of the federation server itself (e.g mapping `/send_join` responses to internal `ServerRoom` state) is now transferred to a new `ServerRoomImpl` interface. The federation server's helper functions call this interface, allowing tests to swap out functions for their broken rooms.

For example, we may want to reuse the very convenient `HandleMakeSendJoinRequests` but modify the join proto event to include malformed fields. Before this PR this simply wasn't possible as we hard-coded the happy case. Tests can now do:
```go
room := federation.NewServerRoom(gomatrixserverlib.RoomVersionV1, "!foo:localhost")
room.ServerRoomImpl = &federation.ServerRoomImplCustom{
	ServerRoomImplDefault: federation.ServerRoomImplDefault{Room: room},
	ProtoEventCreatorFn: func(def federation.ServerRoomImpl, ev federation.Event) (*gomatrixserverlib.ProtoEvent, error) {
		if ev.Type == spec.MRoomMember {
			ev.Unsigned = map[string]any{"something": "custom"}
		}
		return def.ProtoEventCreator(ev)
	},
}
```

### DIFF
--- a/federation/handle.go
+++ b/federation/handle.go
@@ -59,7 +59,7 @@ func MakeJoinRequestsHandler(s *Server, w http.ResponseWriter, req *http.Request
 // or dealing with HTTP responses itself.
 func MakeRespMakeJoin(s *Server, room *ServerRoom, userID string) (resp fclient.RespMakeJoin, err error) {
 	// Generate a join event
-	proto, err := room.ProtoEventCreator(room, Event{
+	proto, err := room.ProtoEventCreator(Event{
 		Type:     "m.room.member",
 		StateKey: &userID,
 		Content: map[string]interface{}{
@@ -84,7 +84,7 @@ func MakeRespMakeJoin(s *Server, room *ServerRoom, userID string) (resp fclient.
 // or dealing with HTTP responses itself.
 func MakeRespMakeKnock(s *Server, room *ServerRoom, userID string) (resp fclient.RespMakeKnock, err error) {
 	// Generate a knock event
-	proto, err := room.ProtoEventCreator(room, Event{
+	proto, err := room.ProtoEventCreator(Event{
 		Type:     "m.room.member",
 		StateKey: &userID,
 		Content: map[string]interface{}{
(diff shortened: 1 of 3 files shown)


federation/server_room.go
```diff
@@ -33,6 +37,15 @@ type Event struct {
 // EXPERIMENTAL
 // ServerRoom represents a room on this test federation server
 type ServerRoom struct {
+	// functions to map Complement events into actual PDUs.
+	// Most tests don't care about this and can use the default creator function,
+	// but if your MSC or tests fiddle with the raw JSON in some way then this
+	// function needs to be replaced. By replacing this function, helper functions
+	// which indirectly create events (e.g joins and leaves) will automatically use
+	// this and will hence work with your custom code. Modifying ProtoEventCreator
+	// will also impact the proto events returned via /make_xxx calls.
```
devonh, line 46:
```suggestion
	// Functions to map Complement events into actual PDUs.
	// Most tests don't care about this and can use the default functions,
	// but if your MSC or tests fiddle with the raw JSON in some way then these
	// function need to be replaced. By replacing these functions, helper functions
	// which indirectly create events (e.g joins and leaves) will automatically use
	// them and will hence work with your custom code.
```

federation/handle.go
```diff
@@ -91,29 +84,22 @@ func MakeRespMakeJoin(s *Server, room *ServerRoom, userID string) (resp fclient.
...
-	if err != nil {
-		err = fmt.Errorf("make_knock cannot set membership content: %w", err)
-		return
-	}
-	stateNeeded, err := gomatrixserverlib.StateNeededForProtoEvent(&proto)
+	proto, err := room.ProtoEventCreator(Event{
+		Type:     "m.room.member",
+		StateKey: &userID,
+		Content: map[string]interface{}{
+			"membership": spec.Join, // XXX this feels wrong?
```
devonh, line 91: Why? It seems okay to me.

devonh (approved):
New API functions should ideally have doc comments to explain their usage.
Even though they are pretty clear just from name & impl alone.

Otherwise these changes look promising and will make it much easier to mess around with future event changes without having to break into complement's internals.

federation/handle.go
kegsay: It's `MakeRespMakeKnock` so surely should be knock not join.

kegsay:
Will chuck some comments in then merge, thanks!

PR #770, opened 2025-03-19
Move internal/config to /config

kegsay:
This allows configs to be created and more importantly referenced as fields in other projects which depend on Complement internals.

Previously, we could only create configs via `NewTestPackage().GetConfig()` which would set the config from env vars.


devonh (approved):
Seems reasonable to me.

PR #772, opened 2025-04-04
Don't PublishAllPorts as we already explicitly do :8008/8448

kegsay:
This may break use cases which want extra ports to be exposed. This may fix issues where Complement sporadically decides that there are port conflicts, causing CI flakes.

### DIFF
--- a/internal/docker/deployer.go
+++ b/internal/docker/deployer.go
@@ -380,8 +380,7 @@ func deployImage(
 			"complement_hs_name":   hsName,
 		},
 	}, &container.HostConfig{
-		CapAdd:          []string{"NET_ADMIN"}, // TODO : this should be some sort of option
-		PublishAllPorts: true,
+		CapAdd: []string{"NET_ADMIN"}, // TODO : this should be some sort of option
 		PortBindings: nat.PortMap{
 			nat.Port("8008/tcp"): []nat.PortBinding{
 				{


internal/docker/deployer.go
```diff
@@ -380,8 +380,7 @@ func deployImage(
 			"complement_hs_name":   hsName,
 		},
 	}, &container.HostConfig{
-		CapAdd:          []string{"NET_ADMIN"}, // TODO : this should be some sort of option
-		PublishAllPorts: true,
```
MadLittleMods, line 384: > This may fix issues where Complement sporadically decides that there are port conflicts, causing CI flakes.

Where are you running into the CI flakes?

I'm curious because previously you said you didn't see them in vanilla complement runs, https://github.com/matrix-org/complement/pull/761

MadLittleMods, line 384: I think this will cause issues for some internal images we use at Element like testing the [Secure Border Gateway](https://element.io/server-suite/secure-border-gateways).

Internal Element link: https://github.com/element-hq/sbg/blob/3871b51144c4eacf1ea0355da2ac625a0bfadce5/complement/Dockerfile#L83-L88

Any advice for how to work around this if this merges?

kegsay: Hmm I wonder if we keep `PublishAllPorts` in the default case and then _only if_ `cfg.HSPortBindingIP` is not the default do we switch to `PortBindings`? That means it will publish all exposed ports by default, unless you bind to a specific IP (which afaik is only used by homerunner).

MadLittleMods: :shrug: Kinda seems fragile logic to gate the behavior change behind and would be hard to reason about for someone suddenly triggering this behavior.

kegsay: Only Homerunner seems to use the port binding IP stuff, and even then it was a niche usage of homerunner, so I've removed setting that for now. This is not being done in https://github.com/matrix-org/complement/pull/776

Issue #775, opened 2025-05-07
Docker v28 update causes `failed to listen on TCP socket: address already in use`


anoadragon453:
From May 5th, we started noticing `failed to listen on TCP socket: address already in use` errors in our Complement jobs on Synapse, for example: https://github.com/element-hq/synapse/actions/runs/14858385920/job/41717289741#step:7:1191

```
  2025/05/06 11:30:21 Sharing [SERVER_NAME=hs1 SYNAPSE_LOG_TESTING=1 SYNAPSE_COMPLEMENT_USE_WORKERS= SYNAPSE_COMPLEMENT_DATABASE=sqlite] host environment variables with container
  2025/05/06 11:30:22 fed.hs_with_application_service.hs1 : failed to deployBaseImage: Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint complement_fed.hs_with_application_service.hs1 (198c5615feb31d0f6a3fd3bb3cd990010b944a0daed9243503292e556aee3c4a): failed to listen on TCP socket: address already in use
  2025/05/06 11:30:22 ============================================
  
  
  2025/05/06 11:30:22 fed.hs_with_application_service.hs1 : Server logs:
  2025/05/06 11:30:22 ============== fed.hs_with_application_service.hs1 : END LOGS ==============
  
  
      federation_room_join_test.go:586: OldDeploy: Failed to construct blueprint: ConstructBlueprintIfNotExist(hs_with_application_service): failed to ConstructBlueprint: errors whilst constructing blueprint hs_with_application_service: [Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint complement_fed.hs_with_application_service.hs1 (198c5615feb31d0f6a3fd3bb3cd990010b944a0daed9243503292e556aee3c4a): failed to listen on TCP socket: address already in use]
```

This appears to have coincided with GitHub updating the `ubuntu24.04` runner, which [included](https://github.com/actions/runner-images/commit/9e60c8f84d5133822f4e9a51f949a59dd989a9b3) an update to the Docker Client and Server components:

```
- Docker Client 26.1.3
- Docker Server 26.1.3
⬇️
- Docker Client 28.0.4
- Docker Server 28.0.4
```

[...]

PR #776, opened 2025-05-07
Update docker to v28.1.1 and pin network interface name

anoadragon453:
See the explanation in https://github.com/moby/moby/issues/49935.

Also see the [Docker v28 release notes](https://docs.docker.com/engine/release-notes/28/#networking-4):

> - Add a new netlabel `com.docker.network.endpoint.ifname` to customize the interface name used when connecting a container to a network. It's supported by all built-in network drivers on Linux. [moby/moby#49155](https://github.com/moby/moby/issues/49155)
> 
>   - When a container is created with multiple networks specified, there's no guarantee on the order networks will be connected to the container. So, if a custom interface name uses the same prefix as the auto-generated names, for example `eth`, the container might fail to start.
>   - The recommended practice is to use a different prefix, for example `en0`, or a numerical suffix high enough to never collide, for example `eth100`.
> 
>   - This label can be specified on `docker network connect` via the `--driver-opt` flag, for example:
> 
>     ```bash
>     docker network connect --driver-opt=com.docker.network.endpoint.ifname=foobar …
>     ```
> 
>   * Or via the long-form `--network` flag on `docker run`, for example:
> 
>     ```bash
>     docker run --network=name=bridge,driver-opt=com.docker.network.endpoint.ifname=foobar …
>     ```

Closes https://github.com/matrix-org/complement/issues/775


anoadragon453:
Next issue to contend with:

```
CreateDirtyDeployment failed: CreateDirtyDeployment: dirty: failed to list networks. Error response from daemon: client version 1.49 is too new. Maximum supported API version is 1.48
```

I plan to split the go1.22 -> 1.23 update into a separate PR once CI is passing.

PR #777, opened 2025-05-07
fix: downgrade docker to v28.0.4

3nprob:
- Targeting #776

### Pull Request Checklist


kegsay:
Thanks, but this is still too new. Developers who install Docker are often behind. For example, Docker Desktop currently only supports up to 1.47. Need to go slightly further back. Will apply directly to the target PR.

PR #778, opened 2025-05-13
Custom `Deployment` quality of life improvements (pass `t`, config)

MadLittleMods:
Spawning from a real use-case with a custom `Deployment`/`Deployer` (Element-internal). I think this is the first working example of a custom deployment so it makes sense that we ran into a couple oversights.

This PR introduces some quality of life for working with a custom `Deployment`. For reference, custom deployments were introduced in https://github.com/matrix-org/complement/pull/750.

 - Pass `t` to allow the custom deployment creation callback (`complement.WithDeployment(callback)`) to handle errors gracefully
 - Also pass in `complement.Config` so the custom Deployment can align behavior to what's configured
 
### Dev notes

<details>
<summary>Reference: Using a custom <code>Deployment</code> with Complement</summary>

```go
func TestMain(m *testing.M) {
    complement.TestMain(m, "custom_deployment",
        complement.WithDeployment(complement.WithDeployment(func(numServers int) Deployment {
            // return the custom deployment which implements `complement.Deployment`
        })
    )
}
```

</details>

### Pull Request Checklist

### DIFF
--- a/internal/docker/deployment.go
+++ b/internal/docker/deployment.go
@@ -57,16 +57,6 @@ func (hsDep *HomeserverDeployment) SetEndpoints(baseURL string, fedBaseURL strin
 	}
 }
 
-func (d *Deployment) GetFullyQualifiedHomeserverName(hsName string) (string, error) {
-	_, ok := d.HS[hsName]
-	if !ok {
-		return "", fmt.Errorf("Deployment.GetFullyQualifiedHomeserverName - HS name '%s' not found", hsName)
-	}
-	// We have network aliases for each Docker container that will resolve the `hsName` to
-	// the container.
-	return hsName, nil
-}
-
 // DestroyAtCleanup destroys the entire deployment. It should be called at cleanup time for dirty
 // deployments only. Handles configuration options for things which should run at container destroy
 // time, like post-run scripts and printing logs.
--- a/internal/docker/deployment.go
+++ b/internal/docker/deployment.go
@@ -57,14 +57,14 @@ func (hsDep *HomeserverDeployment) SetEndpoints(baseURL string, fedBaseURL strin
 	}
 }
 
-func (d *Deployment) GetFullyQualifiedHomeserverName(hsName string) (string, error) {
+func (d *Deployment) GetFullyQualifiedHomeserverName(t ct.TestLike, hsName string) string {
 	_, ok := d.HS[hsName]
 	if !ok {
-		return "", fmt.Errorf("Deployment.GetFullyQualifiedHomeserverName - HS name '%s' not found", hsName)
+		ct.Fatalf(t, "Deployment.GetFullyQualifiedHomeserverName - HS name '%s' not found", hsName)
 	}
 	// We have network aliases for each Docker container that will resolve the `hsName` to
 	// the container.
-	return hsName, nil
+	return hsName
 }
 
 // DestroyAtCleanup destroys the entire deployment. It should be called at cleanup time for dirty
--- a/internal/docker/deployment.go
+++ b/internal/docker/deployment.go
@@ -57,6 +57,16 @@ func (hsDep *HomeserverDeployment) SetEndpoints(baseURL string, fedBaseURL strin
 	}
 }
 
+func (d *Deployment) GetFullyQualifiedHomeserverName(hsName string) (string, error) {
+	_, ok := d.HS[hsName]
+	if !ok {
+		return "", fmt.Errorf("Deployment.GetFullyQualifiedHomeserverName - HS name '%s' not found", hsName)
+	}
+	// We have network aliases for each Docker container that will resolve the `hsName` to
+	// the container.
+	return hsName, nil
+}
+
 // DestroyAtCleanup destroys the entire deployment. It should be called at cleanup time for dirty
 // deployments only. Handles configuration options for things which should run at container destroy
 // time, like post-run scripts and printing logs.
--- a/internal/docker/deployment.go
+++ b/internal/docker/deployment.go
@@ -57,14 +57,14 @@ func (hsDep *HomeserverDeployment) SetEndpoints(baseURL string, fedBaseURL strin
 	}
 }
 
-func (d *Deployment) GetFullyQualifiedHomeserverName(t ct.TestLike, hsName string) string {
+func (d *Deployment) GetFullyQualifiedHomeserverName(hsName string) (string, error) {
 	_, ok := d.HS[hsName]
 	if !ok {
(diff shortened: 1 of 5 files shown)


test_main.go
```diff
@@ -6,32 +6,40 @@ import (
...
-	cleanup          func()
-	customDeployment func(numServers int) Deployment
+	// Args:
+	// - We pass in the Complement config (`testPackage.Config`) so the deployer can inspect
+	// `DebugLoggingEnabled`, `SpawnHSTimeout`, `PackageNamespace` etc.
+	cleanup func(config *config.Complement)
+	// Args:
+	// - We pass in `t` as there needs to be a way to handle an error in the custom deployer.
+	// - We pass in the Complement config (`testPackage.Config`) so the deployer can inspect
+	// `DebugLoggingEnabled`, `SpawnHSTimeout`, `PackageNamespace`, etc.
```
MadLittleMods, line 26: Another use case for needing `config.Complement`  is needing access to `CACertificate`/`CAPrivateKey` since they won't magically be in your environment (normally mounted under `/complement/ca/ca.crt`, etc)

internal/docker/deployment.go
```diff
@@ -57,6 +57,16 @@ func (hsDep *HomeserverDeployment) SetEndpoints(baseURL string, fedBaseURL strin
...
+func (d *Deployment) GetFullyQualifiedHomeserverName(t ct.TestLike, hsName string) string {
+	_, ok := d.HS[hsName]
+	if !ok {
+		ct.Fatalf(t, "Deployment.GetFullyQualifiedHomeserverName - HS name '%s' not found", hsName)
+	}
+	// We have network aliases for each Docker container that will resolve the `hsName` to
+	// the container.
+	return hsName
+}
+
```
devonh, line 69: I don't understand how this works.
It looks like we pass in `hsName`, only to turn around and return it right back?

devonh: Oh I see. That is exactly how it is intended to work.
It's other custom deployments that may need to modify this API which makes this necessary.
For base complement, this is simply a passthrough.

devonh:
Do you know what those CI failures are about? The error doesn't seem related to these changes.

internal/docker/deployment.go
```diff
@@ -57,6 +57,16 @@ func (hsDep *HomeserverDeployment) SetEndpoints(baseURL string, fedBaseURL strin
 	}
```
MadLittleMods, line 57: > Do you know what those CI failures are about? The error doesn't seem related to these changes.
>
> *@devonh, https://github.com/matrix-org/complement/pull/778#issuecomment-2881802935*

No idea :shrug:

devonh: Hmm seems like we may benefit from changing any use of `client.NewEnvClient` to 
```go
cli, err := client.NewClientWithOpts(
	client.FromEnv,
	client.WithAPIVersionNegotiation(),
)
```
https://pkg.go.dev/github.com/docker/docker/client#NewClientWithOpts

I think that may avoid this intermittent sounding issue with docker versioning altogether.

devonh: I think there are 2 locations of this, in builder.go & deployer.go

MadLittleMods: :-1: Was already on it (changed both places)

MadLittleMods: CI is green :white_check_mark: 

Great shout on the fix @devonh!

devonh (approved):

test_package.go
```diff
@@ -20,6 +20,13 @@ import (
 
 // Deployment provides a way for tests to interact with a set of homeservers.
 type Deployment interface {
+	// Returns the resolvable server name (host) of a homeserver given its short alias
+	// (e.g., "hs1", "hs2").
+	//
+	// In the case of the standard Docker deployment, this will be the same `hs1`, `hs2`
+	// but may be different for other custom deployments (ex.
+	// `shardDeployment1.GetFullyQualifiedHomeserverName(t, "hs1")` -> `hs1.shard1:8081`).
+	GetFullyQualifiedHomeserverName(t ct.TestLike, hsName string) string
```
MadLittleMods, line 29: @kegsay Just to double-check with the Complement taste-maker, do these changes make sense to you?

Are you okay with this breaking change to the `complement.Deployment` interface?

Are you okay with the breaking changes to `complement.WithDeployment(...)` / `complement.WithCleanup(...)` changes?

kegsay: I'm happy with adding more params to `With...` options.

I'm unhappy with adding `GetFullyQualifiedHomeserverName`. See https://github.com/matrix-org/complement/pull/778/files#r2091384008

internal/docker/deployment.go
MadLittleMods: Ugh, I left out the use case in any explanation here. 

`deployment.GetFullyQualifiedHomeserverName(...)` is useful for situations like the following where you have to specify the via servers in a federated join request during a test:

```go
alice.MustJoinRoom(t, bobRoomID, []string{
	deployment.GetFullyQualifiedHomeserverName(t, "hs2"),
})
```

We don't have access to the `CustomDeployment` type in the tests since  `complement.Deploy(...)` returns the generic `complement.Deployment` interface. I could just cheekily cast it to my own `CustomDeployment` for my own tests but then all of the other Complement tests aren't compatible.

I guess I also left out refactoring all of the existing Complement tests to utilize this for specifying via servers. Might be better to do this in a follow-up PR anyway for more understandable diffs.

Sorry for the missing pieces here :bow: (updated the PR description as well)

I'm sure there are other ways to solve this situation so feel free to suggest something better!

---

`d.HS[hsName].BaseURL` would be the host-accessible URL of the homeserver.

This utility is useful for the federation to access each other, not host access.

kegsay: Right, thanks for the context. I still don't grok why this needs core Complement changes. What is stopping you from having a function like `GetFullyQualifiedHomeserverName(t, deployment "hs2")` where the impl of that function is what you would be putting as the impl for the `Deployment` interface version of the same function? Why do you need to tie this to the `Deployment` directly?

kegsay: Update the interface and test call sites in a separate PR please.

test_package.go
MadLittleMods: Thanks for the review @kegsay!

The `GetFullyQualifiedHomeserverName` changes have been split off to https://github.com/matrix-org/complement/pull/780

MadLittleMods:
Thanks for the review @devonh and @kegsay 🦉

PR #780, opened 2025-05-16
Add `deployment.GetFullyQualifiedHomeserverName(hsName)`

MadLittleMods:
*Split off from https://github.com/matrix-org/complement/pull/778 per [discussion](https://github.com/matrix-org/complement/pull/778#discussion_r2091384008),*

Spawning from a real use-case with a custom `Deployment`/`Deployer` (Element-internal).

Introduce `complement.Deployment.GetFullyQualifiedHomeserverName(hsName)` to allow the per-deployment short homeserver aliases like `hs1` to be mapped to something else that's resolvable in your custom deployments context. Example: `hs1` -> `hs1.shard1:8081`.

This is useful for situations like the following where you have to specify the via servers in a federated join request during a test:

```go
alice.MustJoinRoom(t, bobRoomID, []string{
	deployment.GetFullyQualifiedHomeserverName(t, "hs2"),
})
```

### Todo

 - [ ] Update tests to use `GetFullyQualifiedHomeserverName(...)`

### Pull Request Checklist


client/client.go
```diff
@@ -139,7 +140,11 @@ func (c *CSAPI) CreateRoom(t ct.TestLike, body map[string]interface{}) *http.Res
 }
 
 // MustJoinRoom joins the room ID or alias given, else fails the test. Returns the room ID.
-func (c *CSAPI) MustJoinRoom(t ct.TestLike, roomIDOrAlias string, serverNames []string) string {
+//
+// Args:
+//   - `serverNames`: The list of servers to attempt to join the room through.
+//     These should be a resolvable addresses within the deployment network.
+func (c *CSAPI) MustJoinRoom(t ct.TestLike, roomIDOrAlias string, serverNames []spec.ServerName) string {
```
MadLittleMods, line 147: Basically, wherever we expect people to use `deployment.GetFullyQualifiedHomeserverName(t, hsName)`, I've updated these function signatures to accept `spec.ServerName` instead of just plain strings.

I think this is also more semantically correct for the places that this needs to be a resolvable address in the federation.

federation/handle.go
```diff
@@ -117,7 +117,7 @@ func MakeRespMakeKnock(s *Server, room *ServerRoom, userID string) (resp fclient
 // the current server is returned to the joining server.
 func SendJoinRequestsHandler(s *Server, w http.ResponseWriter, req *http.Request, expectPartialState bool, omitServersInRoom bool) {
 	fedReq, errResp := fclient.VerifyHTTPRequest(
-		req, time.Now(), spec.ServerName(s.serverName), nil, s.keyRing,
```
MadLittleMods, line 120: Since we've converted some of these types to `spec.ServerName`, we no longer need the conversion.

federation/server_room.go
```diff
@@ -498,21 +498,27 @@ func (i *ServerRoomImplDefault) GenerateSendJoinResponse(room *ServerRoom, s *Se
...
 	// return state and auth chain
 	return fclient.RespSendJoin{
-		Origin:         spec.ServerName(s.serverName),
+		Origin:         s.serverName,
 		AuthEvents:     gomatrixserverlib.NewEventJSONsFromEvents(authEvents),
 		StateEvents:    gomatrixserverlib.NewEventJSONsFromEvents(stateEvents),
 		MembersOmitted: expectPartialState,
-		ServersInRoom:  serversInRoom,
+		// TODO: It feels like `ServersInRoom` should be `[]spec.ServerName` instead of `[]string`
+		ServersInRoom: serversInRoomStrings,
```
MadLittleMods, line 522: This feels like a [`gomatrixserverlib` problem](https://github.com/matrix-org/gomatrixserverlib/blob/479c9abf1ca5d5c256ebea5b1d8f6a5c7b69f94d/fclient/federationtypes.go#L279-L280).

(not going to fix in this PR)

tests/federation_room_send_test.go
```diff
@@ -63,7 +64,7 @@ func TestOutboundFederationSend(t *testing.T) {
 	roomAlias := srv.MakeAliasMapping("flibble", serverRoom.RoomID)
 
 	// the local homeserver joins the room
-	alice.MustJoinRoom(t, roomAlias, []string{deployment.GetConfig().HostnameRunningComplement})
+	alice.MustJoinRoom(t, roomAlias, []spec.ServerName{srv.ServerName()})
```
MadLittleMods, line 67: As far as I can tell, this was just a mistake from before. Unclear how it worked before since it's missing the randomly assigned port.

test_package.go
```diff
@@ -15,11 +15,19 @@ import (
...
 
 // Deployment provides a way for tests to interact with a set of homeservers.
 type Deployment interface {
+	// Returns the resolvable server name (host) of a homeserver given its short alias
+	// (e.g., "hs1", "hs2").
+	//
+	// In the case of the standard Docker deployment, this will be the same `hs1`, `hs2`
+	// but may be different for other custom deployments (ex.
+	// `shardDeployment1.GetFullyQualifiedHomeserverName(t, "hs1")` -> `hs1.shard1:8081`).
+	GetFullyQualifiedHomeserverName(t ct.TestLike, hsName string) spec.ServerName
```
MadLittleMods, line 30: The main change of this PR is here. It's a breaking change to the `complement.Deployment` interface

The rest of the diff is essentially just updating to use this new utility.

internal/docker/deployment.go
```diff
@@ -57,6 +58,16 @@ func (hsDep *HomeserverDeployment) SetEndpoints(baseURL string, fedBaseURL strin
...
 
+func (d *Deployment) GetFullyQualifiedHomeserverName(t ct.TestLike, hsName string) spec.ServerName {
+	_, ok := d.HS[hsName]
+	if !ok {
+		ct.Fatalf(t, "Deployment.GetFullyQualifiedHomeserverName - HS name '%s' not found", hsName)
+	}
+	// We have network aliases for each Docker container that will resolve the `hsName` to
+	// the container.
+	return spec.ServerName(hsName)
+}
```
MadLittleMods, line 69: This is the implementation for the new `GetFullyQualifiedHomeserverName` method that we added to the `complement.Deployment` interface.

tests/csapi/apidoc_presence_test.go
```diff
@@ -27,7 +28,9 @@ func TestPresence(t *testing.T) {
 
 	// to share presence alice and bob must be in a shared room
 	roomID := alice.MustCreateRoom(t, map[string]interface{}{"preset": "public_chat"})
-	bob.MustJoinRoom(t, roomID, []string{"hs1"})
+	bob.MustJoinRoom(t, roomID, []spec.ServerName{
+		deployment.GetFullyQualifiedHomeserverName(t, "hs1"),
```
MadLittleMods, line 32: For the reviewer: I've tried to be thorough in updating everything on this list from the PR description. This would be the main thing to think about. Are there other spots that we need to use `GetFullyQualifiedHomeserverName(...)` instead of the hard-coded `hs1` values.

>  - [x] Update tests to use `GetFullyQualifiedHomeserverName(...)`
>      - [x] Join room (`MustJoinRoom`, `JoinRoom`)
>      - [x] Knock room (`mustKnockOnRoomSynced`, `knockOnRoomWithStatus`)
>      - [x] `srv.MustJoinRoom`, `srv.MustLeaveRoom`, `srv.MustSendTransaction`
>      - [x] `FederationClient` -> `fedClient.MakeJoin`, `fedClient.SendJoin`, etc
>      - [x] `fclient`, `fclient.NewFederationRequest`
>      - [x] `m.space.child` `via`
>      - [x] `m.space.parent` `via`
>      - [x] `m.room.join_rules` `restricted` `via`
>      - [x] `gomatrixserverlib.EDU` `Destination`

I've also reviewed the diff itself to ensure that I didn't accidentally swap `hs1` for `hs2` somewhere.

internal/docker/deployment.go
```diff
@@ -57,6 +58,16 @@ func (hsDep *HomeserverDeployment) SetEndpoints(baseURL string, fedBaseURL strin
...
+
```
MadLittleMods, line 70: Currently, I don't think the built-in Complement `Deployment` implementation supports multiple `Deployments` at the same time (`hs1`, `hs2` would conflict between them). Since one of the goals of this PR is to unlock that functionality for custom `Deployment`'s, it could make some sense to also refactor and update that here as well.

I'd rather leave it as-is until we need it or at-least do this in a follow-up PR.

See the PR description for more context on multiple `Deployment`.

kegsay: The PR description doesn't give more context on why multiple `Deployment` is desirable:

> But imagine a case where we have multiple Deployment and we want the homeservers to communicate with each other. 

_Why_ would you do this? The `Deployment` is meant to encapsulate an _entire_ deployment, all servers and network links between them. I don't understand the rationale.

kegsay (approved):
The addition of FQDN LGTM.

I'm not particularly convinced about the need for >1 `Deployment` at the same time, this isn't how Complement was designed to work. Given test authors have the power to control their `Deplyoment` impl, it makes sense to add FQDN support and remove the assumption that "hs1" is routable in their deployment.

MadLittleMods:
Thanks for the review @kegsay! Curious to get your further opinions/options/guidance on how best to address our [multi-`Deployment` use case](https://github.com/matrix-org/complement/pull/780#discussion_r2093701255).

tests/csapi/apidoc_presence_test.go
MadLittleMods: Added `m.room.server_acl` to the list and is being updated in https://github.com/matrix-org/complement/pull/804

PR #781, opened 2025-05-26
Fix Complement not using `HSPortBindingIP` (`127.0.0.1`) for the homeserver `BaseURL`

MadLittleMods:
Fix Complement not using `config.HSPortBindingIP` (`127.0.0.1`) for the homeserver `BaseURL`.

Regressed in https://github.com/matrix-org/complement/pull/776 which started using the Docker default `HostIP` (`0.0.0.0`) because we removed the specific `HSPortBindingIP` port bindings for `8008` and `8448` (see "Root cause" section below).

### Problem

This caused downstream issues in some of our out-of-repo Complement tests (Element internal) which stress some SSO/OIDC login flows with Synapse. 

Before https://github.com/matrix-org/complement/pull/776, [`SsoRedirectServlet`](https://github.com/element-hq/synapse/blob/33ba8860c43d4770ea119a09a4fcbbf366f3b32e/synapse/rest/client/login.py#L645-L683) received a request like this `http://127.0.0.1:8008/_matrix/client/v3/login/sso/redirect/oidc-test_idp?redirectUrl=http%3A%2F%2Fapp.my-fake-client.io%2Fclient-callback`. But now it's seeing `http://0.0.0.0:8008/_matrix/client/v3/login/sso/redirect/oidc-test_idp?redirectUrl=http%3A%2F%2Fapp.my-fake-client.io%2Fclient-callback` and tries to redirect to the canonical `http://127.0.0.1:8008/` address which would then go to the step we expect in the tests (authorization endpoint).

Basically, the `host` header in the request used to be set to `127.0.0.1:8008` but now it's `0.0.0.0:8008`. Synapse wants to use the canonical URL so the cookies are available so it redirects you to the canonical `public_baseurl` version first. Relevant Synapse code that cares about using the canonical URL to get cookies back -> [`SsoRedirectServlet`](https://github.com/element-hq/synapse/blob/33ba8860c43d4770ea119a09a4fcbbf366f3b32e/synapse/rest/client/login.py#L645-L683)

We could alternatively, update our out-of-repo tests to account for this extra redirect but it seems more appropriate to update Complement to use the the configured `HSPortBindingIP` as expected.

### Root cause

The host name used in the requests during the Complement tests is from the `client.BaseURL` which is sourced from the [Docker container port mapping/bindings](https://github.com/matrix-org/complement/blob/28a09014afd1f9c75a3549b4cd9d8fc480ba1149/internal/docker/deployer.go#L521) in Complement.

In https://github.com/matrix-org/complement/pull/776, we removed the explicit port bindings for `8008` with `HSPortBindingIP` ([`127.0.0.1`](https://github.com/matrix-org/complement/blob/28a09014afd1f9c75a3549b4cd9d8fc480ba1149/config/config.go#L180)) so now it uses Docker's default `0.0.0.0`.

```patch
diff --git a/internal/docker/deployer.go b/internal/docker/deployer.go
index cdcf51a5..a186b30d 100644
--- a/internal/docker/deployer.go
+++ b/internal/docker/deployer.go
@@ -382,20 +380,8 @@ func deployImage(
 	}, &container.HostConfig{
 		CapAdd:          []string{"NET_ADMIN"}, // TODO : this should be some sort of option
 		PublishAllPorts: true,
-		PortBindings: nat.PortMap{
-			nat.Port("8008/tcp"): []nat.PortBinding{
-				{
-					HostIP: cfg.HSPortBindingIP,
-				},
-			},
-			nat.Port("8448/tcp"): []nat.PortBinding{
-				{
-					HostIP: cfg.HSPortBindingIP,
-				},
-			},
-		},
```

#### Locally link Complement package into your tests

See https://thewebivore.com/using-replace-in-go-mod-to-point-to-your-local-module/

So you can edit the Complement source code and use it in the tests.

 1. Add the following to `complement/go.mod`
    ```
    // TODO: Remove as this is only for local development
    replace github.com/matrix-org/complement => /home/eric/Documents/github/element/complement
    ```
 1. `cd complement && go mod tidy && cd ..`
 1. Run Complement how you normally would

### Pull Request Checklist

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -106,6 +106,7 @@ type Complement struct {
 	// disable this behaviour being added later, once this has stablised.
 	EnableDirtyRuns bool
 
+	// The hostname that will be used to bind the homeserver ports to, e.g. `127.0.0.1`
 	HSPortBindingIP string
 
 	// Name: COMPLEMENT_POST_TEST_SCRIPT
--- a/config/config.go
+++ b/config/config.go
@@ -106,14 +106,13 @@ type Complement struct {
 	// disable this behaviour being added later, once this has stablised.
 	EnableDirtyRuns bool
 
-	// The IP that will be used to bind the homeserver ports to, e.g. `127.0.0.1`  (only
-	// allow localhost access), `0.0.0.0` (bind to all available interfaces).
+	// The IP that is used to connect to the running homeserver from the host.
 	//
 	// For Complement tests, this is always configured as `127.0.0.1` but can be
 	// overridden by homerunner to allow binding to a different IP address
 	// (`HOMERUNNER_HS_PORTBINDING_IP`).
 	//
-	// This field is also used for the host-accessible homeserver URLs (as the hostname)
+	// This field is used for the host-accessible homeserver URLs (as the hostname)
 	// so clients in your tests can access the homeserver.
 	HSPortBindingIP string
 
--- a/config/config.go
+++ b/config/config.go
@@ -106,7 +106,15 @@ type Complement struct {
 	// disable this behaviour being added later, once this has stablised.
 	EnableDirtyRuns bool
 
-	// The hostname that will be used to bind the homeserver ports to, e.g. `127.0.0.1`
+	// The IP that will be used to bind the homeserver ports to, e.g. `127.0.0.1`  (only
+	// allow localhost access), `0.0.0.0` (bind to all available interfaces).
+	//
+	// For Complement tests, this is always configured as `127.0.0.1` but can be
+	// overridden by homerunner to allow binding to a different IP address
+	// (`HOMERUNNER_HS_PORTBINDING_IP`).
+	//
+	// This field is also used for the host-accessible homeserver URLs (as the hostname)
+	// so clients in your tests can access the homeserver.
 	HSPortBindingIP string
 
 	// Name: COMPLEMENT_POST_TEST_SCRIPT
--- a/internal/docker/builder.go
+++ b/internal/docker/builder.go
@@ -539,27 +539,46 @@ func printPortBindingsOfAllComplementContainers(docker *client.Client, contextSt
 	log.Printf("=============== %s : END ALL COMPLEMENT DOCKER PORT BINDINGS ===============\n\n\n", contextStr)
 }
 
-func endpoints(p nat.PortMap, csPort, ssPort int) (baseURL, fedBaseURL string, err error) {
-	csapiPort := fmt.Sprintf("%d/tcp", csPort)
-	csapiPortInfo, ok := p[nat.Port(csapiPort)]
-	if !ok {
-		return "", "", fmt.Errorf("port %s not exposed - exposed ports: %v", csapiPort, p)
+func endpoints(p nat.PortMap, hsPortBindingIP string, csPort, ssPort int) (baseURL, fedBaseURL string, err error) {
+	csapiPortBinding, err := findPortBinding(p, hsPortBindingIP, csPort)
+	if err != nil {
+		return "", "", fmt.Errorf("Problem finding CS API port: %s", err)
 	}
-	if len(csapiPortInfo) == 0 {
-		return "", "", fmt.Errorf("port %s exposed with not mapped port: %+v", csapiPort, p)
-	}
-	baseURL = fmt.Sprintf("http://"+csapiPortInfo[0].HostIP+":%s", csapiPortInfo[0].HostPort)
+	baseURL = fmt.Sprintf("http://"+csapiPortBinding.HostIP+":%s", csapiPortBinding.HostPort)
 
-	ssapiPort := fmt.Sprintf("%d/tcp", ssPort)
-	ssapiPortInfo, ok := p[nat.Port(ssapiPort)]
-	if !ok {
-		return "", "", fmt.Errorf("port %s not exposed - exposed ports: %v", ssapiPort, p)
+	ssapiPortBinding, err := findPortBinding(p, hsPortBindingIP, ssPort)
+	if err != nil {
+		return "", "", fmt.Errorf("Problem finding SS API port: %s", err)
 	}
-	if len(ssapiPortInfo) == 0 {
-		return "", "", fmt.Errorf("port %s exposed with not mapped port: %+v", ssapiPort, p)
-	}
-	fedBaseURL = fmt.Sprintf("https://"+csapiPortInfo[0].HostIP+":%s", ssapiPortInfo[0].HostPort)
+	fedBaseURL = fmt.Sprintf("https://"+ssapiPortBinding.HostIP+":%s", ssapiPortBinding.HostPort)
 	return
+}
+
+// Find a matching port binding for the given host/port in the nat.PortMap.
+func findPortBinding(p nat.PortMap, hsPortBindingIP string, port int) (portBinding nat.PortBinding, err error) {
+	portString := fmt.Sprintf("%d/tcp", port)
+	portBindings, ok := p[nat.Port(portString)]
+	if !ok {
+		return nat.PortBinding{}, fmt.Errorf("port %s not exposed - exposed ports: %v", portString, p)
+	}
+	if len(portBindings) == 0 {
+		return nat.PortBinding{}, fmt.Errorf("port %s exposed with not mapped port: %+v", portString, p)
+	}
+
+	for _, pb := range portBindings {
+		if pb.HostIP == hsPortBindingIP {
+			return pb, nil
+		} else if pb.HostIP == "0.0.0.0" {
+			// `0.0.0.0` means "all interfaces", so we can assume that this will be listening
+			// for connections from `hsPortBindingIP` as well.
+			return nat.PortBinding{
+				HostIP:   hsPortBindingIP,
+				HostPort: pb.HostPort,
+			}, nil
+		}
+	}
+
+	return portBindings[0], nil
 }
 
 type result struct {
--- a/internal/docker/builder.go
+++ b/internal/docker/builder.go
@@ -539,7 +539,7 @@ func printPortBindingsOfAllComplementContainers(docker *client.Client, contextSt
 	log.Printf("=============== %s : END ALL COMPLEMENT DOCKER PORT BINDINGS ===============\n\n\n", contextStr)
 }
 
-// Transform the homeserver ports into the base URL and federation base URL.
+// endpoints transforms the homeserver ports into the base URL and federation base URL.
 func endpoints(p nat.PortMap, hsPortBindingIP string, csPort, ssPort int) (baseURL, fedBaseURL string, err error) {
 	csapiPortBinding, err := findPortBinding(p, hsPortBindingIP, csPort)
 	if err != nil {
@@ -555,7 +555,7 @@ func endpoints(p nat.PortMap, hsPortBindingIP string, csPort, ssPort int) (baseU
 	return
 }
 
-// Find a matching port binding for the given host/port in the `nat.PortMap`.
+// findPortBinding finds a matching port binding for the given host/port in the `nat.PortMap`.
 //
 // This function will return the first port binding that matches the given host IP. If a
 // `0.0.0.0` binding is found, we will assume that it is listening on all interfaces,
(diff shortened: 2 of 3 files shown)


internal/docker/deployer.go
```diff
@@ -313,9 +313,13 @@ func (d *Deployer) StartServer(hsDep *HomeserverDeployment) error {
 	}
 
 	// Wait for the container to be ready.
-	baseURL, fedBaseURL, err := waitForPorts(ctx, d.Docker, hsDep.ContainerID)
+	err = waitForPorts(ctx, d.Docker, hsDep.ContainerID)
 	if err != nil {
-		return fmt.Errorf("failed to get ports for container %s: %s", hsDep.ContainerID, err)
+		return fmt.Errorf("failed to wait for ports on container %s: %s", hsDep.ContainerID, err)
+	}
+	baseURL, fedBaseURL, err := getHostAccessibleHomeserverUrls(ctx, d.Docker, hsDep.ContainerID, d.config.HSPortBindingIP)
```
MadLittleMods, line 320: To clean things up, I've refactored `waitForPorts(...)` to only do the waiting part as described. We now assemble `BaseURL` and `FedBaseUrl` via `getHostAccessibleHomeserverUrls(...)`

internal/docker/deployer.go
```diff
@@ -504,26 +517,90 @@ func copyToContainer(docker *client.Client, containerID, path string, data []byt
...
+		if inspectionErr, ok := err.(*ContainerInspectionError); ok && inspectionErr.Fatal {
+			// If the error is fatal, we should not retry.
+			return fmt.Errorf("Fatal inspection error: %s", err)
+		}
 	}
-	return baseURL, fedBaseURL, nil
+	return nil
+}
+
+type ContainerInspectionError struct {
```
richvdh, line 576: this should probably be private (`containerInspectionError`) if it's only for use by the private function below.

internal/docker/deployer.go
```diff
@@ -504,26 +517,90 @@ func copyToContainer(docker *client.Client, containerID, path string, data []byt
...
+	Fatal bool
+}
+
+func (e *ContainerInspectionError) Error() string { return e.msg }
+
+func inspectPortsOnContainer(
+	ctx context.Context,
+	docker *client.Client,
+	containerID string,
+) (inspectResponse container.InspectResponse, err error) {
```
richvdh, line 590: if this always returns a `ContainerInspectionError`, could we put that in the return type?

Alternatively/additionally, maybe some documentation to say under which circumstances it returns what sort of error?

internal/docker/deployer.go
```diff
@@ -504,26 +517,90 @@ func copyToContainer(docker *client.Client, containerID, path string, data []byt
...
 	for time.Since(inspectStartTime) < time.Second {
-		inspect, err = docker.ContainerInspect(ctx, containerID)
-		if err != nil {
-			return "", "", err
-		}
-		if inspect.State != nil && !inspect.State.Running {
-			// the container exited, bail out with a container ID for logs
-			return "", "", fmt.Errorf("container is not running, state=%v", inspect.State.Status)
-		}
-		baseURL, fedBaseURL, err = endpoints(inspect.NetworkSettings.Ports, 8008, 8448)
```
richvdh, line 521: it looks like this used to carry on around the loop when `endpoints` returned an error; obviously now that the `endpoints` call is moved to `getHostAccessibleHomeserverUrls`, that isn't happening.

That seems like it may be a problem. (Is it really the case that as soon as `ContainerInspect` returns non-error, the ports are ready? If so, maybe this function should be called `waitForContainer` or something rather than `waitForPorts`)

internal/docker/deployer.go
```diff
@@ -504,26 +517,90 @@ func copyToContainer(docker *client.Client, containerID, path string, data []byt
...
+
+func getHostAccessibleHomeserverUrls(ctx context.Context, docker *client.Client, containerID string, hsPortBindingIP string) (baseURL string, fedBaseURL string, err error) {
+	inspectResponse, err := inspectPortsOnContainer(ctx, docker, containerID)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to inspect ports: %w", err)
+	}
+
+	baseURL, fedBaseURL, err = endpoints(inspectResponse.NetworkSettings.Ports, hsPortBindingIP, 8008, 8448)
+
+	// Sanity check that the URL's match the expected binding hostname. It's important
```
richvdh:
```suggestion
	// Sanity check that the URLs match the expected binding hostname. It's important
```

internal/docker/deployer.go
```diff
@@ -504,26 +517,90 @@ func copyToContainer(docker *client.Client, containerID, path string, data []byt
...
+		return fmt.Errorf("failed to parse URL %s: %s", inputUrl, err)
+	}
+	if parsedUrl.Hostname() != expectedHostname {
+		return fmt.Errorf("expected hostname %s in URL %s, got %s", expectedHostname, inputUrl, parsedUrl.Hostname())
+	}
+
+	return nil
+}
+
+func getHostAccessibleHomeserverUrls(ctx context.Context, docker *client.Client, containerID string, hsPortBindingIP string) (baseURL string, fedBaseURL string, err error) {
```
richvdh, line 534: convention in go is URL, rather than Url

```suggestion
func getHostAccessibleHomeserverURLs(ctx context.Context, docker *client.Client, containerID string, hsPortBindingIP string) (baseURL string, fedBaseURL string, err error) {
```

internal/docker/deployer.go
```diff
@@ -504,26 +517,90 @@ func copyToContainer(docker *client.Client, containerID, path string, data []byt
...
+	inspectResponse, err := inspectPortsOnContainer(ctx, docker, containerID)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to inspect ports: %w", err)
+	}
+
+	baseURL, fedBaseURL, err = endpoints(inspectResponse.NetworkSettings.Ports, hsPortBindingIP, 8008, 8448)
+
+	// Sanity check that the URL's match the expected binding hostname. It's important
+	// that we use the canonical publically accessible hostname for the homeserver as ...
```
richvdh: this seems to have got lost in the middle, somewhere?

also, "publicly".

config/config.go
```diff
@@ -106,6 +106,7 @@ type Complement struct {
 	// disable this behaviour being added later, once this has stablised.
 	EnableDirtyRuns bool
 
+	// The hostname that will be used to bind the homeserver ports to, e.g. `127.0.0.1`
```
richvdh, line 109: is it really a hostname? or is it an IP address?

internal/docker/builder.go
```diff
@@ -539,27 +539,46 @@ func printPortBindingsOfAllComplementContainers(docker *client.Client, contextSt
...
+			return nat.PortBinding{
+				HostIP:   hsPortBindingIP,
+				HostPort: pb.HostPort,
+			}, nil
+		}
 	}
-	fedBaseURL = fmt.Sprintf("https://"+csapiPortInfo[0].HostIP+":%s", ssapiPortInfo[0].HostPort)
-	return
+
+	return portBindings[0], nil
```
richvdh, line 594: we're falling back to the first binding here, whether or not it matches? Is that wise? It could do with a comment, at least.

internal/docker/builder.go
```diff
@@ -539,6 +539,7 @@ func printPortBindingsOfAllComplementContainers(docker *client.Client, contextSt
 	log.Printf("=============== %s : END ALL COMPLEMENT DOCKER PORT BINDINGS ===============\n\n\n", contextStr)
 }
 
+// Transform the homeserver ports into the base URL and federation base URL.
```
richvdh, line 542: It's worth noting that the convention for doc-comments in go is to start with the name of the thing you're describing:

```
// endpoints transforms the homeserver ports into the base URL and federation base URL.
```

I'll not insist you change everything, but if you could bear it in mind when you're fixing up doc-comments, that would be good.

internal/docker/deployer.go
```diff
@@ -529,6 +529,8 @@ func assertHostnameEqual(inputUrl string, expectedHostname string) error {
 	return nil
 }
 
+// Returns URL's that are accessible from the host machine (outside the container) for
```
richvdh, line 532:
```suggestion
// Returns URLs that are accessible from the host machine (outside the container) for
```

config/config.go
```diff
@@ -106,6 +106,7 @@ type Complement struct {
...
 	HSPortBindingIP string
```
richvdh, line 110: More pertinently: this doesn't seem to reflect the actual behaviour you've introduced for HSPortBindingIP here. AFAICT in your change, it doesn't change the address we bind to, just the way we build the URLs that we use for reaching Synapse.

Maybe that's fine. TBH I couldn't find any documentation for what `HSPortBindingIP` is actually supposed to do, which is why I gave up and added a TODO over in #779. (It seems to only be set for homerunner? Which I have never tried out, and only have the haziest idea of its purpose. If you have got this far, maybe you know more than me.)

In any case, if we're going to just use it for the URLs, the comment had better say that rather than anything about binding.

internal/docker/deployer.go
MadLittleMods: I'm no Go expert, but this seems to be the go-to pattern, see https://go.dev/blog/error-handling-and-go (returning generic `error` interface and using type assertion).

I initially had it return `ContainerInspectionError` but kept seeing opposite examples and it's a bit weird with pointer vs object.

I'll add some docs :fast_forward:

config/config.go
MadLittleMods: > More pertinently: this doesn't seem to reflect the actual behaviour you've introduced for HSPortBindingIP here. AFAICT in your change, it doesn't change the address we bind to, just the way we build the URLs that we use for reaching Synapse.

Ideally, we would use it for binding but it's just not compatible with the way we `PublishAllPorts` now. I looked for some config while making this PR but the `HostIp` config is only available with the manual port map. With `PublishAllPorts`, we just bind to all interfaces and just look for a binding match to `HSPortBindingIP`.

Using it directly for binding may still be relevant for other custom `Deployer` but the key thing is that it's at-least listening on the specified IP.

I can add a comment to clarify further in a few places.

internal/docker/deployer.go
MadLittleMods: Looks like you commented on an outdated version of the diff

internal/docker/deployer.go
MadLittleMods: Made it private :shrug:

Staring at lower-case types :grimacing:

internal/docker/builder.go
MadLittleMods: Good catch, this was just an oversight. Updated to return an error :white_check_mark:

config/config.go
```diff
@@ -106,7 +106,15 @@ type Complement struct {
...
-	// The hostname that will be used to bind the homeserver ports to, e.g. `127.0.0.1`
+	// The IP that will be used to bind the homeserver ports to, e.g. `127.0.0.1`  (only
+	// allow localhost access), `0.0.0.0` (bind to all available interfaces).
+	//
+	// For Complement tests, this is always configured as `127.0.0.1` but can be
+	// overridden by homerunner to allow binding to a different IP address
+	// (`HOMERUNNER_HS_PORTBINDING_IP`).
+	//
+	// This field is also used for the host-accessible homeserver URLs (as the hostname)
+	// so clients in your tests can access the homeserver.
```
richvdh, line 117: As long as it's not actually used for binding, I think this is misleading.

```suggestion
	// The IP that is used to connect to the running homeserver from the host.
	//
	// For Complement tests, this is always configured as `127.0.0.1` but can be
	// overridden by homerunner to allow binding to a different IP address
	// (`HOMERUNNER_HS_PORTBINDING_IP`).
	//
	// This field is used for the host-accessible homeserver URLs (as the hostname)
	// so clients in your tests can access the homeserver.
```

internal/docker/deployer.go
```diff
@@ -555,49 +564,57 @@ func getHostAccessibleHomeserverUrls(ctx context.Context, docker *client.Client,
...
 	Fatal bool
 }
 
-func (e *ContainerInspectionError) Error() string { return e.msg }
+func (e *containerInspectionError) Error() string { return e.msg }
 
-func inspectPortsOnContainer(
+// inspectContainer inspects the container with the given ID and returns response.
+//
+// Returns a `containerInspectionError` representing the underlying error and indicates
```
richvdh, line 601:
```suggestion
// On failure, returns a `containerInspectionError` representing the underlying error and indicates
```

richvdh (approved):
LGTM otherwise

config/config.go
MadLittleMods: Updated :white_check_mark: This sounds accurate enough to move on :fast_forward:

We could say "Historically, used for ..."

MadLittleMods:
Thanks for the review @richvdh 🦘

PR #782, opened 2025-05-29
Add tests for MSC4155

Half-Shot:
For https://github.com/element-hq/synapse/pull/18288
This implements tests for https://github.com/matrix-org/matrix-spec-proposals/pull/4155

Tests will not pass as I do not have rights to push to element-hq/Synapse, so it won't be able to find my branch.

### Pull Request Checklist


tests/msc4155/invite_filter_test.go
```diff
@@ -0,0 +1,203 @@
...
+	})
+	t.Run("Can block a single user", func(t *testing.T) {
+		mustSetInviteConfig(t, alice, InviteFilterConfig{
+			BlockedUsers: []string{bob.UserID},
+		})
+		roomID := bob.MustCreateRoom(t, map[string]interface{}{
+			"preset": "private_chat",
+		})
+		MustInviteRoomAndFail(t, bob, roomID, alice.UserID)
+		MustHaveNoInviteInSyncResponse(t, alice)
```
kegsay, line 64: Asserting a negative is tricky because you need to be really really sure you've applied the input first. In its current form, this is racey because `MustCreateRoom` can 200 OK before all synchrotrons are aware of the room. Therefore, Alice could `/sync` and not see an invite section, not because the invite config is working but because the HS is going slow.

To guard against this, you need to wait until the synchrotron has seen the room. You can do this pretty easily in Complement as this is pretty typical:
```go
bob.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(bob.UserID, roomID))
```
Add this after you have created the room and you should be good. Also note that this form uses `MustSyncUntil` and not `MustSync` because a single-shot `/sync` request could miss the room creation due to it not being on the synchrotrons yet and then be flakey.. but because this is testing a positive (there's an output) you can keep syncing until you see the output, but you can't do this for testing that an invite _doesn't_ appear.

tests/msc4155/invite_filter_test.go
```diff
@@ -0,0 +1,181 @@
...
+		alice.MustSyncUntil(t, client.SyncReq{}, client.SyncInvitedTo(alice.UserID, roomIDAllowed))
+		alice.MustJoinRoom(t, roomIDAllowed, []spec.ServerName{})
+		alice.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(alice.UserID, roomIDAllowed))
+		bob.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(alice.UserID, roomIDAllowed))
+	})
+}
+
+// Tests that a given invite filter config is properly set
+func mustSetInviteConfig(t *testing.T, c *client.CSAPI, cfg InviteFilterConfig) {
+	b, err := json.Marshal(&cfg)
```
anoadragon453, line 139: Could you add a quick comment here explaining why the Marshal/Unmarshal dance is needed (to tranform `InviteFilterConfig` -> `map[string]interface{}` presumably).

Half-Shot:
>Could you add a few more sanity checks

These are done, with the exception of `Escaping a glob works, so you can match a user with ID @*?:example.com`. There isn't a documented way to escape globs in Matrix yet, and I wouldn't want to introduce a test without spec backing it.

PR #787, opened 2025-07-03
Add tests for edge cases of multiple membership changes in sync period

anoadragon453:
Addresses https://github.com/element-hq/synapse/issues/14683 case seen in the real world.

Requires https://github.com/element-hq/synapse/pull/18648 for these tests to pass in Synapse.

### Pull Request Checklist

### DIFF
--- a/tests/csapi/sync_test.go
+++ b/tests/csapi/sync_test.go
@@ -731,6 +731,145 @@ func TestRoomSummary(t *testing.T) {
 	alice.MustSyncUntil(t, client.SyncReq{Since: aliceSince}, client.SyncJoinedTo(bob.UserID, roomID), joinedCheck)
 }
 
+// TestLeaveReinviteSync tests that when a user is kicked and then re-invited,
+// they only see an invite in their sync response, not both an invite and leave event.
+func TestLeaveReinviteSync(t *testing.T) {
+	deployment := complement.Deploy(t, 1)
+	defer deployment.Destroy(t)
+
+	alice := deployment.Register(t, "hs1", helpers.RegistrationOpts{})
+	bob := deployment.Register(t, "hs1", helpers.RegistrationOpts{})
+
+	// 1. Alice creates a room and Bob joins
+	roomID := alice.MustCreateRoom(t, map[string]any{
+		"preset": "public_chat",
+	})
+	bob.MustJoinRoom(t, roomID, nil)
+
+	// 2. Bob does a sync and verifies they see the join
+	bobSince := bob.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(bob.UserID, roomID))
+
+	// 3. Alice kicks Bob from the room and then re-invites them.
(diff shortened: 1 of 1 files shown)


tests/csapi/sync_test.go
```diff
@@ -731,6 +731,145 @@ func TestRoomSummary(t *testing.T) {
...
+	)
+
+	// Wait until Bob is kicked.
+	alice.MustSyncUntil(t, client.SyncReq{}, client.SyncLeftFrom(bob.UserID, roomID))
+
+	// Alice re-invites Bob
+	alice.MustInviteRoom(t, roomID, bob.UserID)
+
+	// 4. Bob does a sync
+	jsonRes, _ := bob.MustSync(t, client.SyncReq{Since: bobSince})
```
MadLittleMods, line 860: This doesn't guarantee that the leave/invite will be federated to Bob's homeserver by the time Bob syncs.

Perhaps we need another `hs2` user in the room to observe first

If we just use `bob.MustSyncUntil(t, client.SyncReq{Since: bobSince}, client.SyncInvitedTo(bob.UserID, roomID))` the leave and invite could be separated in different sync responses as we paginate.

MadLittleMods (changes requested):

tests/csapi/sync_test.go
anoadragon453: Very good point. Another user would work. I'll modify the test to include that.

tests/csapi/sync_test.go
```diff
@@ -731,6 +731,163 @@ func TestRoomSummary(t *testing.T) {
...
+	// For kicking, we need to use the POST /rooms/{roomId}/kick endpoint
+	alice.MustDo(t, "POST", []string{"_matrix", "client", "v3", "rooms", roomID, "kick"},
+		client.WithJSONBody(t, map[string]any{
+			"user_id": bob.UserID,
+		}),
+	)
+
+	// Wait until Bob is kicked.
+	aliceSince := alice.MustSyncUntil(t, client.SyncReq{}, client.SyncLeftFrom(bob.UserID, roomID))
+	charlieSince = charlie.MustSyncUntil(t, client.SyncReq{Since: charlieSince}, client.SyncLeftFrom(bob.UserID, roomID))
```
MadLittleMods, line 865: I don't think we need to worry about getting a `since` token in any of these places execpt for `bob` which we are actually testing incremental sync for.

The other sync's are just asserting pre-conditions.

```suggestion
	charlie.MustSyncUntil(t, client.SyncReq{}, client.SyncLeftFrom(bob.UserID, roomID))
```

MadLittleMods, line 865: Add a comment about why we wait here as well

(needs wrapping)

```suggestion
	// Wait until hs2 sees that Bob has been kicked. If we don't wait, then we have no idea whether `hs2` received the kick event.
	charlieSince = charlie.MustSyncUntil(t, client.SyncReq{Since: charlieSince}, client.SyncLeftFrom(bob.UserID, roomID))
```

PR #788, opened 2025-07-09
Add tests for rich topic handling (previously MSC3765)

Johennes:
Rich topics landed with Matrix 1.15. These tests were broken out from https://github.com/element-hq/synapse/pull/18195/files#r2170245892.

### Pull Request Checklist

- [X] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)

### DIFF
--- a/tests/csapi/apidoc_room_create_test.go
+++ b/tests/csapi/apidoc_room_create_test.go
@@ -86,7 +86,7 @@ func TestRoomCreate(t *testing.T) {
 			// The mime type must be unset or text/plain
 			mime := content.Get("m\\.topic.m\\.text.0.mimetype")
 			if mime.Exists() {
-				must.Equal(t, mime.String(), "text/plain", "no m.topic content block")
+				must.Equal(t, mime.String(), "text/plain", "expected rich topic mimetype to be unset (defaults to text/plain) or explicitly set as text/plain")
 			}
 		})
 		// POST /createRoom makes a room with a topic via initial_state
@@ -143,7 +143,7 @@ func TestRoomCreate(t *testing.T) {
 			// The mime type must be unset or text/plain
 			mime := content.Get("m\\.topic.m\\.text.0.mimetype")
 			if mime.Exists() {
-				must.Equal(t, mime.String(), "text/plain", "no m.topic content block")
+				must.Equal(t, mime.String(), "text/plain", "expected rich topic mimetype to be unset (defaults to text/plain) or explicitly set as text/plain")
 			}
 		})
 		// sytest: POST /createRoom makes a room with a name


tests/csapi/apidoc_room_create_test.go
```diff
@@ -61,6 +61,72 @@ func TestRoomCreate(t *testing.T) {
...
+				"preset": "public_chat",
+			})
+			content := alice.MustGetStateEventContent(t, roomID, "m.room.topic", "")
+			must.MatchGJSON(t, content, match.JSONKeyEqual("topic", "Test Room"))
+
+			// There is no m.topic property
+			must.MatchGJSON(t, content, match.JSONKeyMissing("m\\.topic"))
+		})
+		// POST /createRoom makes a room with a topic via initial_state overwritten by topic
+		t.Run("POST /createRoom makes a room with a topic via initial_state overwritten by topic", func(t *testing.T) {
```
MadLittleMods, line 100: For my own reference, this behavior is spec'ed in the `/createRoom` endpoint, https://spec.matrix.org/v1.14/client-server-api/#post_matrixclientv3createroom


> The server MUST apply the normal state resolution rules when creating the new room, including checking power levels for each event. It MUST apply the events implied by the request in the following order:
>
> [...]
> 6. Events listed in `initial_state`, in the order that they are listed.
> 7. Events implied by `name` and `topic` (`m.room.name` and `m.room.topic` state events).
> [...]

MadLittleMods (approved):

tests/csapi/apidoc_room_create_test.go
```diff
@@ -61,6 +61,72 @@ func TestRoomCreate(t *testing.T) {
 			})
 			content := alice.MustGetStateEventContent(t, roomID, "m.room.topic", "")
 			must.MatchGJSON(t, content, match.JSONKeyEqual("topic", "Test Room"))
+
```
MadLittleMods, line 64: It looks like we have to gate the tests that expect `m.topic` so that Dendrite doesn't run them https://github.com/matrix-org/complement/actions/runs/16165352807/job/45668238300?pr=788#step:8:1971

For this one, I guess we just have to split it out separately.

tests/csapi/apidoc_room_create_rich_topic_test.go
```diff
@@ -0,0 +1,78 @@
+//go:build !dendrite_blacklist
+// +build !dendrite_blacklist
```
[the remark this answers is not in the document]
Johennes: Oh, I missed `runtime.SkipIf`. This is definitely better, yes. Have switched things around accordingly.

tests/csapi/apidoc_room_create_test.go
```diff
@@ -61,6 +62,78 @@ func TestRoomCreate(t *testing.T) {
...
+				match.JSONKeyArrayOfSize("m\\.topic.m\\.text", 1),
+				match.JSONKeyPresent("m\\.topic.m\\.text.0.body"),
+				match.JSONKeyEqual("m\\.topic.m\\.text.0.body", "Test Room"))
+
+			// The mime type must be unset or text/plain
+			mime := content.Get("m\\.topic.m\\.text.0.mimetype")
+			if mime.Exists() {
+				must.Equal(t, mime.String(), "text/plain", "no m.topic content block")
+			}
+		})
```
MadLittleMods, line 80: We still want the first part of the test to run and succeed/fail on Dendrite so I think we still need this bottom part to be separate.

tests/csapi/apidoc_room_create_test.go
```diff
@@ -61,6 +62,78 @@ func TestRoomCreate(t *testing.T) {
...
+						"state_key": "",
+					},
+				},
+				"preset": "public_chat",
+			})
+			content := alice.MustGetStateEventContent(t, roomID, "m.room.topic", "")
+			must.MatchGJSON(t, content, match.JSONKeyEqual("topic", "Test Room"))
+
+			// Rich topics not implemented yet on Dendrite
+			runtime.SkipIf(t, runtime.Dendrite)
```
MadLittleMods, line 124: We should be skipping at the top of this test

MadLittleMods (changes requested):

tests/csapi/apidoc_room_create_test.go
Johennes: Sorry, I had assumed `SkipIf` to only skip the remainder of the test but it looks like it marks the whole test as skipped unless it fails. Have move the statement to the top with e61bcb72e52cec903aad34e102d16d37d1549069.

tests/csapi/apidoc_room_create_test.go
```diff
@@ -62,6 +63,89 @@ func TestRoomCreate(t *testing.T) {
...
+			// The plain text topic is duplicated into m.topic
+			must.MatchGJSON(t, content,
+				match.JSONKeyArrayOfSize("m\\.topic.m\\.text", 1),
+				match.JSONKeyPresent("m\\.topic.m\\.text.0.body"),
+				match.JSONKeyEqual("m\\.topic.m\\.text.0.body", "Test Room"))
+
+			// The mime type must be unset or text/plain
+			mime := content.Get("m\\.topic.m\\.text.0.mimetype")
+			if mime.Exists() {
+				must.Equal(t, mime.String(), "text/plain", "no m.topic content block")
```
MadLittleMods, line 146:
```suggestion
				must.Equal(t, mime.String(), "text/plain", "expected rich topic mimetype to be unset (defaults to text/plain) or explicitly set as text/plain")
```

MadLittleMods:
Thanks for adding Complement tests for this @Johennes 🐍

PR #791, opened 2025-08-11
Room version 12 tests

kegsay:


tests/v12_test.go
```diff
@@ -0,0 +1,1395 @@
...
+
+func asEventIDs(pdus []gomatrixserverlib.PDU) []string {
+	eventIDs := make([]string, len(pdus))
+	for i := range pdus {
+		eventIDs[i] = pdus[i].EventID()
+	}
+	return eventIDs
+}
+
+func TestMSC4311FullCreateEventOnStrippedState(t *testing.T) {
```
MadLittleMods, line 1360: I think this test is mixing up what [MSC4311](https://github.com/matrix-org/matrix-spec-proposals/pull/4311) proposes. Perhaps these were changes to the MSC that came after?

For the client API's like `/sync`, it only proposes that `m.room.create` is a required *stripped* state event.

For the federation API's, alongside requiring `m.room.create`, it also mandates using the full event PDU format for all events in the `invite_room_state`/`knock_room_state` on `m.room.member` events (in `unsigned`)

MadLittleMods: It looks like this is being addressed in https://github.com/matrix-org/complement/pull/796

MadLittleMods: Removing the flawed implementation in Synapse via https://github.com/element-hq/synapse/pull/19723

PR #800, opened 2025-09-01
Lock /createRoom calls

kegsay:
To reduce the chance of multiple rooms being made in the same millisecond which makes v12 rooms unhappy as it causes the same room ID to be generated. This causes flakey tests on v12.


anoadragon453 (approved):
LGTM, thanks!

PR #807, opened 2025-10-02
Verify that ephemeral events down `/sync` don't have a `room_id` field

anoadragon453:
A regression test for https://github.com/element-hq/synapse/pull/19002, which modifies the code in Synapse that strips the `room_id` field from ephemeral events before sending them down to clients.

Ephemeral events such as [`m.typing`](https://spec.matrix.org/v1.16/client-server-api/#mtyping) don't have a `room_id` field when sent to clients. But they do have the field [when sent over federation](https://spec.matrix.org/v1.16/server-server-api/#typing-notifications). These tests ensure this is the case.

### DIFF
--- a/tests/csapi/apidoc_room_receipts_test.go
+++ b/tests/csapi/apidoc_room_receipts_test.go
@@ -17,15 +17,19 @@ func createRoomForReadReceipts(t *testing.T, c *client.CSAPI) (string, string) {
 
 	c.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(c.UserID, roomID))
 
-	eventID := c.SendEventSynced(t, roomID, b.Event{
+	eventID := sendMessageIntoRoom(t, c, roomID)
+
+	return roomID, eventID
+}
+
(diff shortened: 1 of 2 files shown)


tests/csapi/apidoc_room_receipts_test.go
```diff
@@ -45,7 +49,35 @@ func TestRoomReceipts(t *testing.T) {
...
+			t,
+			client.SyncReq{Since: sinceToken},
+			client.SyncEphemeralHas(roomID, func(r gjson.Result) bool {
+				if r.Get("type").Str != "m.read" {
+					return false
+				}
+
+				// Ensure that the `room_id` field does NOT exist.
+				if r.Get("room_id").Exists() {
+					t.Fatalf("Read receipt included `room_id` field down sync: %s", r.Raw)
```
MadLittleMods, line 73:
```suggestion
					t.Fatalf("Read receipt should not contain 'room_id' field when syncing but saw: %s", r.Raw)
```

I think this better explains that this is unexpected/erroneous, etc.

(also applies below)

MadLittleMods (approved):

PR #811, opened 2025-10-20
Convert Sytest `Name/topic keys are correct` to Complement

anoadragon453:
This PR converts [the "Name/topic keys are correct" sytest](https://github.com/matrix-org/sytest/blob/anoa/missing_algorithms_fields/tests/30rooms/70publicroomslist.pl#L1-L111) to Complement.

This was spawned from said test failing in https://github.com/element-hq/synapse/pull/19071, and not wanting to debug perl.

The test was first converted to go via Claude Code, then verified manually. Finally, the test failed when modifying what Synapse returns for `/publicRooms`, which is what we want to see.

### DIFF
--- a/tests/csapi/public_rooms_test.go
+++ b/tests/csapi/public_rooms_test.go
@@ -18,7 +18,7 @@ import (
 func TestPublicRooms(t *testing.T) {
 	deployment := complement.Deploy(t, 1)
 	defer deployment.Destroy(t)
-	server_name := deployment.GetFullyQualifiedHomeserverName(t, "hs1")
+	hs1ServerName := deployment.GetFullyQualifiedHomeserverName(t, "hs1")
 
 	// sytest: Can search public room list
 	t.Run("Can search public room list", func(t *testing.T) {
@@ -29,6 +29,16 @@ func TestPublicRooms(t *testing.T) {
 			"name":       "Test Name",
 			"topic":      "Test Topic Wombles",
 		})
+
+		// Remove the room from the public rooms list to avoid polluting other tests.
+		defer authedClient.MustDo(
+			t,
+			"PUT",
+			[]string{"_matrix", "client", "v3", "directory", "list", "room", roomID},
+			client.WithJSONBody(t, map[string]interface{}{
+				"visibility": "private",
+			}),
+		)
 
 		authedClient.MustDo(
 			t,
@@ -56,16 +66,6 @@ func TestPublicRooms(t *testing.T) {
 				return true
 			}),
 		)
-
-		// Remove the room from the public rooms list to avoid polluting other tests.
-		authedClient.MustDo(
-			t,
-			"PUT",
-			[]string{"_matrix", "client", "v3", "directory", "list", "room", roomID},
-			client.WithJSONBody(t, map[string]interface{}{
-				"visibility": "private",
-			}),
-		)
 	})
 
 	// sytest: Name/topic keys are correct
@@ -89,7 +89,7 @@ func TestPublicRooms(t *testing.T) {
 
 		for _, roomConfig := range roomConfigs {
 			t.Run(fmt.Sprintf("Creating room with alias %s", roomConfig.alias), func(t *testing.T) {
-				expectedCanonicalAlias := fmt.Sprintf("#%s:%s", roomConfig.alias, server_name)
+				expectedCanonicalAlias := fmt.Sprintf("#%s:%s", roomConfig.alias, hs1ServerName)
 
 				// Create the room
 				roomOptions := map[string]interface{}{
@@ -107,6 +107,16 @@ func TestPublicRooms(t *testing.T) {
 
 				roomID := authedClient.MustCreateRoom(t, roomOptions)
 				t.Logf("Created room %s with alias %s", roomID, roomConfig.alias)
+
+				// Remove the room from the public rooms list to avoid polluting other tests.
+				defer authedClient.MustDo(
+					t,
+					"PUT",
+					[]string{"_matrix", "client", "v3", "directory", "list", "room", roomID},
+					client.WithJSONBody(t, map[string]interface{}{
+						"visibility": "private",
+					}),
+				)
 
 				// Poll /publicRooms until the room appears with the correct data
 
@@ -210,16 +220,6 @@ func TestPublicRooms(t *testing.T) {
 
 					t.Fail()
 				}
-
-				// Remove the room from the public rooms list to avoid polluting other tests.
-				authedClient.MustDo(
-					t,
-					"PUT",
-					[]string{"_matrix", "client", "v3", "directory", "list", "room", roomID},
-					client.WithJSONBody(t, map[string]interface{}{
-						"visibility": "private",
-					}),
-				)
 			})
 		}
 	})


tests/csapi/public_rooms_test.go
```diff
@@ -0,0 +1,188 @@
...
+					roomID: roomID,
+					name:   config.name,
+					topic:  config.topic,
+				}
+				t.Logf("Created room %s with alias %s", roomID, config.alias)
+			}
+
+			// Poll /publicRooms until all our rooms appear with correct data
+			authedClient.MustDo(t, "GET", []string{"_matrix", "client", "v3", "publicRooms"},
+				client.WithRetryUntil(15*time.Second, func(res *http.Response) bool {
```
MadLittleMods: Is 15 seconds necessary in real-world testing?

We could re-use `client.SyncUntilTimeout` which is usually 5 seconds for `register`/`login`.

For reference, I had a [10 second timeout](https://github.com/matrix-org/matrix-viewer/blob/5c448f24c252ebc4068824f1e1e47cefe7527b5f/test/test-utils/client-utils.js#L426-L431) in the Matrix Viewer tests but that might have been arbitrary as well.

tests/csapi/public_rooms_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+						// Skip rooms that aren't ours
+						if canonicalAlias == "" {
+							continue
+						}
+
+						// Find which of our rooms this matches
+						var matchedAlias string
+						for alias := range createdRooms {
+							expectedAlias := "#" + alias + ":hs1"
```
MadLittleMods: Hard-coding `hs1` won't necessarily work for all custom Complement `Deployer`/`Deployment`.

We should use `deployment.GetFullyQualifiedHomeserverName(t, "hs1")`

tests/csapi/public_rooms_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+					// Check if we found all our rooms
+					if len(foundRooms) != len(createdRooms) {
+						missing := []string{}
+						for alias := range createdRooms {
+							if !foundRooms[alias] {
+								missing = append(missing, alias)
+							}
+						}
+						t.Logf("Missing rooms in public list: %v (found %d/%d)", missing, len(foundRooms), len(createdRooms))
```
MadLittleMods: This feels pretty weird and misleading as a room only gets added to `foundRooms` if it was found **and** it has all of the expected fields.

I would refactor away from this pattern.

tests/csapi/public_rooms_test.go
```diff
@@ -0,0 +1,188 @@
...
+					// Check each room in the public rooms list
+					for _, roomData := range chunk.Array() {
+						// Verify required keys are present
+						must.MatchGJSON(
+							t,
+							roomData,
+							match.JSONKeyPresent("world_readable"),
+							match.JSONKeyPresent("guest_can_join"),
+							match.JSONKeyPresent("num_joined_members"),
+						)
```
MadLittleMods: This raises and exits the process when it finds something wrong. Whereas the checks below are more lenient and we check all rooms before presenting a result.

We should align one way or the other.

I think it would be nice if we used `err := should.MatchGJSON(jsonResult, matchers...)` and gathered up the errors to present for each expected room at the end.

tests/csapi/public_rooms_test.go
anoadragon453: Sure. Looking at the latest run, the whole test took 850ms: https://github.com/matrix-org/complement/actions/runs/18659975767/job/53197931572#step:8:3607

So I think 5 or 10 seconds is appropriate. I'll re-use `SyncUnitTimeout` to it can be tweaked along with other timeouts..

tests/csapi/public_rooms_test.go
```diff
@@ -12,11 +13,13 @@ import (
 	"github.com/matrix-org/complement/helpers"
 	"github.com/matrix-org/complement/match"
 	"github.com/matrix-org/complement/must"
+	"github.com/matrix-org/complement/should"
 )
 
 func TestPublicRooms(t *testing.T) {
 	deployment := complement.Deploy(t, 1)
 	defer deployment.Destroy(t)
+	hostname := deployment.GetFullyQualifiedHomeserverName(t, "hs1")
```
MadLittleMods: This isn't just a `hostname`. It's the "server name" of the homeserver.

tests/csapi/public_rooms_test.go
```diff
@@ -40,22 +43,8 @@ func TestPublicRooms(t *testing.T) {
 					},
 				}),
 				client.WithRetryUntil(15*time.Second, func(res *http.Response) bool {
```
MadLittleMods:
```suggestion
				client.WithRetryUntil(authedClient.SyncUntilTimeout, func(res *http.Response) bool {
```

tests/csapi/public_rooms_test.go
```diff
@@ -71,5 +60,181 @@ func TestPublicRooms(t *testing.T) {
...
+							validationErrors = append(validationErrors, err)
+						}
+					}
+
+					if len(validationErrors) > 0 {
+						for _, e := range validationErrors {
+							t.Errorf("Validation error for room %s: %s", roomID, e.Error())
+						}
+
+						t.Fail()
```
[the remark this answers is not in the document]
MadLittleMods: Ahh, I guess that's just my lack of Go showing. :+1: 

> `func (c *T) Fail()`
> Fail marks the function as having failed but continues execution.
>
> *-- https://pkg.go.dev/testing#T.Fail*

tests/csapi/public_rooms_test.go
```diff
@@ -12,64 +12,230 @@ import (
...
+
+		// Remove the room from the public rooms list to avoid polluting other tests.
+		authedClient.MustDo(
+			t,
+			"PUT",
+			[]string{"_matrix", "client", "v3", "directory", "list", "room", roomID},
+			client.WithJSONBody(t, map[string]interface{}{
+				"visibility": "private",
+			}),
+		)
```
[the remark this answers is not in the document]
anoadragon453: Yeah, it's a no-op if the room was already private. So right after the room is created seems best.

tests/csapi/public_rooms_test.go
```diff
@@ -12,64 +12,230 @@ import (
...
 					}
+				}
 
-					foundRoomID := results[0].Get("room_id").Str
-					if foundRoomID != roomID {
-						t.Logf("expected room_id %s in search results, got %s", roomID, foundRoomID)
-						return false
+				if len(validationErrors) > 0 {
+					for _, e := range validationErrors {
+						t.Errorf("Validation error for room %s: %s", roomID, e.Error())
```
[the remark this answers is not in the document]
anoadragon453: Oh, I wasn't aware of this method - that's cool to learn about. There's an example [in the documentation](https://pkg.go.dev/errors@go1.25.4#Join) of how to use it.

I think how it'd work is something like the following:

```go
var validationErrors error

err := doSomething()
if err !=nil {
	errors.Join(validationErrors, err)
}

...a few more of those...

if validationErrors != nil {
	t.Fatalf("Validation errors for room %s:\n %s", roomID, validationErrors)
}
```

which would avoid the loop. Though I don't think it's as readable than the simple array of errors. There's also probably a loop involved in displaying `validationErrors` anyways?

I think I'll leave it how it is for now, but TIL about said method, and combining errors in go in general.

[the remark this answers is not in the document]
MadLittleMods: Another shout: The new meta for creating an error from another is to use `%w` ("error wrapping), ex. `fmt.Errorf("failed to generate signing key: %w", err)`

See https://go.dev/doc/go1.13#error_wrapping

Calling it out because I also learned about it recently when going over a codebase with [`golangci-lint`](https://golangci-lint.run/docs/linters/)

PR #808, opened 2025-10-08
Use SyncLeftFrom to check charlie leave event synced to hs1

chrislearn:
In the test related to TestRestrictedRoomsRemoteJoinLocalUser, I am not sure if my understanding is correct. Is it necessary to wait for a while to ensure that the leave event on hs2 are synchronized to hs1?

Signed-off-by: Chrislearn Young <chris@acroidea.com>

### Pull Request Checklist

- [x] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)

PR #272, opened 2021-12-09
Re-design client.SyncUntil

kegsay:
See https://github.com/matrix-org/complement/issues/271 for the rationale why this is being changed. All affected tests have been updated.

Using functional options similar to `client.MustDoFunc`, API usage now looks like:

Initial sync:
```go
bob.InviteRoom(t, roomID, alice.UserID)
alice.JoinRoom(t, roomID, nil)
alice.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(alice.UserID, roomID))
```

Incremental sync:
```go
since := alice.MustSyncUntil(t, client.SyncReq{TimeoutMillis: "0"}) // get a since token
bob.InviteRoom(t, roomID, alice.UserID)
since = alice.MustSyncUntil(t, client.SyncReq{Since: since}, client.SyncInvitedTo(alice.UserID, roomID))
alice.JoinRoom(t, roomID, nil)
alice.MustSyncUntil(t, client.SyncReq{Since: since}, client.SyncJoinedTo(alice.UserID, roomID))
```

The full list of check functions currently supported mirrors the removed functions:
 - `SyncTimelineHas(roomID, func(gjson.Result) bool)`
 - `SyncInvitedTo(userID, roomID)`
 - `SyncJoinedTo(userID, roomID)`
 - `SyncGlobalAccountDataHas(func(gjson.Result) bool)`

The addition of a `client.SyncReq` struct also now allows test writers to either do initial or incremental syncs. Previously, all the helper Sync functions would be doing initial syncs. Furthermore, helper functions for "X must be joined|invited to room Y" can now be done via a passive watching user in the room, whereas previously the user being invited|joined had to be the one syncing.

In addition, when `MustSyncUntil` fails, it produces more structured output like this:
```
@alice:hs1 MustSyncUntil: timed out after 5.042472596s. Seen 7 /sync responses. Checkers:
[...]

PR #813, opened 2025-10-27
Fix flawed membership checks and align membership checking patterns

MadLittleMods:
*Spawning from https://github.com/matrix-org/complement/pull/808 per my suggestion on https://github.com/matrix-org/complement/pull/808#discussion_r2445478145,*

We have other spots that have flawed membership checks that need to be fixed. For example, when our goal is to wait for the user's membership to be `leave`, we should keep checking until it is `leave`. Currently, there are some spots that wait until *any* membership exists for the user and then asserts `leave` on that which is flawed because that user may have previous membership events that may be picked up first instead of waiting for the `leave`.

This PR fixes those flawed checks and aligns our membership checks so we don't cargo cult this bad pattern elsewhere.

 - Generalize our `client.SyncXXX` helpers to use `syncMembershipIn` utility
    - More robust
    - Standardize extra `checks` on event (previously, only available with `client.SyncJoinedTo`)
 - Introduce `client.SyncBannedFrom` so we can differentiate ban/leave

### Dev notes

 - https://github.com/matrix-org/complement/pull/272
 - https://github.com/matrix-org/complement/pull/511

### Pull Request Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ update-ca-certificates
 
 ## Sytest parity
 
-As of 10 February 2023:
+As of 29 October 2025:
 ```
 $ go build ./cmd/sytest-coverage
 $ ./sytest-coverage -v
@@ -507,7 +507,13 @@ $ ./sytest-coverage -v
     ✓ Can get rooms/{roomId}/members
 
 30rooms/60version_upgrade 0/19 tests
-30rooms/70publicroomslist 0/5 tests
+30rooms/70publicroomslist 2/5 tests
+    × Asking for a remote rooms list, but supplying the local server's name, returns the local rooms list
+    × Can get remote public room list
+    × Can paginate public room list
+    ✓ Can search public room list
+    ✓ Name/topic keys are correct
+
 31sync/01filter 2/2 tests
     ✓ Can create filter
     ✓ Can download filter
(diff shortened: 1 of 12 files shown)


client/sync.go
```diff
@@ -288,40 +296,64 @@ func syncMembershipIn(userID, roomID, membership string, checks ...func(gjson.Re
...
+		// Check the timeline
+		//
+		// This is also important to differentiate between leave/ban because those both
+		// appear in the `leave` `roomTypeKey` and we need to specifically check the
+		// timeline for the membership event to differentiate them.
 		var secondErr error
-		if membership == "join" || membership == "leave" || membership == "ban" {
+		// We assume the passively observing client user is joined to the room
+		if clientUserID != userID ||
+			// Otherwise, if the client is the user whose membership we are checking,
```
anoadragon453: This comment and `if` statement don't align.

MadLittleMods: In any case, I've updated the logic layout and comments to make this more clear:

https://github.com/matrix-org/complement/blob/e1ac188f86cc12575e63e994b4caacd64dceffa2/client/sync.go#L347-L357

MadLittleMods:
Thanks for the review @anoadragon453 🦘

PR #815, opened 2025-11-05
Make sure we close HTTP response body

MadLittleMods:
Make sure we close HTTP response body as that's what the docs say:

> If the returned error is nil, the [Response](https://pkg.go.dev/net/http#Response) will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the [Client](https://pkg.go.dev/net/http#Client)'s underlying [RoundTripper](https://pkg.go.dev/net/http#RoundTripper) (typically [Transport](https://pkg.go.dev/net/http#Transport)) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.
>
> *-- https://pkg.go.dev/net/http#Client.Do*

But it also may leak resources :shrug: 

Spawning from exploring this elsewhere in https://github.com/element-hq/synapse-small-hosts/pull/303

### Dev notes

Error handling with `res.Body.Close()`:

 - https://blevesearch.com/news/Deferred-Cleanup,-Checking-Errors,-and-Potential-Problems/
 - https://stackoverflow.com/questions/47293975/should-i-error-check-close-on-a-response-body
 - https://manishrjain.com/must-close-golang-http-response

There is a [`bodyclose`](https://github.com/timakin/bodyclose) lint which is also [available for `golangci-lint`](https://golangci-lint.run/docs/linters/configuration/#bodyclose)

There are many false-positives though, for example:

 - https://github.com/timakin/bodyclose/issues/39

---

HTTP client methods that deal with response bodies:

```
DoFederationRequest
DoHTTPRequest

Do
Get
Put
Post
Delete
Head
```

### Pull Request Checklist

### DIFF
--- a/internal/io.go
+++ b/internal/io.go
@@ -22,14 +22,14 @@ import (
 //
 // ```go
 // res, err := client.Do(req)
-//	defer func(c io.Closer) {
-//		if c != nil {
-//			err := c.Close()
-//			if err != nil {
-//				log.Fatalf("error closing request body stream %v", err)
-//			}
-//		}
-//	}(res.Body)
+// defer func(c io.Closer) {
+// 	if c != nil {
+// 		err := c.Close()
+// 		if err != nil {
+// 			log.Fatalf("error closing request body stream %v", err)
+// 		}
+// 	}
+// }(res.Body)
 // ```
 func CloseIO(c io.Closer, contextString string) {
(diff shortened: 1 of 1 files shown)


MadLittleMods:
Thanks for the review @anoadragon453 🦞

PR #841, opened 2026-02-02
MSC4242: State DAGs

kegsay:
MSC4242: State DAGs

### Pull Request Checklist

### DIFF
--- a/tests/msc4242/main_test.go
+++ b/tests/msc4242/main_test.go
@@ -1,0 +1,11 @@
+package tests
+
+import (
+	"testing"
+
+	"github.com/matrix-org/complement"
+)
+
+func TestMain(m *testing.M) {
+	complement.TestMain(m, "msc4242")
+}
--- a/tests/msc4242/msc4242_api_test.go
+++ b/tests/msc4242/msc4242_api_test.go
@@ -291,9 +291,11 @@ func TestGraph(t *testing.T) {
 }
 
 type Graph struct {
-	stateGraph     map[string][]string
-	eventGraph     map[string][]string
-	events         map[string]gomatrixserverlib.PDU
+	stateGraph map[string][]string
+	eventGraph map[string][]string
+	events     map[string]gomatrixserverlib.PDU
+	// If true, walks the graph induced by prev_events instead of prev_state_events
+	// Used for testing normal /get_missing_events behaviour
 	WalkPrevEvents bool
 }
 
--- a/tests/msc4242/msc4242_api_test.go
+++ b/tests/msc4242/msc4242_api_test.go
@@ -1,0 +1,552 @@
+package tests
+
+// The tests in this directory test the actual API changes for MSC4242 by using
+// a Complement test federation server.
+//
+// These tests reference codes to exhaustively cover the MSC.
+// Some tests are subsequently broken down into 'Inbound' and 'Outbound'
+// which represents the direction of the request being tested
+// FROM THE PERSPECTIVE OF THE HOMESERVER. They are represented with the code 'IO'.
+//
+// E.g for /send_join: if the state DAG is valid and connected, accept.
+//   SJ01-Out: HS does /send_join. Complement sends back a /send_join response with a valid state DAG.
+//             The homeserver does the validation check.
+//   SJ00-In:  Complement does /send_join. HS sends back a valid state DAG.
+//             Complement does the validation check.
+//
+// Not all tests have counterparts because we cannot force the HS to send back
+// faulty state which will fail validation, so typically only the happy case tests will
+// have counterparts. This means tests without counterparts are typically 'Outbound'
+// as we can force the Complement response to be faulty, with the notable exception of
+// pushed events via /send which injects faulty events.
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"slices"
+	"sort"
+	"strconv"
+	"testing"
+
+	"github.com/gorilla/mux"
+	"github.com/matrix-org/complement"
+	"github.com/matrix-org/complement/client"
+	"github.com/matrix-org/complement/ct"
+	"github.com/matrix-org/complement/federation"
+	"github.com/matrix-org/complement/match"
+	"github.com/matrix-org/complement/must"
+	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/gomatrixserverlib/fclient"
+	"github.com/matrix-org/gomatrixserverlib/spec"
+	"github.com/tidwall/gjson"
+)
+
+var (
+	roomVersion gomatrixserverlib.RoomVersion = "org.matrix.msc4242.12"
+	empty                                     = ""
+)
+
+type stateDAGimpl struct {
+	callback               func(resp fclient.RespSendJoin)
+	mutateSendJoinResponse func(resp *fclient.RespSendJoin, joinEvent gomatrixserverlib.PDU)
+	dontCheckFwdExtrems    bool
+}
+type StateDAGRoomOpt func(impl *stateDAGimpl)
+
+// WithCallback is called when Complement joins a state DAG room. It is called with the homeserver's
+// /send_join response.
+func WithCallback(callback func(resp fclient.RespSendJoin)) StateDAGRoomOpt {
+	return func(impl *stateDAGimpl) {
+		impl.callback = callback
+	}
+}
+
+// WithMutateSendJoinResponse is called just before Complement sends a /send_join response back to the homeserver.
+// It allows tests to mutate the response e.g adding or removing state_dag events.
+func WithMutateSendJoinResponse(mutate func(resp *fclient.RespSendJoin, joinEvent gomatrixserverlib.PDU)) StateDAGRoomOpt {
+	return func(impl *stateDAGimpl) {
+		impl.mutateSendJoinResponse = mutate
+	}
+}
+
+// WithDontCheckForwardExtremities if set will not do validation on ForwardExtremities when setting prev_events/prev_state_events.
+func WithDontCheckForwardExtremities() StateDAGRoomOpt {
+	return func(impl *stateDAGimpl) {
+		impl.dontCheckFwdExtrems = true
+	}
+}
+
+// ServerRoomImplStateDAG makes the room use state DAGs i.e set prev_state_events and generate valid /send_join responses.
+func ServerRoomImplStateDAG(t ct.TestLike, srv *federation.Server, opts ...StateDAGRoomOpt) federation.ServerRoomImpl {
+	findLastStateEventID := func(room *federation.ServerRoom) string {
+		room.TimelineMutex.RLock()
+		defer room.TimelineMutex.RUnlock()
+		for i := len(room.Timeline) - 1; i >= 0; i-- {
+			if room.Timeline[i].StateKey() != nil {
+				return room.Timeline[i].EventID()
+			}
+		}
+		t.Logf("%s: failed to find any state event in %d timeline events, no prev_state_events will be set!", room.RoomID, len(room.Timeline))
+		return ""
+	}
+	impl := &stateDAGimpl{}
+	for _, opt := range opts {
+		opt(impl)
+	}
+	return &federation.ServerRoomImplCustom{
+		ServerRoomImplDefault: federation.ServerRoomImplDefault{},
+		ProtoEventCreatorFn: func(def federation.ServerRoomImpl, room *federation.ServerRoom, ev federation.Event) (*gomatrixserverlib.ProtoEvent, error) {
+			proto, err := def.ProtoEventCreator(room, ev)
+			if err != nil {
+				return nil, err
+			}
+			proto.AuthEvents = nil
+
+			if ev.Type == spec.MRoomCreate && ev.StateKey != nil && *ev.StateKey == "" {
+				proto.PrevStateEvents = &[]string{}
+			} else {
+				// if the fwd extrems are state, use that.
+				var fwdExtrems []gomatrixserverlib.PDU
+				for _, id := range room.ForwardExtremities {
+					pdu, ok := room.GetEventInTimeline(id)
+					if ok && pdu.StateKey() != nil {
+						fwdExtrems = append(fwdExtrems, pdu)
+					}
+				}
+				if len(fwdExtrems) == 0 {
+					proto.PrevStateEvents = &[]string{
+						findLastStateEventID(room),
+					}
+				} else {
+					ids := make([]string, len(fwdExtrems))
+					for i := range ids {
+						ids[i] = fwdExtrems[i].EventID()
+					}
+					proto.PrevStateEvents = &ids
+				}
+			}
+			if impl.dontCheckFwdExtrems {
+				cpy := make([]string, len(room.ForwardExtremities))
+				copy(cpy, room.ForwardExtremities)
+				proto.PrevStateEvents = &cpy
+				proto.PrevEvents = cpy
+			}
+			return proto, nil
+		},
+		PopulateFromSendJoinResponseFn: func(def federation.ServerRoomImpl, room *federation.ServerRoom, joinEvent gomatrixserverlib.PDU, resp fclient.RespSendJoin) {
+			stateDAGEvents := resp.StateDAG.UntrustedEvents(roomVersion)
+			sort.Slice(stateDAGEvents, func(i, j int) bool {
+				return stateDAGEvents[i].Depth() < stateDAGEvents[j].Depth()
+			})
+			// we assume no forks and no rejected events, so we can just bluntly replace events in depth order
+			// to work out the current state
+			for _, state := range stateDAGEvents {
+				room.ReplaceCurrentState(state)
+			}
+			room.AddEvent(joinEvent)
+			if impl.callback != nil {
+				impl.callback(resp)
+			}
+		},
+		GenerateSendJoinResponseFn: func(def federation.ServerRoomImpl, room *federation.ServerRoom, s *federation.Server, joinEvent gomatrixserverlib.PDU, expectPartialState, omitServersInRoom bool) fclient.RespSendJoin {
+			res := fclient.RespSendJoin{
+				ServersInRoom: []string{},
+			}
+			res.Event = joinEvent.JSON()
+			res.MembersOmitted = omitServersInRoom
+			for _, ev := range room.Timeline {
+				if ev.StateKey() != nil {
+					res.StateDAG = append(res.StateDAG, ev.JSON())
+				}
+			}
+			serversInRoom := room.ServersInRoom()
+			for _, srvName := range serversInRoom {
+				res.ServersInRoom = append(res.ServersInRoom, string(srvName))
+			}
+			if impl.mutateSendJoinResponse != nil {
+				impl.mutateSendJoinResponse(&res, joinEvent)
+			}
+			room.AddEvent(joinEvent)
+			return res
+		},
+	}
+}
+
+type MSC4242Event struct {
+	federation.Event
+	PrevStateEvents []string
+}
+
+func AsPDUs(t ct.TestLike, events []json.RawMessage) []gomatrixserverlib.PDU {
+	t.Helper()
+	roomVer := gomatrixserverlib.MustGetRoomVersion(roomVersion)
+	pdus := make([]gomatrixserverlib.PDU, len(events))
+	for i := range events {
+		event, err := roomVer.NewEventFromTrustedJSON(events[i], false)
+		must.NotError(t, fmt.Sprintf("failed to load event %d", i), err)
+		pdus[i] = event
+	}
+	return pdus
+}
+
+func AsEventIDs(t ct.TestLike, events []gomatrixserverlib.PDU) []string {
+	t.Helper()
+	eventIDs := make([]string, len(events))
+	for i := range events {
+		eventIDs[i] = events[i].EventID()
+	}
+	return eventIDs
+}
+
+func AsEventJSONs(pdus []gomatrixserverlib.PDU) []json.RawMessage {
+	jsons := make([]json.RawMessage, len(pdus))
+	for i := range pdus {
+		jsons[i] = pdus[i].JSON()
+	}
+	return jsons
+}
+
+// Checks that the StateGraph works
+func TestGraph(t *testing.T) {
+	deployment := complement.Deploy(t, 1)
+	defer deployment.Destroy(t)
+
+	srv := federation.NewServer(t, deployment,
+		federation.HandleKeyRequests(),
+		// accept incoming presence transactions, etc
+		federation.HandleTransactionRequests(nil, nil),
+		// accept incoming /event requests
+		federation.HandleEventRequests(),
+		federation.HandleMakeSendJoinRequests(),
+	)
+	cancel := srv.Listen()
+	defer cancel()
+	bob := srv.UserID("bob")
+	room := srv.MustMakeRoom(t, roomVersion,
+		federation.InitialRoomEvents(roomVersion, bob),
+		federation.WithImpl(ServerRoomImplStateDAG(t, srv)),
+	)
+	// 100 linear events
+	for i := 0; i < 100; i++ {
+		ev := srv.MustCreateEvent(t, room, federation.Event{
+			Type:     spec.MRoomMember,
+			Sender:   bob,
+			StateKey: &bob,
+			Content: map[string]interface{}{
+				"membership":  spec.Join,
+				"displayname": fmt.Sprintf("event %d", i),
+			},
+		})
+		room.AddEvent(ev)
+	}
+	sg := NewStateGraph()
+	sg.Update(room.Timeline)
+	want := AsEventIDs(t, room.Timeline)
+	slices.Reverse(want)
+	lastEvent := room.Timeline[len(room.Timeline)-1]
+	got := AsEventIDs(t, sg.GetMissingEvents([]string{lastEvent.EventID()}, 1000))
+	if !slices.Equal(got, want[1:]) { // we don't return lastEvent
+		ct.Fatalf(t, "failed to walk back up the graph:\ngot  %v\nwant %v", got, want[1:])
+	}
+
+	// now try forks
+	var forkEvents []gomatrixserverlib.PDU
+	for i := 0; i < 10; i++ {
+		ev := srv.MustCreateEvent(t, room, federation.Event{
+			Type:     spec.MRoomMember,
+			Sender:   bob,
+			StateKey: &bob,
+			Content: map[string]interface{}{
+				"membership":  spec.Join,
+				"displayname": fmt.Sprintf("fork event %d", i),
+			},
+		})
+		forkEvents = append(forkEvents, ev)
+	}
+	for _, ev := range forkEvents {
+		room.AddEvent(ev)
+	}
+	sg.Update(forkEvents)
+	room.ForwardExtremities = AsEventIDs(t, forkEvents)
+	merge := srv.MustCreateEvent(t, room, federation.Event{
+		Type:     spec.MRoomMember,
+		Sender:   bob,
+		StateKey: &bob,
+		Content: map[string]interface{}{
+			"membership":  spec.Join,
+			"displayname": "merge",
+		},
+	})
+	room.AddEvent(merge)
+	sg.Update([]gomatrixserverlib.PDU{merge})
+	// should return in ascii sort order
+	want = AsEventIDs(t, forkEvents)
+	slices.Sort(want)
+	got = AsEventIDs(t, sg.GetMissingEvents([]string{merge.EventID()}, 10))
+	if !slices.Equal(got, want) {
+		ct.Fatalf(t, "failed to walk back up the fork graph:\ngot  %v\nwant %v", got, want)
+	}
+}
+
+type Graph struct {
+	stateGraph     map[string][]string
+	eventGraph     map[string][]string
+	events         map[string]gomatrixserverlib.PDU
+	WalkPrevEvents bool
+}
+
+func NewStateGraph() *Graph {
+	return &Graph{
+		stateGraph: make(map[string][]string),
+		eventGraph: make(map[string][]string),
+		events:     make(map[string]gomatrixserverlib.PDU),
+	}
+}
+
+func (g *Graph) Update(pdus []gomatrixserverlib.PDU) {
+	for _, pdu := range pdus {
+		g.stateGraph[pdu.EventID()] = pdu.PrevStateEventIDs()
+		g.eventGraph[pdu.EventID()] = pdu.PrevEventIDs()
+		g.events[pdu.EventID()] = pdu
+	}
+}
+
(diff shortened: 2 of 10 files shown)


tests/msc4242/msc4242_e2e_test.go
```diff
@@ -0,0 +1,60 @@
...
+	defer deployment.Destroy(t)
+
+	alice := deployment.Register(t, "hs1", helpers.RegistrationOpts{})
+	bob := deployment.Register(t, "hs2", helpers.RegistrationOpts{})
+	roomID := alice.MustCreateRoom(t, map[string]interface{}{
+		"room_version": roomVersion,
+		"preset":       "public_chat",
+	})
+	// ensure we are verifying current state by walking the state dag by creating no-op state dag changes.
+	changeDisplayName(t, alice, "alice", 5)
```
MadLittleMods: Why 5?

MadLittleMods: Why does this matter to this test?

tests/msc4242/msc4242_e2e_test.go
```diff
@@ -0,0 +1,60 @@
...
+		eventIDs = append(eventIDs, cli.SendEventSynced(t, roomID, b.Event{
+			Type: "m.room.message",
+			Content: map[string]any{
+				"msgtype": "m.text",
+				"body":    fmt.Sprintf("sendTextMessages %s %d", prefix, i),
+			},
+		}))
+	}
+	return
+}
```
MadLittleMods: `sendTextMessages` isn't used in this file. Good spot for it?

tests/msc4242/msc4242_api_test.go
MadLittleMods: Does this need the `_test` suffix? It's just a bunch of utils

tests/msc4242/msc4242_faulty_events_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+const faultyStateEventType = "faulty.state.event"
+
+// faultyEventTestCases are all the ways you can create faulty events.
+// When GenerateEvents is called, the room must be a normal room with the initial creation events only,
+// as the generator makes various assumptions e.g that the sender is joined, that there are join rules, etc.
+var faultyEventTestCases = []FaultyEventTestCase{
+	{
+		Name:       "Missing all prev_state_events on an event",
+		CodeSuffix: "A",
```
MadLittleMods: Thoughts on more descriptive test ID's?

tests/msc4242/msc4242_get_missing_events_test.go
```diff
@@ -0,0 +1,1180 @@
...
+			})
+			room.AddEvent(sentinel)
+			eventsToSend = append(eventsToSend, sentinel)
+			t.Logf("%s SENTINEL %v", sentinel.EventID(), string(sentinel.JSON()))
+
+			srv.MustSendTransaction(t, deployment, "hs1", AsEventJSONs(eventsToSend), nil)
+
+			alice.MustSyncUntil(t, client.SyncReq{}, client.SyncTimelineHasEventID(roomID, sentinel.EventID()))
+
+			assertGetMissingEventsRecursively(tc.name, room.RoomID, sentinel, tc.getMissingEventsLimit, tc.walkStateDAG, wantEventIDs)
```
MadLittleMods: Long-line of arguments could be split up.

Since there are so many args, it's also hard to follow (and easy to mess up since some have the same types right next to each other) what they are for. Since we only call this in one spot, it's probably fine.

tests/msc4242/msc4242_get_missing_events_test.go
```diff
@@ -0,0 +1,1180 @@
...
+			wantEventIDs := eventIDsPerRequest[i:j]
+			resp, err := srv.FederationClient(deployment).LookupMissingEvents(
+				context.Background(), spec.ServerName(srv.ServerName()), "hs1", roomID, fclient.MissingEvents{
+					Limit:          batchSize,
+					EarliestEvents: []string{},
+					LatestEvents:   AsEventIDs(t, backwardsExtremities),
+					StateDAG:       stateDAG,
+				}, roomVersion,
+			)
+			must.NotError(t, "failed to send /gme request", err)
```
MadLittleMods: The person running into Complement errors most likely won't have the context to immediately know what `/gme` means. I think the abbreviation is annoying anyway

```suggestion
			must.NotError(t, "failed to send /get_missing_events request", err)
```

tests/msc4242/msc4242_get_missing_events_test.go
```diff
@@ -0,0 +1,1180 @@
...
+			pdus := resp.Events.TrustedEvents(roomVersion, false)
+			got := AsEventIDs(t, pdus)
+			if !stateDAG {
+				// /gme returns in 'chronological' order  for non-state DAG requests, so flip it to get the same ordering
+				// as state dag requests.
+				slices.Reverse(got)
+			}
+			t.Logf("assertGetMissingEvents i=%d j=%d got=%v", i, j, got)
```
MadLittleMods: We should make the log more descriptive. I can't tell what it's showing.

tests/msc4242/msc4242_get_missing_events_test.go
```diff
@@ -0,0 +1,1180 @@
...
+			sg.Update(pdus)
+			backwardsExtremities = sg.BackwardsExtremities(backwardsExtremities)
+		}
+	}
+
+	testCases := []struct {
+		testCode              string
+		name                  string
+		generateEventsToWalk  func(room *federation.ServerRoom) (events []gomatrixserverlib.PDU)
+		wantWalkOrder         func(room *federation.ServerRoom, initialEvents, generatedEvents []gomatrixserverlib.PDU) (eventIDs []string)
```
MadLittleMods: Comment to describe what `wantWalkOrder` is.

Expected order of events we expect to see as from `/get_missing_events` as we paginate (assumed).

tests/msc4242/msc4242_get_missing_events_test.go
```diff
@@ -0,0 +1,1180 @@
...
+		getMissingEventsLimit int
+		walkStateDAG          bool
```
MadLittleMods: Comment to describe why what/for? What does this do? What does this test?

tests/msc4242/msc4242_get_missing_events_test.go
```diff
@@ -0,0 +1,1180 @@
...
+				w.WriteHeader(400)
+				w.Write([]byte(err.Error()))
+				return
+			}
+			var resp fclient.RespMissingEvents
+			if !body.StateDAG {
+				// this should be the unifying event, so return the last event in each fork
+				if len(body.LatestEvents) == 1 && body.LatestEvents[0] == latestEvent.EventID() {
+					resp.Events = gomatrixserverlib.NewEventJSONsFromEvents([]gomatrixserverlib.PDU{unifyingEvent})
+					t.Logf("Complement responding to /get_missing_events with 1 event")
```
MadLittleMods:
```suggestion
					t.Logf("Complement responding to /get_missing_events (normal DAG) with 1 event")
```

etc

tests/msc4242/msc4242_get_missing_events_test.go
```diff
@@ -0,0 +1,1180 @@
...
+	//   |             \                 /
+	//   |              BOB_SET_TOPIC (ok)
+	//   |                  /
+	//   |                 /
+	//   BOB_SET_ROOM_NAME
+	//   (rejected, a prev_state_event is rejected)
+	// CHARLIE SET_ROOM_NAME is rejected because charlie didn't join.
+	// BOB_PROFILE is rejected because it is hanging off a rejected event.
+	// DORIS_SET_ROOM_NAME soft fails because there is a concurrent ban.
+	// Technically it isn't soft-fail, it's just state res but you get the idea.
```
MadLittleMods: Can we just describe this as an intentional state rollback instead?

tests/msc4242/msc4242_get_missing_events_test.go
```diff
@@ -0,0 +1,1180 @@
...
+	//   |           /                    \
+	//   |      BOB_BAN_DORIS            DORIS_SET_ROOM_NAME
+	//   |         (ok)                      (soft-fail)
+	//   |            \                   /
+	//   |             \                 /
+	//   |              BOB_SET_TOPIC (ok)
+	//   |                  /
+	//   |                 /
+	//   BOB_SET_ROOM_NAME
+	//   (rejected, a prev_state_event is rejected)
```
MadLittleMods: Is this ASCII art manually created? If generated, would be nice to have the source so people can edit it as necessary in the future

MadLittleMods: To nest it like the others (better attach it to the event)

```suggestion
	//    (rejected, a prev_state_event is rejected)
```

tests/msc4242/msc4242_get_missing_events_test.go
```diff
@@ -0,0 +1,1180 @@
...
+			testCode: "GME03D",
+			name:     "returns nothing for a valid event which points to multiple events, one of which is a rejected event ID",
+			req: fclient.MissingEvents{
+				Limit:          5,
+				EarliestEvents: []string{},
+				LatestEvents:   []string{bobSetRoomName.EventID()},
+				StateDAG:       true,
+			},
+			wantEventIDs: []string{},
+		},
```
MadLittleMods: This style of test case is much easier to reason about than the others.

I haven't stared into the abyss enough to see the exact solution but is there any simplification that `TestMSC4242GetMissingEventsInbound` could use from this. Setup -> test cases (request and expectation)

tests/msc4242/msc4242_send_join_test.go
```diff
@@ -0,0 +1,752 @@
...
+		t.Logf("Event ID %s", stateEvent.EventID())
+		t.Logf("%s", string(stateEvent.JSON()))
+		knownEventIDs.Add(stateEvent.EventID())
+		stateDAG[stateEvent.EventID()] = prevStateEvents
+	}
+
+	must.Equal(t, wantStateEventTypes.Cardinality(), 0, fmt.Sprintf("did not see all room state: missing %v", wantStateEventTypes.String()))
+
+	// Include the prev state events of the join event itself as that is what we care about
+	joinEventprevStateEvents := prevStateEvents(t, gjson.ParseBytes(sendJoinResp.Event))
```
MadLittleMods:
```suggestion
	joinEventPrevStateEvents := prevStateEvents(t, gjson.ParseBytes(sendJoinResp.Event))
```

tests/msc4242/msc4242_send_join_test.go
```diff
@@ -0,0 +1,752 @@
...
+	for _, pae := range joinEventprevStateEvents {
+		allKnownPrevStateEvents.Add(pae)
+	}
+	stateDAG[joinEvent.EventID()] = joinEventprevStateEvents
+
+	t.Logf("known event IDs: %v", knownEventIDs.String())
+	t.Logf("all known prev state events: %v", allKnownPrevStateEvents.String())
+	// we expect 10 known prev state events:
+	// - the initial set from room creation (create,join,pl,join_rules, his vis)
```
MadLittleMods:
```suggestion
	// - the initial set from room creation (create,join,power_levels,join_rules,history_visibility)
```

tests/msc4242/msc4242_send_join_test.go
```diff
@@ -0,0 +1,752 @@
...
+		federation.HandleTransactionRequests(nil, nil),
+		// accept incoming /event requests
+		federation.HandleEventRequests(),
+		federation.HandleMakeSendJoinRequests(),
+	)
+	cancel := srv.Listen()
+	defer cancel()
+	bob := srv.UserID("bob")
+	testCases := faultyEventTestCases
+	var mu sync.Mutex
```
MadLittleMods: What's the point of this lock? (comment)

tests/msc4242/msc4242_send_join_test.go
```diff
@@ -0,0 +1,752 @@
...
+	roomID := alice.MustCreateRoom(t, map[string]interface{}{
+		"room_version": roomVersion,
+		"preset":       "public_chat",
+	})
+	charlie.MustJoinRoom(t, roomID, []string{"hs1"})
+	changeDisplayName(t, alice, "alice", 4)
+	changeDisplayName(t, alice, "final", 1)
+	textEventIDs := sendTextMessages(t, alice, roomID, "before_join", 5)
+	t.Logf("sent messages: %v", textEventIDs)
+	// we must sync with LL members enabled.
```
MadLittleMods:
```suggestion
	// we must sync with LL members enabled because TODO
```

(expand on why it's necessary)

MadLittleMods (commented):
Partial review

Reviewed up to `TestMSC4242OnSendJoinSJ04`

Haven't looked at the utils in `tests/msc4242/msc4242_api_test.go` yet

tests/msc4242/msc4242_e2e_test.go
kegsay: It's used in send_join_test so moved there.

tests/msc4242/msc4242_faulty_events_test.go
kegsay: The intent is to have a shorthand way to express the test name.

MadLittleMods: Why?

tests/msc4242/msc4242_faulty_events_test.go
```diff
@@ -0,0 +1,223 @@
...
+				Event: federation.Event{
+					Type:   "m.room.message",
+					Sender: sender,
+					Content: map[string]interface{}{
+						"body":    "I am a message",
+						"msgtype": "m.text",
+					},
+				},
+				PrevStateEvents: []string{
+					room.CurrentState(spec.MRoomJoinRules, "").EventID(),
```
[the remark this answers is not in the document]
MadLittleMods: What happens when we have `prev_events` that don't match the correct `prev_state_events`?

I think the [example](https://github.com/matrix-org/matrix-spec-proposals/blob/d14e98ce98477035e4cac45feb765d8b7361ba8b/proposals/4242-state-dags.md?plain=1#L143-L174) from MSC4242 under the *Background: Authorisation rules today* section explains the possible soft failure scenario that could happen.

tests/msc4242/msc4242_get_missing_events_test.go
```diff
@@ -0,0 +1,1195 @@
...
+			if !stateDAG {
+				// /gme returns in 'chronological' order  for non-state DAG requests, so flip it to get the same ordering
+				// as state dag requests.
+				slices.Reverse(got)
+			}
+			t.Logf("assertGetMissingEvents startIndex=%d endIndex=%d gotEventIDs=%v", startIndex, endIndex, got)
+			if !slices.Equal(got, wantEventIDs) {
+				ct.Errorf(t, "failed to see correct event IDs in test '%s'. \nGot  %v \nWant %v", name, got, wantEventIDs)
+				return
+			}
```
MadLittleMods: To get a nicer diff, it would be nice to re-use the `assertEventsInOrder` utility from `tests/csapi/room_messages_test.go`

tests/msc4242/msc4242_send_join_test.go
```diff
@@ -0,0 +1,782 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			roomID := alice.MustCreateRoom(t, map[string]interface{}{
+				"room_version": roomVersion,
+				"preset":       "trusted_private_chat", // Bob has the PL to send state events
+				"invite":       []string{bob},
+			})
+			room, _ := MustJoinRoom(t, srv, deployment, spec.ServerName("hs1"), roomID, bob)
```
MadLittleMods:
```suggestion
			room, _ := MustJoinRoom(t, srv, deployment, deployment.GetFullyQualifiedHomeserverName(t, "hs1"), roomID, bob)
```

tests/msc4242/msc4242_send_join_test.go
```diff
@@ -0,0 +1,782 @@
...
+				eventJSONs[i] = events[i].JSON()
+				eventIDs[i] = events[i].EventID()
+			}
+			srv.MustSendTransaction(t, deployment, "hs1", eventJSONs, nil)
+			// we don't know when the server has processed these events sadly, as rejected events don't appear down /sync.
+			// We must sleep prior to leaving as our leave event will reference events in the above transaction
+			// potentially, and we don't want to cause /get_missing_events requests.
+			time.Sleep(time.Second)
+			// Leave so we can re-join and check the /send_join response
+			srv.MustLeaveRoom(t, deployment, "hs1", roomID, bob)
```
MadLittleMods:
```suggestion
			srv.MustLeaveRoom(t, deployment, deployment.GetFullyQualifiedHomeserverName(t, "hs1"), roomID, bob)
```

tests/msc4242/msc4242_send_test.go
```diff
@@ -0,0 +1,239 @@
...
+			statePDU := mustCreateEvent(t, srv, room, state)
+			sentinel := srv.MustCreateEvent(t, room, federation.Event{
+				Type:   "m.room.message",
+				Sender: charlie,
+				Content: map[string]interface{}{
+					"msgtype": "m.text",
+					"body":    "Sentinel check",
+				},
+			})
+			srv.MustSendTransaction(t, deployment, "hs1", AsEventJSONs([]gomatrixserverlib.PDU{msgPDU, statePDU, sentinel}), nil)
```
MadLittleMods:
```suggestion
			srv.MustSendTransaction(t, deployment, deployment.GetFullyQualifiedHomeserverName(t, "hs1"), AsEventJSONs([]gomatrixserverlib.PDU{msgPDU, statePDU, sentinel}), nil)
```

tests/msc4242/msc4242_state_test.go
```diff
@@ -0,0 +1,788 @@
...
+				if err := json.NewEncoder(w).Encode(&resp); err != nil {
+					must.NotError(t, "failed to encode response", err)
+				}
+			}
+			waiter := helpers.NewWaiter()
+			getMissingEventsStateDAGHandler = func(w http.ResponseWriter, _ *fclient.MissingEvents) {
+				w.WriteHeader(502)
+				waiter.Finish()
+			}
+			srv.MustSendTransaction(t, deployment, "hs1", AsEventJSONs([]gomatrixserverlib.PDU{events["C"]}), nil)
```
MadLittleMods:
```suggestion
			srv.MustSendTransaction(t, deployment, deployment.GetFullyQualifiedHomeserverName(t, "hs1"), AsEventJSONs([]gomatrixserverlib.PDU{events["C"]}), nil)
```

tests/msc4242/msc4242_state_test.go
```diff
@@ -0,0 +1,788 @@
...
+		return client.SyncTimelineHasEventID(room.RoomID, state2.EventID())(clientUserID, topLevelSyncJSON)
+	})
+	for wantType, wantEventID := range wantCurrentState {
+		if currentState[wantType] != wantEventID {
+			ct.Errorf(t, "current state for %v is %v but wanted %v", wantType, currentState[wantType], wantEventID)
+		}
+	}
+
+	// send the second batch
+	srv.MustSendTransaction(t, deployment, "hs1", AsEventJSONs([]gomatrixserverlib.PDU{
```
MadLittleMods:
```suggestion
	srv.MustSendTransaction(t, deployment, deployment.GetFullyQualifiedHomeserverName(t, "hs1"), AsEventJSONs([]gomatrixserverlib.PDU{
```

tests/msc4242/msc4242_get_missing_events_test.go
```diff
@@ -0,0 +1,1180 @@
...
+		i := 0
+		backwardsExtremities := []gomatrixserverlib.PDU{fromEvent}
+		for {
+			j := i + batchSize
+			if j > len(eventIDsPerRequest) {
+				j = len(eventIDsPerRequest)
+			}
+			wantEventIDs := eventIDsPerRequest[i:j]
+			resp, err := srv.FederationClient(deployment).LookupMissingEvents(
+				context.Background(), spec.ServerName(srv.ServerName()), "hs1", roomID, fclient.MissingEvents{
```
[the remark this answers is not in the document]
MadLittleMods: Commented on all of the spots that were missed.

 - Could have a new-type for this.
 - We could make the tests fail by changing the underlying hostnames in the default Complement `Deployment`.

tests/msc4242/msc4242_get_missing_events_test.go
MadLittleMods: This still applies in a couple other spots

tests/msc4242/msc4242_faulty_events_test.go
```diff
@@ -0,0 +1,223 @@
...
+						Type:     faultyStateEventType,
+						Sender:   sender,
+						StateKey: &empty,
+						Content: map[string]interface{}{
+							"name": "This event is missing all prev_state_events so should be rejected",
+						},
+						PrevEvents: []string{room.CurrentState(spec.MRoomMember, sender).EventID()},
+					},
+					// no event in the state DAG has this event ID.
+					PrevStateEvents: []string{"$4PRgaFIMcD9z4vzgkUUm0YI5CZHYORUPzWGJac6guAo"},
```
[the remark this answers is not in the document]
MadLittleMods: > No, I want this to look like a normal event ID, else HSes may reject early because it isn't a valid SHA256 (length).
>
> *-- @kegsay, https://github.com/matrix-org/complement/pull/841#discussion_r2830559039*

We can pad it out however long it needs to be. But seems like servers should treat these opaquely.

```suggestion
					PrevStateEvents: []string{"$eventIdDoesNotExist000000000000000000000000"},
```

tests/msc4242/msc4242_get_missing_events_test.go
MadLittleMods: This was updated to:

```

	// DORIS_SET_ROOM_NAME state rollbacks because there is a concurrent ban.
```

tests/msc4242/msc4242_send_join_test.go
```diff
@@ -0,0 +1,752 @@
...
+	})
+	changeDisplayName(t, alice, "alice", 5)
+	sendTextMessages(t, alice, roomID, "before_join", 5)
+	roomSyncData, _ := alice.MustSync(t, client.SyncReq{
+		Filter: `{"event_format":"federation","room":{"timeline":{"limit":50}}}`,
+	})
+
+	// the join event should have the same prev_state_events as the last message in the room
+	lastEvent := roomSyncData.Get("rooms.join." + client.GjsonEscape(roomID) + ".timeline.events.@reverse.0")
+	must.Equal(t, lastEvent.Exists(), true, "last timeline entry for room does not exist")
```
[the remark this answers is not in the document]
MadLittleMods:
```
	// make sure there is a last timeline event
	must.Equal(t, lastEvent.Exists(), true, "last timeline entry for room does not exist")
```

Why?

These changes also weren't applied to the commented out Faster-remote room joins variant that uses the same code.

tests/msc4242/msc4242_send_join_test.go
```diff
@@ -0,0 +1,782 @@
...
+				charlieBan := mustCreateEvent(t, srv, room, MSC4242Event{
+					Event: federation.Event{
+						Type:     spec.MRoomMember,
+						Sender:   bob,
+						StateKey: &charlie,
+						Content: map[string]interface{}{
+							"membership": "ban",
+						},
+						PrevEvents: []string{
+							charlieMod.EventID(),
```
MadLittleMods: Hard to spot what's concurrent here, we should comment about it

tests/msc4242/msc4242_send_test.go
```diff
@@ -0,0 +1,239 @@
...
+				federation.WithImpl(ServerRoomImplStateDAG(t, srv)),
+			)
+			alice.MustJoinRoom(t, room.RoomID, []spec.ServerName{srv.ServerName()})
+			since := alice.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(alice.UserID, room.RoomID))
+			faultyEvents := tc.GenerateEvents(t, srv, room, bob)
+
+			srv.MustSendTransaction(t, deployment, "hs1", AsEventJSONs(faultyEvents), nil)
+
+			// send another txn with a sentinel to know that the faulty events have been processed, and
+			// ensure we pin it on an event which we know what not rejected (Alice's join)
```
MadLittleMods:
```suggestion
			// ensure we pin it on an event which we know was not rejected (Alice's join)
```

tests/msc4242/msc4242_state_test.go
```diff
@@ -0,0 +1,788 @@
...
+			if isLimited && !seenEvents[events["C"].EventID()] {
+				for _, backfilledEvent := range backfill(t, alice, roomVersion, room.RoomID, prevBatch, 1) {
+					t.Logf("/messages => %s", backfilledEvent.EventID())
+					seenEvents[backfilledEvent.EventID()] = true
+				}
+			}
+			must.Equal(t, seenEvents[events["A"].EventID()], true, "did not see event A")
+			must.Equal(t, seenEvents[events["B"].EventID()], true, "did not see event B")
+			must.Equal(t, seenEvents[events["C"].EventID()], true, "did not see event C")
+			must.Equal(t, seenEvents[events["D"].EventID()], true, "did not see event D")
```
MadLittleMods: It would be more helpful if we presented the full picture of what's missing/seen when this fails.

tests/msc4242/msc4242_state_test.go
```diff
@@ -0,0 +1,788 @@
...
+	//           +---|----STATE1
+	//           |   |         |
+	//           |    `---STATE2   <-- first batch
+	//           |    `        |
+	//           +---|--- STATE3
+	//               |         |
+	//                `---STATE4   <--- 2nd batch
+	//
+	// Servers should be ignoring prev_events when calculating state.
+	// If they are, the state at STATE2 will include STATE1. If they are reading prev_events instead, then it will not include STATE1.
```
MadLittleMods: I understand the concept but not how we're getting this from the graph

tests/msc4242/msc4242_api_test.go
```diff
@@ -0,0 +1,554 @@
...
+func ServerRoomImplStateDAG(t ct.TestLike, srv *federation.Server, opts ...StateDAGRoomOpt) federation.ServerRoomImpl {
+	findLastStateEventID := func(room *federation.ServerRoom) string {
+		room.TimelineMutex.RLock()
+		defer room.TimelineMutex.RUnlock()
+		for i := len(room.Timeline) - 1; i >= 0; i-- {
+			if room.Timeline[i].StateKey() != nil {
+				return room.Timeline[i].EventID()
+			}
+		}
+		t.Logf("%s: failed to find any state event in %d timeline events, no prev_state_events will be set!", room.RoomID, len(room.Timeline))
```
MadLittleMods: Is this ever desired? (or programming error)

tests/msc4242/main_test.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"testing"
+
+	"github.com/matrix-org/complement"
+)
+
+func TestMain(m *testing.M) {
+	complement.TestMain(m, "msc4242")
+}
```
MadLittleMods: Note: While I have read the entire MSC and this PR, I don't think I've onboarded enough to recognize if there are missing test cases. It's my first time with my own hand in the fire when it comes to State DAG's.

This PR is so large, it's probably worth having an LLM also sanity check this against the MSC for completeness.

kegsay: I've shoved this through Claude and it has made some obvious suggestions:

> Missing: G: has >0 prev_state_events and it is an m.room.create event

> STATE03: XXX receiving a state event via /send which does NOT alter the current state (say it's a concurrent topic change which does not win state res) **should | should not** go down /sync and be present in the timeline.

> Faster remote room joins is commented out.

And some less obvious ones:

> No e2e soft-failure convergence test
>
> The spec dedicates a section to the specific failure mode where soft-failing an event prevents a later server from seeing it and converging on correct state. SJ04C covers the fix (including concurrent events in /send_join), but there's no multi-server e2e test (msc4242_e2e_test.go only has the basic TestMSC4242FederationSimple). The spec's scenario: "Bystander soft-fails C; long-dead server rejoins and never learns C's effect on state" is not exercised.

> 5. No test for prev_state_events cross-room rejection
>
> The auth rules section adds:
> > If there are entries which do not belong in the same room, reject.
>
> None of the faultyEventTestCases generates a prev_state_events entry pointing to a state event from a different room. This is distinct from "bogus event ID" (CodeSuffix A/B) as those are unknown IDs, not known-but-wrong-room IDs.

When probed for more missing test cases, it produced:

> earliest_events never exercised with non-empty values
>
> Every call to /get_missing_events across all tests passes EarliestEvents: []string{}. The spec explicitly defines the algorithm as mapping earliest_events to an initial "seen" set to bound the walk. There's no test verifying the server actually stops walking when it hits a known event ID. This matters because the recursive pagination in assertGetMissingEventsRecursively works around this by updating latest_events instead, so a broken earliest_events implementation would never be caught.

> Outbound prev_state_events on non-state events never checked
>
> The spec states all events — including messages — must carry prev_state_events. The tests verify the HS handles incoming events with various prev_state_events configurations, and TestMSC4242SendJoinSJ01Inbound checks prev_state_events on the join event specifically. But there's no test inspecting events sent by hs1 (messages, profile changes, etc.) to confirm they actually include prev_state_events pointing to the correct state DAG forward extremities. The E2E test only checks event IDs arrive, not event structure.

> Duplicate event IDs in prev_state_events not tested
>
> The spec removes the check for duplicate (type, state_key) pairs because prev_state_events can legitimately reference two events with the same tuple on different branches. But there's no faulty event test case for a prev_state_events list containing the same event ID twice. Whether this should be rejected (malformed) or tolerated (treated as deduplicated) isn't tested either way.

> No test that /state, /state_ids, /event_auth are not called during normal operation
>
> The spec explicitly obsoletes these three endpoints. While srv.UnexpectedRequestsAreErrors being true (the default) would catch this in some tests, several tests set it to false, and no test specifically asserts the HS does not fall back to these endpoints when processing events or calculating room state. A targeted test with a Complement server that returns 404 for all three endpoints and verifies the HS still converges correctly would strengthen this.

I'll consider these and resolve this issue when done.

tests/msc4242/msc4242_send_join_test.go
```diff
@@ -0,0 +1,752 @@
...
+			[2]string{spec.MRoomMember, alice.UserID},
+			// NB: Bob is NOT part of this as his join is the 'Event' in the send_join response
+			[2]string{"m.room.power_levels", ""},
+			[2]string{"m.room.join_rules", ""},
+			[2]string{"m.room.history_visibility", ""},
+		},
+	)
+
+	// ensure the state DAG is connected: that means we do not have any prev_state_events
+	// that we haven't been told about.
```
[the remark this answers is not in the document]
MadLittleMods: Other discussions:

 - https://github.com/element-hq/synapse/pull/19424#discussion_r2855962111
 - https://github.com/element-hq/synapse/pull/19425#discussion_r2861897652

PR #845, opened 2026-02-24
MSC4140: drop delayed state interference tests

AndrewFerr:
Don't test that a user's delayed state events are cancelled by a more recent state event from a different user.  That behaviour [has been delegated to an "alternative"](https://github.com/matrix-org/matrix-spec-proposals/pull/4140/changes/3efecea123eef4d672daa67b76b75a91899fed90) in the delayed event MSC4140 for some time now, and may be removed from Synapse ([Synapse PR](https://github.com/element-hq/synapse/pull/19495))

Signed-off-by: Andrew Ferrazzutti <andrewf@element.io>

### Pull Request Checklist


tests/msc4140/delayed_event_test.go
MadLittleMods, line 1: > Don't test that a user's delayed state events are cancelled by a more recent state event from a different user.  That behaviour has been delegated to an "alternative" in the delayed event [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for some time now [...]

Can you link the relevant discussion for reference?

MadLittleMods, line 1: > and may be removed from Synapse.

Is this tracked somewhere? Probably best as an issue and link it from a `FIXME` comment within the code. It would be nice not to forget about when time comes to stabilize this feature.

tests/msc4140/delayed_event_test.go
```diff
@@ -360,85 +360,6 @@ func TestDelayedEvents(t *testing.T) {
 		})
 	})
 
-	t.Run("delayed state is not cancelled by new state from the same user", func(t *testing.T) {
```
MadLittleMods, line 363: > Don't test that a user's delayed state events are cancelled by a more recent state event from a different user. 

This is for the *same* user. Should we actually remove it as well?

tests/msc4140/delayed_event_test.go
AndrewFerr: I linked to the commit that moved the behaviour to the Alternatives section: https://github.com/matrix-org/matrix-spec-proposals/pull/4140/changes/3efecea123eef4d672daa67b76b75a91899fed90

AndrewFerr: It's being done by this PR: https://github.com/element-hq/synapse/pull/19495

tests/msc4140/delayed_event_test.go
AndrewFerr: IMO this may as well be removed, because there's nothing in the spec that suggests new state from the same user should cancel their pending delayed state, and there's no point in testing that unspecified behaviour doesn't happen.

tests/msc4140/delayed_event_test.go
MadLittleMods: Thanks for the link! Added to the PR description :fast_forward:

PR #847, opened 2026-03-04
test: events sent during remote join handshake should not be lost

FrenchGithubUser:
Test for https://github.com/element-hq/synapse/pull/19390

### Disclaimer

As I'm not super fluent with go, I used a lot of AI to write this test. However, I did review it as best as I could together with someone else from Famedly.

### Pull Request Checklist

### DIFF
--- a/tests/federation_room_join_test.go
+++ b/tests/federation_room_join_test.go
@@ -595,8 +595,10 @@ func TestEventBetweenMakeJoinAndSendJoinIsNotLost(t *testing.T) {
 	// After send_join, we wait for hs1 to send us either:
 	//  - the message event itself, or
 	//  - any event whose prev_events reference the message (e.g. a dummy event)
-	// atomic.Value is used because messageEventID is written on the main goroutine
-	// and read on the HTTP handler goroutine, with no other synchronization.
+    //
+	// atomic.Value is used because messageEventID is written on the main goroutine and
+	// read on the HTTP handler goroutine, and needs synchronization (without
+	// synchronization, writes are not guaranteed to be observed by other goroutines)
 	var messageEventID atomic.Value
 	messageDiscoverableWaiter := helpers.NewWaiter()
 
--- a/tests/federation_room_join_test.go
+++ b/tests/federation_room_join_test.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -594,7 +595,9 @@ func TestEventBetweenMakeJoinAndSendJoinIsNotLost(t *testing.T) {
 	// After send_join, we wait for hs1 to send us either:
 	//  - the message event itself, or
 	//  - any event whose prev_events reference the message (e.g. a dummy event)
-	var messageEventID string
+	// atomic.Value is used because messageEventID is written on the main goroutine
+	// and read on the HTTP handler goroutine, with no other synchronization.
+	var messageEventID atomic.Value
 	messageDiscoverableWaiter := helpers.NewWaiter()
 
 	srv := federation.NewServer(t, deployment,
@@ -613,10 +616,7 @@ func TestEventBetweenMakeJoinAndSendJoinIsNotLost(t *testing.T) {
 	// srv.rooms). Instead we parse the raw transaction body ourselves.
 	srv.Mux().Handle("/_matrix/federation/v1/send/{transactionID}", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		body, err := io.ReadAll(req.Body)
-		if err != nil {
-			t.Fatalf("failed to read request body in /send handler: %v", err)
-			return
-		}
+		must.NotError(t, "failed to read request body in /send handler: %v", err)
 		txn := gjson.ParseBytes(body)
 		txn.Get("pdus").ForEach(func(_, pdu gjson.Result) bool {
 			eventID := pdu.Get("event_id").String()
@@ -626,24 +626,27 @@ func TestEventBetweenMakeJoinAndSendJoinIsNotLost(t *testing.T) {
 			// messageEventID is set after make_join but before send_join.
 			// Transactions can arrive before that window, so skip PDUs that
 			// arrive before we know which event to look for.
-			if messageEventID == "" {
+			msgID, _ := messageEventID.Load().(string)
+			if msgID == "" {
 				return true
 			}
 
 			// Check if this IS the message event (server pushed it directly).
-			if eventID == messageEventID {
+			if eventID == msgID {
 				messageDiscoverableWaiter.Finish()
 				return true
 			}
 
-			// Check if this event's prev_events directly reference the message
-			// (e.g. a dummy event tying the two forward extremities together).
-			// If so, the joining server can backfill from that event and will
-			// discover the message. We only check one level of prev_events:
-			// if the reference is deeper in the DAG the joining server can
-			// still reach the message through backfill.
+			// Check if this event's prev_events directly reference the message (e.g. a dummy
+			// event tying the two forward extremities together). If so, the joining server
+			// can backfill from that event and will discover the message.
+			//
+			// XXX: We only check one level of prev_events: if the reference is deeper in the
+			// DAG, it's valid and the joining server can still reach the message through
+			// backfill but our checks don't account for that yet (feel free to edit this
+			// assertion if you run into this)
 			pdu.Get("prev_events").ForEach(func(_, prevEvent gjson.Result) bool {
-				if prevEvent.String() == messageEventID {
+				if prevEvent.String() == msgID {
 					messageDiscoverableWaiter.Finish()
 					return false
 				}
@@ -683,14 +686,14 @@ func TestEventBetweenMakeJoinAndSendJoinIsNotLost(t *testing.T) {
 	// Step 2: Alice sends a message on hs1. This advances the DAG past the
 	// point captured by make_join's prev_events. The Complement server is not
 	// yet in the room, so it won't receive this event via normal federation.
-	messageEventID = alice.SendEventSynced(t, roomID, b.Event{
+	messageEventID.Store(alice.SendEventSynced(t, roomID, b.Event{
 		Type: "m.room.message",
 		Content: map[string]interface{}{
 			"msgtype": "m.text",
 			"body":    "Message sent between make_join and send_join",
 		},
-	})
-	t.Logf("Alice sent message %s between make_join and send_join", messageEventID)
+	}))
+	t.Logf("Alice sent message %s between make_join and send_join", messageEventID.Load())
 
 	// Step 3: Build and sign the join event, then send_join.
 	// The join event's prev_events are from step 1 (before the message),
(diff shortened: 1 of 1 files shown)


tests/federation_room_join_test.go
```diff
@@ -574,6 +575,131 @@ func TestSendJoinPartialStateResponse(t *testing.T) {
...
+	// After send_join, we wait for hs1 to send us either:
+	//  - the message event itself, or
+	//  - any event whose prev_events reference the message (e.g. a dummy event)
+	var messageEventID string
+	messageDiscoverableWaiter := helpers.NewWaiter()
+
+	srv := federation.NewServer(t, deployment,
+		federation.HandleKeyRequests(),
+	)
+	srv.UnexpectedRequestsAreErrors = false
```
MadLittleMods, line 607: Which unexpected requests do we expect?

We should try to avoid this if possible. Otherwise, we should comment what kinds of requests are flying around

tests/federation_room_join_test.go
```diff
@@ -574,6 +575,131 @@ func TestSendJoinPartialStateResponse(t *testing.T) {
...
+
+	// Custom /send handler: the Complement server won't be in the room until
+	// send_join completes, so we can't use HandleTransactionRequests (which
+	// requires the room in srv.rooms). Instead we parse the raw transaction.
+	srv.Mux().Handle("/_matrix/federation/v1/send/{transactionID}", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		body, _ := io.ReadAll(req.Body)
```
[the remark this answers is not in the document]
MadLittleMods: 👍 Handling the error is prudent here

tests/federation_room_join_test.go
```diff
@@ -574,6 +575,131 @@ func TestSendJoinPartialStateResponse(t *testing.T) {
...
+		txn := gjson.ParseBytes(body)
+		txn.Get("pdus").ForEach(func(_, pdu gjson.Result) bool {
+			eventID := pdu.Get("event_id").String()
+			eventType := pdu.Get("type").String()
+			t.Logf("Received PDU via /send: type=%s id=%s", eventType, eventID)
+
+			if messageEventID == "" {
+				return true
+			}
```
MadLittleMods, line 631: Comment why we do this

tests/federation_room_join_test.go
```diff
@@ -574,6 +575,131 @@ func TestSendJoinPartialStateResponse(t *testing.T) {
...
+
+	srv := federation.NewServer(t, deployment,
+		federation.HandleKeyRequests(),
+	)
+	srv.UnexpectedRequestsAreErrors = false
+
+	// Custom /send handler: the Complement server won't be in the room until
+	// send_join completes, so we can't use HandleTransactionRequests (which
+	// requires the room in srv.rooms). Instead we parse the raw transaction.
+	srv.Mux().Handle("/_matrix/federation/v1/send/{transactionID}", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
```
MadLittleMods: Comment what we're trying to accomplish with this handler

tests/federation_room_join_test.go
```diff
@@ -574,6 +575,131 @@ func TestSendJoinPartialStateResponse(t *testing.T) {
...
+
+			// Check if this event's prev_events reference the message
+			// (e.g. a dummy event tying the forward extremities together).
+			pdu.Get("prev_events").ForEach(func(_, prevEvent gjson.Result) bool {
+				if prevEvent.String() == messageEventID {
+					messageDiscoverableWaiter.Finish()
+					return false
+				}
+				return true
+			})
```
MadLittleMods: Comment why this is good enough.

I assume because people can backfill to get the message.

But then this brings up the question, what if it's a prev_event further back in a chain? (we currently don't detect that as good enough) - we need to at-least explain this edge case although perhaps we should `/backfill` up to the beginning of the room and ensure the `messageEventID` was in there.

tests/federation_room_join_test.go
```diff
@@ -574,6 +575,131 @@ func TestSendJoinPartialStateResponse(t *testing.T) {
...
+
+			return true
+		})
+		w.WriteHeader(200)
+		w.Write([]byte(`{"pdus":{}}`))
```
MadLittleMods, line 659: Comment why responding this way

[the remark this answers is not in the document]
MadLittleMods: This seems like nonsense ⏩

tests/federation_room_join_test.go
MadLittleMods: It looks like this is partially addressed below but we should also explain it here:

```
	// Step 4: hs1 should make the missed message discoverable to the joining
	// server. We accept either receiving the message event directly, or
	// receiving any event whose prev_events reference it (allowing the
	// joining server to backfill).
```

tests/federation_room_join_test.go
MadLittleMods, line 14: Has all of the review been addressed here? Please re-request review if so (I assume GitHub allows you to do that)

I don't see replies, 👍 reactions, or discussions resolved on everything. And the rebase makes it harder to follow.

tests/federation_room_join_test.go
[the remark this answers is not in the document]
FrenchGithubUser: done

tests/federation_room_join_test.go
FrenchGithubUser: done

tests/federation_room_join_test.go
FrenchGithubUser: done

tests/federation_room_join_test.go
FrenchGithubUser: done

tests/federation_room_join_test.go
FrenchGithubUser: done

tests/federation_room_join_test.go
FrenchGithubUser: done

tests/federation_room_join_test.go
FrenchGithubUser: yes my bad, all your comments have been addressed

tests/federation_room_join_test.go
```diff
@@ -574,6 +575,151 @@ func TestSendJoinPartialStateResponse(t *testing.T) {
...
+	// transactions once we've joined. We use a raw handler because the
+	// Complement server is not fully in the room until send_join completes, so
+	// we can't use HandleTransactionRequests (which requires the room in
+	// srv.rooms). Instead we parse the raw transaction body ourselves.
+	srv.Mux().Handle("/_matrix/federation/v1/send/{transactionID}", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		body, err := io.ReadAll(req.Body)
+		if err != nil {
+			t.Fatalf("failed to read request body in /send handler: %v", err)
+			return
+		}
```
MadLittleMods: We can use `must.NotError(...)` here

tests/federation_room_join_test.go
MadLittleMods: This was updated to the following:

```
			// Check if this event's prev_events directly reference the message
			// (e.g. a dummy event tying the two forward extremities together).
			// If so, the joining server can backfill from that event and will
			// discover the message. We only check one level of prev_events:
			// if the reference is deeper in the DAG the joining server can
			// still reach the message through backfill.
```

This sounds pretty good now but I think it needs a slight adjustment to set the right expecations:

```
			// Check if this event's prev_events directly reference the message (e.g. a dummy
			// event tying the two forward extremities together). If so, the joining server
			// can backfill from that event and will discover the message.
			//
			// XXX: We only check one level of prev_events: if the reference is deeper in the
			// DAG, it's valid and the joining server can still reach the message through
			// backfill but our checks don't account for that yet (feel free to edit this
			// assertion if you run into this)
```

tests/federation_room_join_test.go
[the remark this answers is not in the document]
MadLittleMods: I'm not a Go expert so I don't know for sure whether this is relevant but it sounds plausible.

Based on https://tip.golang.org/ref/mem, without synchronization, writes are not guaranteed to be observed by other goroutines

> the assignment to a is not followed by any synchronization event, so it is not guaranteed to be observed by any other goroutine. In fact, an aggressive compiler might delete the entire go statement.
> 
> If the effects of a goroutine must be observed by another goroutine, use a synchronization mechanism such as a lock or channel communication to establish a relative ordering.

MadLittleMods (commented):
Thanks for the updates so far @FrenchGithubUser! This is looking close

FrenchGithubUser:
Had a few hiccups with git, I hope I didn't mess something up, but looks fine

tests/federation_room_join_test.go
```diff
@@ -574,6 +576,153 @@ func TestSendJoinPartialStateResponse(t *testing.T) {
...
+
+	alice := deployment.Register(t, "hs1", helpers.RegistrationOpts{})
+
+	// We track the message event ID sent between make_join and send_join.
+	// After send_join, we wait for hs1 to send us either:
+	//  - the message event itself, or
+	//  - any event whose prev_events reference the message (e.g. a dummy event)
+	// atomic.Value is used because messageEventID is written on the main goroutine
+	// and read on the HTTP handler goroutine, with no other synchronization.
+	var messageEventID atomic.Value
```
MadLittleMods:
```suggestion
    //
	// atomic.Value is used because messageEventID is written on the main goroutine and
	// read on the HTTP handler goroutine, and needs synchronization (without
	// synchronization, writes are not guaranteed to be observed by other goroutines)
	var messageEventID atomic.Value
```

tests/federation_room_join_test.go
```diff
@@ -574,6 +576,153 @@ func TestSendJoinPartialStateResponse(t *testing.T) {
...
+// This test verifies that events sent into a room between a /make_join and
+// /send_join are not lost to the joining server. When an event is created
+// during the join handshake, the join event's prev_events (set at make_join
+// time) won't reference it, creating two forward extremities. The server
+// handling the join should ensure the joining server can discover the missed
+// event, for example by sending a follow-up event that references both
+// extremities, prompting the joining server to backfill.
+//
+// See https://github.com/element-hq/synapse/pull/19390
+func TestEventBetweenMakeJoinAndSendJoinIsNotLost(t *testing.T) {
```
MadLittleMods, line 588: Thanks for the updates @FrenchGithubUser. The test looks good now. Just a case of deciding where it should go:

Discussion for whether this should live in the Complement repo or as a in-repo Complement test, https://github.com/element-hq/synapse/pull/19390#discussion_r3164501605

tests/federation_room_join_test.go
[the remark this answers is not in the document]
MadLittleMods: `atomic.Value` was added :white_check_mark:

tests/federation_room_join_test.go
anoadragon453: I can't load the linked discussion (due to GitHub bugs...) but it seems as if we've decided to move the test back to Synapse's in-repo Complement tests.

Thus, is this PR still necessary?

MadLittleMods: > Thus, is this PR still necessary?

PR is not necessary :fast_forward:

PR #849, opened 2026-03-12
Add tests for MSC4429: Profile Updates for Legacy Sync

anoadragon453:
This PR introduces some initial Complement tests for the current draft of [MSC4429: Profile Updates for Legacy Sync](https://github.com/matrix-org/matrix-spec-proposals/pull/4429).

**Note:** This PR was almost entirely written by OpenAI Codex, but has been reviewed manually. Extra review scrutiny is advised.

Tests are expected to fail until the associated homeserver implementation has been written.

### Pull Request Checklist


tests/msc4429/msc4429_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+	"github.com/matrix-org/complement"
+	"github.com/matrix-org/complement/client"
+	"github.com/matrix-org/complement/helpers"
+	"github.com/matrix-org/complement/match"
+	"github.com/matrix-org/complement/must"
+)
+
+const (
+	msc4429UsersStable   = "users"
```
MadLittleMods, line 18: Seems pre-mature to support stable given https://github.com/matrix-org/matrix-spec-proposals/pull/4429 is no where near that point 🤔 

Perhaps we just comment that part out for now. Useful for our future reference when we stabilize

tests/msc4429/msc4429_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+		bob.MustSetDisplayName(t, "Bob Display")
+		mustSetProfileField(t, bob, "m.status", map[string]interface{}{
+			"text":  "busy",
+			"emoji": "🛑",
+		})
+
+		// Exclude 'displayname'
+		filter := mustBuildMSC4429Filter(t, []string{"m.status"})
+		res, _ := alice.MustSync(t, client.SyncReq{Filter: filter})
```
MadLittleMods, line 40: Might be flaky if we don't use a sync until utility for profile updates to show up. (read after write consistency)

The worker processing profile updates might not be the same one that serves `/sync`.

tests/msc4429/msc4429_test.go
```diff
@@ -0,0 +1,207 @@
...
+			t.Fatalf("missing m.status profile update for %s in initial sync: %s", bob.UserID, res.Raw)
+		}
+		must.MatchGJSON(t, update, match.JSONKeyEqual("", map[string]interface{}{
+			"text":  "busy",
+			"emoji": "🛑",
+		}))
+		assertNoProfileUpdate(t, res, bob.UserID, "displayname")
+	})
+
+	t.Run("No updates without profile_fields filter", func(t *testing.T) {
```
MadLittleMods, line 53: Explain why, "opt-in" per [MSC4429](https://github.com/matrix-org/matrix-spec-proposals/pull/4429)

tests/msc4429/msc4429_test.go
```diff
@@ -0,0 +1,207 @@
...
+		alice := deployment.Register(t, "hs1", helpers.RegistrationOpts{LocalpartSuffix: "alice-nofilter"})
+		bob := deployment.Register(t, "hs1", helpers.RegistrationOpts{LocalpartSuffix: "bob-nofilter"})
+
+		mustCreateSharedRoom(t, alice, bob)
+
+		// No filter = no profile fields returned.
+		_, since := alice.MustSync(t, client.SyncReq{})
+		mustSetProfileField(t, bob, "m.status", map[string]interface{}{
+			"text": "away",
+		})
```
MadLittleMods, line 63: As a sanity check, we should have another observer user that sees the `m.status` with a filter, and then `alice` syncs without a filter

MadLittleMods, line 63: Separate action from previous request

```suggestion

		mustSetProfileField(t, bob, "m.status", map[string]interface{}{
			"text": "away",
		})
```

tests/msc4429/msc4429_test.go
```diff
@@ -0,0 +1,207 @@
...
+	})
+
+	t.Run("No updates without profile_fields filter", func(t *testing.T) {
+		alice := deployment.Register(t, "hs1", helpers.RegistrationOpts{LocalpartSuffix: "alice-nofilter"})
+		bob := deployment.Register(t, "hs1", helpers.RegistrationOpts{LocalpartSuffix: "bob-nofilter"})
+
+		mustCreateSharedRoom(t, alice, bob)
+
+		// No filter = no profile fields returned.
+		_, since := alice.MustSync(t, client.SyncReq{})
```
MadLittleMods, line 60: We don't actually assert anything for the initial sync request

tests/msc4429/msc4429_test.go
```diff
@@ -0,0 +1,207 @@
...
+		mustSetProfileField(t, bob, "m.status", map[string]interface{}{
+			"text": "away",
+		})
+
+		res, _ := alice.MustSync(t, client.SyncReq{Since: since})
+		assertNoProfileUpdate(t, res, bob.UserID, "m.status")
+	})
+
+	t.Run("Incremental sync returns the latest update", func(t *testing.T) {
```
MadLittleMods, line 69: Perhaps we should have another test to make sure more subsequent incremental `/sync` include new updates for the new user.

This ensures that any caching (remembering which users were sent down before) is correct.

MadLittleMods: I guess, we kinda have this already with "Cleared profile field is returned as null"

tests/msc4429/msc4429_test.go
```diff
@@ -0,0 +1,207 @@
...
+	if unstableRes.Exists() {
+		return unstableRes, true
+	}
+	return gjson.Result{}, false
+}
+
+func assertNoProfileUpdate(t *testing.T, res gjson.Result, userID, field string) {
+	t.Helper()
+	if update, ok := getProfileUpdate(res, userID, field); ok {
+		t.Fatalf("unexpected profile update for %s %s: %s", userID, field, update.Raw)
```
MadLittleMods, line 186: This is correct but the language is ambiguous whether we expected no profile update or just a different profile update.

tests/msc4429/msc4429_test.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/matrix-org/complement/match"
+	"github.com/matrix-org/complement/must"
+)
+
+const (
+	msc4429UsersStable   = "users"
+	msc4429UsersUnstable = "org\\.matrix\\.msc4429\\.users"
+)
+
+func TestMSC4429ProfileUpdates(t *testing.T) {
```
anoadragon453, line 22: @Half-Shot [mentioned](https://github.com/matrix-org/matrix-spec-proposals/pull/4429/changes#r3011490866) on the MSC that it would be good for the homeserver to send down `null`s to clear a user's profile once they no longer share any rooms with you (after leaving a room).

It would be good to add a test case for this, ideally replacing the profile object with just a `null` as mentioned in [my follow-up comment](https://github.com/matrix-org/matrix-spec-proposals/pull/4429/changes#r3015338850), to indicate "all fields should be cleared".

tests/msc4429/msc4429_test.go
anoadragon453: > I guess, we kinda have this already with "Cleared profile field is returned as null"

Sort of. The data structure of the `/sync` response JSON doesn't allow to return two subsequent status updates for the same user and field:

```jsonc
{
  "users": {
    "@user:example.org": {
      "profile_updates": {
        "m.status": {
          "text": "Swimming in the Great Lakes!",
          "emoji": "🏊️"
        }
      }
    }
  }
}
```

What might be useful is an additional test with:

* Bob updating their `m.status` field
* Alice syncing without a sync filter which includes the field
* Bob updating their `displayname` field
* Alice syncing with a `m.status` and `displayname` filter
* Alice should only receive the `displayname` update

tests/msc4429/msc4429_test.go
anoadragon453: I believe this is just to get a since token, so we can perform an incremental sync afterwards.

Clarified the comments in 4b93b71a59ca431a1f10cefe714c44f2384055bc.

tests/msc4429/msc4429_test.go
anoadragon453: Addressed in fd4b5fd84fd325aa9c922ff70fe534ba0c1e7be4.

tests/msc4429/msc4429_test.go
anoadragon453: Good point. Clarified in 0d22df0e43f004a515188007014ecd1af63678d2.

tests/msc4429/msc4429_test.go
anoadragon453: D'oh, and yet we use it in all the other tests.

Fixed in be51f167e40c83910cf10fe2eb5d48688543263d.

anoadragon453:
The tests in this PR weren't actually running in CI :facepalm: I've fixed that in https://github.com/matrix-org/complement/pull/849/commits/4cddfeab619504b8ed4a9c86df3892953cb63c75.

All were expected to succeed (despite all failing), other than the newly added `A user leaving the last shared room returns a profile update of null` which @jaywink is currently working on. Presumably the Synapse implementation needs adapting to fix the rest of the tests as well.

Still, the tests here should be ready for another round of review.

tests/msc4429/msc4429_test.go
anoadragon453: Done in https://github.com/matrix-org/complement/pull/849/commits/4cddfeab619504b8ed4a9c86df3892953cb63c75.

Associated MSC update: https://github.com/matrix-org/matrix-spec-proposals/commit/5ab7aede

.github/workflows/ci.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
           - homeserver: Synapse
             repo: element-hq/synapse
             tags: synapse_blacklist
-            packages: ./tests/msc3874 ./tests/msc3902 ./tests/msc4306
+            packages: ./tests/msc3874 ./tests/msc3902 ./tests/msc4306 ./tests/msc4429
```
MadLittleMods, line 55: > The tests in this PR weren't actually running in CI :facepalm: I've fixed that in https://github.com/matrix-org/complement/pull/849/commits/4cddfeab619504b8ed4a9c86df3892953cb63c75.

I assume you meant https://github.com/matrix-org/complement/pull/849/changes/2b67137ea2b3c495584c231b6313c7389ec6b4b8 but in any case, we don't really worry about updating this list.

It's already heavily out of sync with the list we run in the Synapse project, [`element-hq/synapse` -> `scripts-dev/complement.sh#L272-L289`](https://github.com/element-hq/synapse/blob/16fe5fb46235eac86a343712f7b3554742955218/scripts-dev/complement.sh#L272-L289)

At this point, I think we just rely on it as a general sanity check that Complement is still working with Synapse.

tests/msc4429/msc4429_test.go
```diff
@@ -0,0 +1,305 @@
...
+			"emoji": "🟡",
+		})
+
+		// Assert that charlie receives bob's profile update in an incremental sync
+		// with the appropriate filter set.
+		filter := mustBuildMSC4429Filter(t, []string{"m.status"})
+		charlie.MustSyncUntil(t, client.SyncReq{Filter: filter}, syncHasProfileUpdate(bob.UserID, "m.status", map[string]interface{}{
+			"text":  "away",
+			"emoji": "🟡",
+		}))
```
MadLittleMods, line 78: Kinda feels like this should also be an incremental sync to better demonstrate the data is there with the right filter.

tests/msc4429/msc4429_test.go
```diff
@@ -0,0 +1,305 @@
...
+		if err := match.JSONKeyEqual("", expected)(update); err != nil {
+			return fmt.Errorf("profile update mismatch for %s %s: %w", userID, field, err)
+		}
+		return nil
+	}
+}
+
+// syncHasProfileUpdatesNull checks whether a sync response contains a null
+// profile_updates value for the given user.
+func syncHasProfileUpdatesNull(userID string) client.SyncCheckOpt {
```
MadLittleMods, line 354: Seems like the usage could be replaced by `syncHasProfileUpdate(...)`

Or at-least this function re-uses `syncHasProfileUpdate(...)`

MadLittleMods (approved):

tests/msc4429/msc4429_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+	// Use one of the users to create the room.
+	roomID := users[0].MustCreateRoom(t, map[string]interface{}{
+		"preset": "public_chat",
+	})
+
+	// Join all of the given users to the room.
+	for _, user := range users {
+		user.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(user.UserID, roomID))
+		user.MustJoinRoom(t, roomID, nil)
```
jaywink, line 261: This should be before the `user.MustSyncUntil` afaict.

tests/msc4429/msc4429_test.go
```diff
@@ -0,0 +1,305 @@
...
+			"emoji": "🛑",
+		})
+
+		// Alice /sync's, but only asks for "m.status" changes.
+		// Exclude 'displayname'.
+		filter := mustBuildMSC4429Filter(t, []string{"m.status"})
+		res, _ := alice.MustSync(t, client.SyncReq{Filter: filter})
+
+		// We should see the m.status profile update.
+		alice.MustSyncUntil(t, client.SyncReq{Filter: filter}, syncHasProfileUpdate(alice.UserID, "m.status", map[string]interface{}{
```
jaywink:
```suggestion
		alice.MustSyncUntil(t, client.SyncReq{Filter: filter}, syncHasProfileUpdate(bob.UserID, "m.status", map[string]interface{}{
```

This should check for bob's update, not alice's.

jaywink (changes requested):

.github/workflows/ci.yaml
anoadragon453: I was planning to update it to match Synapse's list in a separate PR. Seems useful to prevent Complement changes breaking Synapse CI?

tests/msc4429/msc4429_test.go
anoadragon453: Mm, it's a bit different as `syncHasProfileUpdate` expects a field name and an expected value. Where in the `null` case, the whole `profile_updates` dict is replaced by a `null` value.

So you'd probably need to add a fourth argument to `syncHasProfileUpdate` to indicate that you want to check for `null`... and that's starting to feel a bit messy to me.

tests/msc4429/msc4429_test.go
anoadragon453: I added this test in https://github.com/matrix-org/complement/pull/849/commits/897e7ab59d3911942cfb5c8648328a0105932dd0, and it passes the existing Synapse code :tada:

anoadragon453:
This PR is good to go from my perspective.

I'm now just waiting for the associated Synapse PR to be approved, before we merge it (plus the associated [sytest PR](https://github.com/matrix-org/sytest/pull/1420)) all together.

tests/msc4429/main_test.go
```diff
@@ -0,0 +1,11 @@
+package tests
```
MadLittleMods, line 1: > This PR is good to go from my perspective.
> 
> I'm now just waiting for the associated Synapse PR to be approved, before we merge it (plus the associated https://github.com/matrix-org/sytest/pull/1420) all together.
>
> *-- @anoadragon453, https://github.com/matrix-org/complement/pull/849#issuecomment-4792573477*

The Synapse PR was merged https://github.com/element-hq/synapse/pull/19556 but this hasn't been merged yet.

This is causing the Synapse CI to fail at the moment :x:, see https://github.com/element-hq/synapse/issues/20076

jaywink (approved):

jaywink:
Merging as synapse side has been merged.

PR #850, opened 2026-03-13
Stop testing unspecced POST /send

tulir:
For https://github.com/element-hq/synapse/pull/19557

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -118,7 +118,7 @@ jobs:
 
       - run: |
           set -o pipefail &&
-          ${{ matrix.env }} go test -v -json -tags "${{ matrix.tags }}" -timeout "${{ matrix.timeout }}" ./tests ./tests/csapi ${{ matrix.packages }} | .ci/scripts/gotestfmt
+          ${{ matrix.env }} go test -v -json -tags "${{ matrix.tags }}" -timeout "${{ matrix.timeout }}" ./tests ./tests/csapi ${{ matrix.packages }} # | .ci/scripts/gotestfmt
         shell: bash # required for pipefail to be A Thing. pipefail is required to stop gotestfmt swallowing non-zero exit codes
         name: Run Complement Tests
         env:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -118,7 +118,7 @@ jobs:
 
       - run: |
           set -o pipefail &&
-          ${{ matrix.env }} go test -v -json -tags "${{ matrix.tags }}" -timeout "${{ matrix.timeout }}" ./tests ./tests/csapi ${{ matrix.packages }} # | .ci/scripts/gotestfmt
+          ${{ matrix.env }} go test -v -json -tags "${{ matrix.tags }}" -timeout "${{ matrix.timeout }}" ./tests ./tests/csapi ${{ matrix.packages }} | .ci/scripts/gotestfmt
         shell: bash # required for pipefail to be A Thing. pipefail is required to stop gotestfmt swallowing non-zero exit codes
         name: Run Complement Tests
         env:
--- a/helpers/txnid.go
+++ b/helpers/txnid.go
@@ -9,6 +9,8 @@ import (
 var txnCounter atomic.Int64
 var start = time.Now().Unix()
 
+// GetTxnID generates a unique transaction ID for use with endpoints like /send/{eventType/{txnId}.
+// The prefix is only for debugging purposes.
 func GetTxnID(prefix string) (txnID string) {
 	return fmt.Sprintf("%s-%d-%d", prefix, start, txnCounter.Add(1))
 }
--- a/helpers/txnid.go
+++ b/helpers/txnid.go
@@ -1,0 +1,14 @@
+package helpers
+
+import (
+	"fmt"
+	"sync/atomic"
+	"time"
+)
+
+var txnCounter atomic.Int64
+var start = time.Now().Unix()
+
+func GetTxnID(prefix string) (txnID string) {
+	return fmt.Sprintf("%s-%d-%d", prefix, start, txnCounter.Add(1))
+}
--- a/tests/csapi/invalid_test.go
+++ b/tests/csapi/invalid_test.go
@@ -1,7 +1,6 @@
 package csapi_tests
 
 import (
-	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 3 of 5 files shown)


tulir:
Not sure what that test failure is 🤔

tests/csapi/invalid_test.go
```diff
@@ -56,8 +57,8 @@ func TestJson(t *testing.T) {
 				[]byte(`{"body": NaN}`),
 			}
 
-			for _, testCase := range testCases {
-				res := alice.Do(t, "POST", []string{"_matrix", "client", "v3", "rooms", roomID, "send", "complement.dummy"}, client.WithJSONBody(t, testCase))
+			for i, testCase := range testCases {
+				res := alice.Do(t, "POST", []string{"_matrix", "client", "v3", "rooms", roomID, "send", "complement.dummy", fmt.Sprintf("invalidval-%d", i)}, client.WithJSONBody(t, testCase))
```
tulir: Missed this one

```suggestion
				res := alice.Do(t, "PUT", []string{"_matrix", "client", "v3", "rooms", roomID, "send", "complement.dummy", fmt.Sprintf("invalidval-%d", i)}, client.WithJSONBody(t, testCase))
```

anoadragon453:
> Not sure what that test failure is 🤔

`panic: BUG: Empty package name encountered.` will occur when `gotestfmt` receives an error and isn't able to parse the output. Try temporarily removing ` | .ci/scripts/gotestfmt` from the following line in CI:

https://github.com/matrix-org/complement/blob/12210e3e418b12c62bac5868d06aa72a9cfc5119/.github/workflows/ci.yaml#L121

(I fixed this in a repo somewhere to prevent this obtuse error from shadowing the real one, but currently failing to find it or the solution...)

.github/workflows/ci.yaml
```diff
@@ -118,7 +118,7 @@ jobs:
 
       - run: |
           set -o pipefail &&
-          ${{ matrix.env }} go test -v -json -tags "${{ matrix.tags }}" -timeout "${{ matrix.timeout }}" ./tests ./tests/csapi ${{ matrix.packages }} | .ci/scripts/gotestfmt
+          ${{ matrix.env }} go test -v -json -tags "${{ matrix.tags }}" -timeout "${{ matrix.timeout }}" ./tests ./tests/csapi ${{ matrix.packages }} # | .ci/scripts/gotestfmt
```
MadLittleMods: > `panic: BUG: Empty package name encountered.` will occur when `gotestfmt` receives an error and isn't able to parse the output. Try temporarily removing ` | .ci/scripts/gotestfmt` from the following line in CI:
>
> *-- @anoadragon453, https://github.com/matrix-org/complement/pull/850#issuecomment-4077117196*

We did this in Synapse for example, https://github.com/element-hq/synapse/pull/19326

tulir:
It was just a compile error, tests should pass now

tests/csapi/invalid_test.go
```diff
@@ -34,8 +35,8 @@ func TestJson(t *testing.T) {
 				[]byte(`{"body": 1.1}`),
 			}
 
-			for _, testCase := range testCases {
-				res := alice.Do(t, "POST", []string{"_matrix", "client", "v3", "rooms", roomID, "send", "complement.dummy"}, client.WithJSONBody(t, testCase))
+			for i, testCase := range testCases {
+				res := alice.Do(t, "PUT", []string{"_matrix", "client", "v3", "rooms", roomID, "send", "complement.dummy", fmt.Sprintf("invalidnum-%d", i)}, client.WithJSONBody(t, testCase))
```
MadLittleMods: Spec for reference: [`PUT /_matrix/client/v3/rooms/{roomId}/send/{eventType}/{txnId}`](https://spec.matrix.org/v1.16/client-server-api/#put_matrixclientv3roomsroomidsendeventtypetxnid)

MadLittleMods: The "invalidnum" makes me think this a particular detail that matters. But it's actually the invalid number `testCases` above that matter.

Can we just use `txn-xxx` instead? I'm guessing we're just trying to namespace them to avoid tests colliding with each other. Perhaps we can just do this like we have in other tests:

https://github.com/matrix-org/complement/blob/12210e3e418b12c62bac5868d06aa72a9cfc5119/tests/room_timestamp_to_event_test.go#L256-L264

tests/csapi/room_messages_test.go
```diff
@@ -37,8 +38,8 @@ func TestSendAndFetchMessage(t *testing.T) {
 
 	_, token := alice.MustSync(t, client.SyncReq{})
 
-	// first use the non-txn endpoint
-	alice.MustDo(t, "POST", []string{"_matrix", "client", "v3", "rooms", roomID, "send", "m.room.message"}, client.WithJSONBody(t, map[string]interface{}{
+	// first use the send endpoint
+	alice.MustDo(t, "PUT", []string{"_matrix", "client", "v3", "rooms", roomID, "send", "m.room.message", strconv.FormatInt(time.Now().UnixNano(), 10)}, client.WithJSONBody(t, map[string]interface{}{
```
MadLittleMods: This could result in duplicate transactions as far as I can tell. We should probably use something that increments.

tests/csapi/room_messages_test.go
```diff
@@ -22,7 +23,7 @@ import (
 	"github.com/matrix-org/gomatrixserverlib/spec"
 )
 
-// sytest: POST /rooms/:room_id/send/:event_type sends a message
+// sytest: PUT /rooms/:room_id/send/:event_type/:txn_id sends a message
```
MadLittleMods: The `xxx` part of `sytest: xxx` part might matter and need to match the name in Sytest?

tulir: yeah sytest has that test: https://github.com/matrix-org/sytest/blob/1be04cce46c0f84abac736390dc3fab17f35a756/tests/10apidoc/34room-messages.pl#L23

tests/csapi/invalid_test.go
tulir: Extracted all the transaction ids (this one, the comment below and the old `getTxnID`) into a new method in the helpers package

tests/csapi/room_messages_test.go
MadLittleMods: Given this test is being removed in https://github.com/matrix-org/sytest/pull/1419, I think we can just remove this comment entirely

tests/csapi/room_messages_test.go
MadLittleMods: Code was updated to use `helpers.GetTxnID(...)`

helpers/txnid.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"fmt"
+	"sync/atomic"
+	"time"
+)
+
+var txnCounter atomic.Int64
+var start = time.Now().Unix()
+
+func GetTxnID(prefix string) (txnID string) {
```
MadLittleMods, line 14: We should add a doc comment for what this does and is useful for

MadLittleMods (commented):
Looks good, just a few clean-up items

tests/csapi/room_messages_test.go
tulir: Oh right, there's already another test below for the transaction ID version

tulir: Removed the comment, left the test just for the /messages testing

helpers/txnid.go
MadLittleMods: Done in https://github.com/matrix-org/complement/pull/850/commits/5de6121a8057828c4a4bb8ef71c178d3bc35e03c

MadLittleMods (approved):

MadLittleMods:
Thanks for cleaning up the test suite @tulir 🐻‍❄️

PR #852, opened 2026-03-26
`TestJumpToDateEndpoint`: Clarify why we use `/context` `end` for paginating backwards despite the spec saying "A token that can be used to paginate forwards with."

MadLittleMods:
Clarify why we use `/context` `end` for paginating backwards despite the spec saying ["A token that can be used to paginate forwards with."](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv3roomsroomidcontexteventid).

There is no functional change to these tests.

This is spawning from [using `/context` in another Complement test](https://github.com/element-hq/synapse/pull/19611) and noticing this oddity.

### Pull Request Checklist


tests/room_timestamp_to_event_test.go
```diff
@@ -213,10 +213,20 @@ func TestJumpToDateEndpoint(t *testing.T) {
...
+				)
 				contextResResBody := client.ParseJSON(t, contextRes)
+				// Remember: Tokens are positions between events. Normally, you would use the
+				// `start` token to paginate backwards with but for the sake of the test we want
+				// paginate `/messages` and see both A and B in the response; so we use the
+				// `end` token. The `end` token comes after B.
+				//
+				//          start   end
+				//          |       |
+				// [A] <--  ▼  [B]  ▼  <--- [remoteCharlie join]
```
MadLittleMods, line 229: Inspired by these [comments in Synapse](https://github.com/element-hq/synapse/blob/539f708f32af4cc01f6d82f1bb0069ff38e93593/synapse/types/__init__.py#L723-L727)

tests/room_timestamp_to_event_test.go
```diff
@@ -213,10 +213,20 @@ func TestJumpToDateEndpoint(t *testing.T) {
...
 
 				// Get a pagination token from eventB
-				contextRes := remoteCharlie.MustDo(t, "GET", []string{"_matrix", "client", "r0", "rooms", roomID, "context", eventB.EventID}, client.WithContentType("application/json"), client.WithQueries(url.Values{
-					"limit": []string{"0"},
-				}))
+				contextRes := remoteCharlie.MustDo(t, "GET", []string{"_matrix", "client", "r0", "rooms", roomID, "context", eventB.EventID},
+					client.WithContentType("application/json"), client.WithQueries(url.Values{
+						"limit": []string{"0"},
+					}),
+				)
```
MadLittleMods, line 220: Just line wrapping this better. There is no functional change in this PR.

tests/room_timestamp_to_event_test.go
```diff
@@ -226,12 +236,15 @@ func TestJumpToDateEndpoint(t *testing.T) {
 					return ev.Get("event_id").Str == eventA.EventID
 				})
 
-				// Paginate backwards from eventB
```
MadLittleMods, line 229: "Paginate backwards from eventB" is a bit of a misnomer as we paginate backwards from the position just after B. Simplified the comment ⏩ as we already have an explanation of the `paginationToken` above.

tests/room_timestamp_to_event_test.go
```diff
@@ -213,10 +213,20 @@ func TestJumpToDateEndpoint(t *testing.T) {
...
-					"limit": []string{"0"},
-				}))
+				contextRes := remoteCharlie.MustDo(t, "GET", []string{"_matrix", "client", "r0", "rooms", roomID, "context", eventB.EventID},
+					client.WithContentType("application/json"), client.WithQueries(url.Values{
+						"limit": []string{"0"},
+					}),
+				)
 				contextResResBody := client.ParseJSON(t, contextRes)
+				// Remember: Tokens are positions between events. Normally, you would use the
+				// `start` token to paginate backwards with but for the sake of the test we want
```
reivilibre, line 223:
```suggestion
				// `start` token to paginate backwards with but for the sake of the test we want to
```

tests/room_timestamp_to_event_test.go
reivilibre: maybe 'Paginate backwards from the point after eventB' ? But maybe it's just getting unnecessarily clumsy.

reivilibre (approved):

tests/room_timestamp_to_event_test.go
MadLittleMods: Updated to use your wording :fast_forward:

MadLittleMods:
Thanks for the review @reivilibre 🦇

PR #858, opened 2026-04-16
feat: support for room v11

FrenchGithubUser:
- Rooms v11 and above don't have the `creator` field anymore during room creation, which broke the tests.
- Synapse changed the way it counted the `state_key` length in events (from codepoints to bytes) at room v11, which broke the tests.

This PR fixes those issues, making complement work with room v11 as default.

Related to https://github.com/element-hq/synapse/pull/18680

### DIFF
--- a/federation/server_room.go
+++ b/federation/server_room.go
@@ -333,15 +333,20 @@ func InitialRoomEvents(roomVer gomatrixserverlib.RoomVersion, creator string) []
 			Type:     "m.room.create",
 			StateKey: b.Ptr(""),
 			Sender:   creator,
-			Content: map[string]interface{}{
-				"creator":      creator,
-				"room_version": roomVer,
-				// We have to add randomness to the create event, else if you create 2x v12+ rooms in the same millisecond
-				// they will get the same room ID, clobbering internal data structures and causing extremely confusing
-				// behaviour. By adding this entropy, we ensure that even if rooms are created in the same millisecond, their
-				// hashes will not be the same.
-				"complement_entropy": util.RandomString(18),
-			},
+			Content: func() map[string]interface{} {
+				content := map[string]interface{}{
+					"room_version": roomVer,
+					// We have to add randomness to the create event, else if you create 2x v12+ rooms in the same millisecond
+					// they will get the same room ID, clobbering internal data structures and causing extremely confusing
+					// behaviour. By adding this entropy, we ensure that even if rooms are created in the same millisecond, their
+					// hashes will not be the same.
+					"complement_entropy": util.RandomString(18),
+				}
+				if gomatrixserverlib.MustGetRoomVersion(gomatrixserverlib.RoomVersion(roomVer)).CreatorInCreateEvent() {
+					content["creator"] = creator
+				}
+				return content
+			}(),
 		},
 		{
 			Type:     "m.room.member",
--- a/federation/server_room.go
+++ b/federation/server_room.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"strconv"
 	"sync"
 	"time"
 
@@ -333,15 +334,22 @@ func InitialRoomEvents(roomVer gomatrixserverlib.RoomVersion, creator string) []
 			Type:     "m.room.create",
 			StateKey: b.Ptr(""),
 			Sender:   creator,
-			Content: map[string]interface{}{
-				"creator":      creator,
-				"room_version": roomVer,
-				// We have to add randomness to the create event, else if you create 2x v12+ rooms in the same millisecond
-				// they will get the same room ID, clobbering internal data structures and causing extremely confusing
-				// behaviour. By adding this entropy, we ensure that even if rooms are created in the same millisecond, their
-				// hashes will not be the same.
-				"complement_entropy": util.RandomString(18),
-			},
+			Content: func() map[string]interface{} {
+				content := map[string]interface{}{
+					"room_version": roomVer,
+					// We have to add randomness to the create event, else if you create 2x v12+ rooms in the same millisecond
+					// they will get the same room ID, clobbering internal data structures and causing extremely confusing
+					// behaviour. By adding this entropy, we ensure that even if rooms are created in the same millisecond, their
+					// hashes will not be the same.
+					"complement_entropy": util.RandomString(18),
+				}
+				// The creator field was removed in room version 11 (MSC4239).
+				n, err := strconv.Atoi(string(roomVer))
+				if err != nil || n < 11 {
+					content["creator"] = creator
+				}
+				return content
+			}(),
 		},
 		{
 			Type:     "m.room.member",
--- a/tests/csapi/rooms_state_test.go
+++ b/tests/csapi/rooms_state_test.go
@@ -4,7 +4,6 @@
 package csapi_tests
 
 import (
-	"strconv"
 	"testing"
 	"time"
 
@@ -16,6 +15,7 @@ import (
 	"github.com/matrix-org/complement/helpers"
 	"github.com/matrix-org/complement/match"
 	"github.com/matrix-org/complement/must"
+	"github.com/matrix-org/gomatrixserverlib"
 )
 
 func TestRoomCreationReportsEventsToMyself(t *testing.T) {
@@ -28,7 +28,6 @@ func TestRoomCreationReportsEventsToMyself(t *testing.T) {
 		Password:        "bobpassword",
 	})
 	defaultVer := alice.GetDefaultRoomVersion(t)
-	defaultVerN, _ := strconv.Atoi(string(defaultVer))
 	roomID := alice.MustCreateRoom(t, map[string]interface{}{})
 
 	t.Run("parallel", func(t *testing.T) {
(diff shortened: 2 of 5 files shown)


kegsay (approved):
LGTM, thank you!

federation/server_room.go
```diff
@@ -333,15 +334,22 @@ func InitialRoomEvents(roomVer gomatrixserverlib.RoomVersion, creator string) []
...
+					// they will get the same room ID, clobbering internal data structures and causing extremely confusing
+					// behaviour. By adding this entropy, we ensure that even if rooms are created in the same millisecond, their
+					// hashes will not be the same.
+					"complement_entropy": util.RandomString(18),
+				}
+				// The creator field was removed in room version 11 (MSC4239).
+				n, err := strconv.Atoi(string(roomVer))
+				if err != nil || n < 11 {
+					content["creator"] = creator
+				}
```
MadLittleMods, line 350: Feels like something that should be handled by this kind of thing:

```
iRoomVer := gomatrixserverlib.MustGetRoomVersion(roomVer)
if iRoomVer.CreatorInCreateEvent() {
```

`CreatorInCreateEvent` doesn't exist yet but aligns with the other things we have there like [`DomainlessRoomIDs`](https://github.com/matrix-org/gomatrixserverlib/blob/20c9de33969e2df883eaa69c283d6a38efce9284/eventversion.go#L39) and [`PrivilegedCreators`](https://github.com/matrix-org/gomatrixserverlib/blob/20c9de33969e2df883eaa69c283d6a38efce9284/eventversion.go#L40)

See https://github.com/matrix-org/complement/pull/840 for how v12 rooms were added

federation/server_room.go
```diff
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"strconv"
 	"sync"
 	"time"
```
MadLittleMods, line 9: The sign-off check is currently failing. See [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off) docs

federation/server_room.go
FrenchGithubUser: that's a good idea! done, and also added a change there https://github.com/matrix-org/gomatrixserverlib/pull/465

federation/server_room.go
FrenchGithubUser: done! sorry for missing that

tests/csapi/rooms_state_test.go
```diff
@@ -38,7 +41,10 @@ func TestRoomCreationReportsEventsToMyself(t *testing.T) {
 					return false
 				}
 				must.Equal(t, ev.Get("sender").Str, alice.UserID, "wrong sender")
-				must.Equal(t, ev.Get("content").Get("creator").Str, alice.UserID, "wrong content.creator")
+				// The creator field was removed in room version 11 (MSC4239).
+				if defaultVerN < 11 {
+					must.Equal(t, ev.Get("content").Get("creator").Str, alice.UserID, "wrong content.creator")
+				}
```
MadLittleMods, line 47: Looks like we can use the same `CreatorInCreateEvent` check in this other spot as well

tests/federation_room_get_missing_events_test.go
```diff
@@ -497,6 +498,22 @@ func TestOutboundFederationEventSizeGetMissingEvents(t *testing.T) {
...
+	//
+	// Since upgrading a room (for example from v10 to v11) won't carry the event
+	// (a new room is created), we don't have to worry about v10 room events in a v11
+	// room.
+	//
+	// So this test is essentially skipped for any default room v11 or higher.
+	verNum, _ := strconv.Atoi(string(ver))
+	if verNum >= 11 {
+		ver = gomatrixserverlib.RoomVersion("10")
+	}
```
MadLittleMods, line 516: This seems like yet another case that we should handle with a new `assert iRoomVer.StrictEventByteLimits()` check

You can see how we do this in Synapse [`synapse/event_auth.py#L487-L499`](https://github.com/element-hq/synapse/blob/a9361c4f51ab35a7dd4955dfdb3db7c57ae2bc9b/synapse/event_auth.py#L487-L499)

MadLittleMods, line 516: Seems like this test would be better if we just always used room version 10 🤔 . The current setup is only better if we're trying to support homeservers that don't support room version 10 yet (do we care about that?)


```suggestion
	// This test crafts a "bad" event which state_key is 280 bytes but only 70
	// codepoints.
	//
	// Room version 11 in Synapse switched from using codepoints to using
	// bytes. Which means the 280-byte state_key would be rejected immediately.
	// Use room version 10 in that case so the codepoint-based limit is in effect.
	//
	// Since upgrading a room (for example from v10 to v11) won't carry the event
	// (a new room is created), we don't have to worry about v10 room events in a v11
	// room.
	//
	// So this test is essentially skipped for any default room v11 or higher.
	ver = gomatrixserverlib.RoomVersion("10")
```

federation/server_room.go
```diff
@@ -333,15 +333,20 @@ func InitialRoomEvents(roomVer gomatrixserverlib.RoomVersion, creator string) []
...
+			Content: func() map[string]interface{} {
+				content := map[string]interface{}{
+					"room_version": roomVer,
+					// We have to add randomness to the create event, else if you create 2x v12+ rooms in the same millisecond
+					// they will get the same room ID, clobbering internal data structures and causing extremely confusing
+					// behaviour. By adding this entropy, we ensure that even if rooms are created in the same millisecond, their
+					// hashes will not be the same.
+					"complement_entropy": util.RandomString(18),
+				}
+				if gomatrixserverlib.MustGetRoomVersion(gomatrixserverlib.RoomVersion(roomVer)).CreatorInCreateEvent() {
```
MadLittleMods, line 345: TODO: Wait for https://github.com/matrix-org/gomatrixserverlib/pull/465 to merge and update the `gomatrixserverlib` version in this PR

tests/federation_room_get_missing_events_test.go
FrenchGithubUser: yea I thought it's better to be flexible (since it's not a big change in the test code), but I can apply your suggestion if you think it's better

FrenchGithubUser: yes indeed, done!

tests/csapi/rooms_state_test.go
FrenchGithubUser: yes indeed, done!

tests/federation_room_get_missing_events_test.go
MadLittleMods: Probably simplest to converge on room version 10.

If there is a use case for the alternative, we can go that route but the comment should better explain the intention.

MadLittleMods (approved):
Changes look good 👍 

(waiting for https://github.com/matrix-org/gomatrixserverlib/pull/465 to merge)

federation/server_room.go
kegsay: `go get github.com/matrix-org/gomatrixserverlib@c9c468727353c4835523349dc7c202e5de9a2e37`

Though I can't seem to do it on this PR else I would.

FrenchGithubUser: done

FrenchGithubUser:
@MadLittleMods the gomatrixserverlib PR has been merged, I guess this PR can be merged now? :)

tests/federation_room_get_missing_events_test.go
```diff
@@ -497,6 +497,21 @@ func TestOutboundFederationEventSizeGetMissingEvents(t *testing.T) {
...
+	//
+	// Room version 11 in Synapse switched from using codepoints to using
+	// bytes. Which means the 280-byte state_key would be rejected immediately.
+	// Use room version 10 in that case so the codepoint-based limit is in effect.
+	//
+	// Since upgrading a room (for example from v10 to v11) won't carry the event
+	// (a new room is created), we don't have to worry about v10 room events in a v11
+	// room.
+	//
+	// So this test is essentially skipped for any default room v11 or higher.
```
MadLittleMods: Re-reading and it doesn't feel like these details matter. The last sentence doesn't read correctly with what's happening either. We run the test just not with v11 or higher and instead choose v10.

```suggestion
```

MadLittleMods: Are you able to allow edits by maintainers? It would be nice to just fix these little things up without waiting

FrenchGithubUser: ah yes, I think I updated the code without updating the comment

FrenchGithubUser: yea, I could accept the suggestion

MadLittleMods: For reference, `Allow edits by maintainers` is a checkbox in the right-side panel of the PR (at the bottom) that you can check.

MadLittleMods:
Thanks for the improvements and getting the `gomatrixserverlib` side all sorted @FrenchGithubUser :rocket:

tests/federation_room_get_missing_events_test.go
FrenchGithubUser: hmm, I don't see it, but I'll keep this in mind, thanks for the tip!

PR #796, opened 2025-08-15
Update MSC4311 tests to conform to the spec and better coverage

turt2live:
Update MSC4311 tests to conform to the spec. MSC4311 states that the client API should still use the stripped state event format. For the federation API, it should use full PDU's. MSC4311 also requires the `m.room.create` event be included in the list of stripped state.

["Stripped state"](https://spec.matrix.org/v1.18/client-server-api/#stripped-state) is a generic term that refers to simplified slice of state shared in `invite_room_state`/`knock_room_state` (`invite_state`/`knock_state` with `/sync`) before someone is joined to a room and can see the full room state.

This PR was originally authored by @turt2live taking a stab at fixing single the flawed test (self-reported: ~~"I have no idea what I'm doing" disclaimer goes here~~) but has since been taken over by @MadLittleMods. And has naturally evolved some new better test coverage as I've worked on a full solution.

Synapse PR: ~~https://github.com/element-hq/synapse/pull/18822~~ https://github.com/element-hq/synapse/pull/19723

The previous tests passed because of a flawed implementation in Synapse which is being removed in https://github.com/element-hq/synapse/pull/19723 as well.

### Dev notes

Running the tests with [Synapse](https://github.com/element-hq/synapse):

```
COMPLEMENT_DIR=../complement ./scripts-dev/complement.sh -run TestMSC4311StrippedStateClientAPI

COMPLEMENT_DIR=../complement ./scripts-dev/complement.sh -run TestMSC4311FullEventsOnStrippedStateFederation

COMPLEMENT_DIR=../complement ./scripts-dev/complement.sh -run TestMSC4311RejectInvalidStrippedStateFederation
```

```
go test ./match -v -count=1 -run TestJSONArraySome
```

PR #871, opened 2026-05-04
4311 stripped state create event

FrenchGithubUser:
I think that there was a small error in `TestMSC4311FullCreateEventOnStrippedState`, the test was checking for extra fields in a stripped event while they shouldn't be there

Comes with https://github.com/element-hq/synapse/pull/19749

### Pull Request Checklist

### DIFF
--- a/tests/v12_test.go
+++ b/tests/v12_test.go
@@ -1357,15 +1357,12 @@ func TestMSC4311FullCreateEventOnStrippedState(t *testing.T) {
 			fmt.Sprintf("rooms.invite.%s.invite_state.events", client.GjsonEscape(roomID)),
 		)
 		must.NotEqual(t, len(inviteState.Array()), 0, "no events in invite_state")
-		// find the create event
+		// find the create event: MSC4311 requires it to be present, but clients
+		// still receive it in stripped state format (no extra fields)
 		found := false
 		for _, ev := range inviteState.Array() {
 			if ev.Get("type").Str == spec.MRoomCreate {
 				found = true
-				// we should have extra fields
-				must.MatchGJSON(t, ev,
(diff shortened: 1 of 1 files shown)


tests/v12_test.go
```diff
@@ -1357,15 +1357,12 @@ func TestMSC4311FullCreateEventOnStrippedState(t *testing.T) {
 			fmt.Sprintf("rooms.invite.%s.invite_state.events", client.GjsonEscape(roomID)),
 		)
 		must.NotEqual(t, len(inviteState.Array()), 0, "no events in invite_state")
-		// find the create event
+		// find the create event: MSC4311 requires it to be present, but clients
+		// still receive it in stripped state format (no extra fields)
```
MadLittleMods, line 1361: > I think that there was a small error in `TestMSC4311FullCreateEventOnStrippedState`,

Yes, this test is [flawed](https://github.com/matrix-org/complement/pull/791#discussion_r3132468346) but there are two distinct pieces to MSC4311:

 1. Ensuring the `m.room.create` event is included in the client API (stripped state)
 1. Ensuring all state events in the federation API are using full PDU including the `m.room.create` which is required.

https://github.com/matrix-org/complement/pull/796 fixes the second issue.

The first issue doesn't have a good test yet and I don't think this test is necessarily a "good" test. It looks like you've adapted what's here 👍  but it could be better. For example, even the test name here needs updating. And we could some of the same `must.MatchGJSON(...)` assertions here instead of these for loops. Also can be simplified to not use federation at all, etc, etc, etc.

MadLittleMods: Overall, this is probably something we can tackle in https://github.com/matrix-org/complement/pull/796

PR #873, opened 2026-05-07
test: make sure that `allowed_room_ids` is included in the `/summary` client-server API response for rooms with restricted join rules, as required by Matrix 1.15.

FrenchGithubUser:
Test for https://github.com/element-hq/synapse/pull/19762

### Pull Request Checklist

### DIFF
--- a/tests/room_summary_test.go
+++ b/tests/room_summary_test.go
@@ -1,0 +1,81 @@
+// Tests the GET /_matrix/client/v1/room_summary/{roomIdOrAlias} endpoint
+// as specified in https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv1room_summaryroomidoralias
+
+package tests
+
+import (
+	"testing"
+
+	"github.com/matrix-org/complement"
+	"github.com/matrix-org/complement/helpers"
+	"github.com/matrix-org/complement/match"
+	"github.com/matrix-org/complement/must"
+)
+
+// TestRoomSummaryAllowedRoomIDs checks that the /summary endpoint includes
+// allowed_room_ids for rooms with restricted join rules, and omits the field
+// for rooms that do not use restricted join rules.
+func TestRoomSummaryAllowedRoomIDs(t *testing.T) {
+	deployment := complement.Deploy(t, 1)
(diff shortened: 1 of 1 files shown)


reivilibre (approved):
appears to match https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv1room_summaryroomidoralias

tests/room_summary_test.go
```diff
@@ -0,0 +1,81 @@
...
+		res := alice.MustDo(t, "GET", []string{"_matrix", "client", "v1", "room_summary", restrictedRoom})
+		must.MatchResponse(t, res, match.HTTPResponse{
+			StatusCode: 200,
+			JSON: []match.JSON{
+				match.JSONKeyEqual("room_id", restrictedRoom),
+				match.JSONKeyEqual("join_rule", "restricted"),
+				match.JSONKeyEqual("allowed_room_ids", []interface{}{space}),
+			},
+		})
+	})
```
MadLittleMods, line 69: Any tests for `/room_summary` can be flaky for workerized Synapse deployments because the `/room_summary` endpoint [uses information from the `room_stats_state` table](https://github.com/element-hq/synapse/blob/5df6d1be652b46fc6201c9eb2541bf1828f83274/synapse/handlers/room_summary.py#L782-L804) which is updated out-of-band as replication reaches the worker that handles background tasks. This is a **read-after-write consistency issue in Synapse**.

I created https://github.com/element-hq/synapse/issues/19905 to track the flakiness for this specific test as we're seeing flaky failures where the `join_rule` is missing from the `/room_summary` response (`room_stats_state` hasn't been updated yet).

I'm commenting as these tests were specifically added to test a change to Synapse (https://github.com/element-hq/synapse/pull/19762). **I think this is more a problem with Synapse, not the test though.** If we wanted to be extra nice on the test side, we could `alice.MustSyncUntil(...)` and wait until the room shows up with the expected state. As a note: this isn't bullet-proof either as the worker serving `/sync` could be different than the worker updating `room_stats_state` and handling `/room_summary`. I think it would probably solve the immediate pesky flakes though.

PR #878, opened 2026-06-08
Fix race in `WithWaitForLeave` that flaked `TestPartialStateJoin/Outgoing_device_list_updates`

erikjohnston:
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### Background

  `TestPartialStateJoin/Outgoing_device_list_updates` (in `tests/msc3902`) flakes in downstream CI (seen in Synapse), failing during test cleanup with:

  ```
  federation_room_join_partial_state_test.go:77: Received unexpected PDU: {…"membership":"leave"…}
  ```

  The "unexpected" PDU is the joining user's own `leave`, sent legitimately as the test tears down — so this is a race in the test helper, not a homeserver bug.

  ### Root cause

  `WithWaitForLeave` registers a PDU handler whose callback pushes matching leaves onto `leaveChannel`, then decides whether to block by peeking at `room.CurrentState`:

  ```go
  removePDUHandler := s.AddPDUHandler(func(e) bool { /* push leaveChannel; return true */ })
  defer removePDUHandler()

  leaveAction()

  if room.CurrentState("m.room.member", userID) == "leave" {  // "already seen" shortcut
      // return without consuming the channel
  } else {
      <-leaveChannel // …or 1s timeout
  }
  ```

  When a transaction arrives, `HandleTransactionRequests` calls `room.AddEvent` (which updates `CurrentState`) **before** invoking the PDU callback. So this interleaving is possible:

  1. fed goroutine: `room.AddEvent(leave)` → `CurrentState(user) == "leave"`, callback not yet run
  2. test goroutine: shortcut sees `"leave"` → returns → `defer removePDUHandler()` deletes the handler
  3. fed goroutine: PDU callback runs → no handler matches → `t.Errorf("Received unexpected PDU")`

  The helper used two different signals for the same event — the callback channel *and* `CurrentState` — and tore its handler down based on the one that fires first. It's flaky because it only fails when the scheduler switches between
  `AddEvent` and the callback (more likely under CI load).

  ### Fix

> [...] to outline how this PR fixes the problem [...]
> 
> We first check whether `userInRoom` according to their own server. It they aren't participating in the room, their server shouldn't send us any more PDU's over federation which means we won't run into the `t.Errorf("Received unexpected PDU")` problem. Note: This logic is slightly flawed as some server implementation could have some queued events that it will still send.
> 
> We also check whether our engineered server is in the room as we might never have joined in the first place if the test failed early.
> 
[...]

PR #880, opened 2026-06-12
Fix cross-test pollution from engineered homeservers re-using the same server names (fix `TestPartialStateJoin` flakes)

MadLittleMods:
### What problem are we trying to solve?

This PR is tackling this problem:

> Many of the faster joins test flakes are due to the homeserver under test failing to contact Complement homeservers after they have been torn down. When this happens, subsequent tests can fail if they use a Complement homeserver that happens to have the same `hostname:port` as one which the homeserver under test has previously marked as offline.
>
> *-- https://github.com/matrix-org/complement/pull/626*

This problem occurs because as an [optimization](https://github.com/matrix-org/complement/pull/471) we try to share the `deployment` across many tests. To be clear, as a dumb-simple solution, if we created a new `deployment` for each test, we wouldn't run into this issue (but then the tests would be "slow" again).

### What does this PR do?

Spawning from https://github.com/matrix-org/complement/pull/878#discussion_r3399412841 where I originally thought we could prevent stray requests to a server just from the federation request [authentication](https://spec.matrix.org/v1.18/server-server-api/#authentication) but turns out, it's already doing the right thing. The only thing we can do is compare the `destination` to our server name which [`VerifyHTTPRequest`](https://github.com/matrix-org/gomatrixserverlib/blob/c9c468727353c4835523349dc7c202e5de9a2e37/fclient/request.go#L216-L227) is already doing. And since the crux of this whole thing is that two Complement engineered servers can use the same recycled port and have identical server names (`host.docker.internal:port`), it doesn't help us here. The rest of the public key stuff that is part of federation requests is just for the origin server to sign itself with, for authenticity.

> `destination`: [Added in v1.3] the server name of the receiving server. This is the same as the `destination` field from the JSON described in step 1. For compatibility with older servers, recipients should accept requests without this parameter, but MUST always send it. If this property is included, but the value does not match the receiving server’s name, the receiving server must deny the request with an HTTP status code 401 Unauthorized.
>
> *-- https://spec.matrix.org/v1.18/server-server-api/#authentication*

Instead, this PR takes an alternative route where we instead never re-use the same `server_name` for a Complement engineered homeserver. This means that even if the real homeserver reaches out to the torn-down engineered homeservers, it just hits 'connection refused' instead of a live, unrelated server. No cross-test pollution.

In the context of `tests/msc3902/federation_room_join_partial_state_test.go`, it means the engineered homeserver will never see a [`t.Errorf("Received unexpected PDU", ...)`](https://github.com/matrix-org/complement/blob/bc2b638b11f5b0a068e39ae51f7ac2782070d14f/tests/msc3902/federation_room_join_partial_state_test.go#L77C7-L77C40) problem.

### How do we currently deal with this?

Currently, we try to workaround the stray request problem by making sure the engineered homeserver leaves any rooms it shares with the real homeserver. This means the real homeserver has no logical reason to contact the other server but it's still technically possible that it might reach out because of some queued federation traffic, etc. I find this approach/pattern pretty bad (especially in the way it's currently implemented). See the "Fix" explanation on https://github.com/matrix-org/complement/pull/878 for more info.

See all of the logic around [`partialStateJoinResult.Destroy`](https://github.com/matrix-org/complement/blob/bc2b638b11f5b0a068e39ae51f7ac2782070d14f/tests/msc3902/federation_room_join_partial_state_test.go#L4399-L4428) and [`WithWaitForLeave`](https://github.com/matrix-org/complement/blob/bc2b638b11f5b0a068e39ae51f7ac2782070d14f/tests/msc3902/federation_room_join_partial_state_test.go#L126-L162) to see how messy this gets and all of the logic that we get to remove because of this new approach.

### Dev notes

`:0` behavior with https://pkg.go.dev/net#Listen

### Todo

 - [x] Remove cleanup from `tests/msc3902/federation_room_join_partial_state_test.go` which is no longer necessary
    - To avoid muddying the purpose and focus of this PR I'll tackle this in a follow-up.
    - -> https://github.com/matrix-org/complement/pull/894
    - -> https://github.com/matrix-org/complement/pull/895
    

### Pull Request Checklist


federation/server.go
```diff
@@ -528,18 +528,96 @@ func (s *Server) Mux() *mux.Router {
...
+	// port (by using `:0`) as it's more efficient. The OS may recycle and re-use freed
+	// ports meaning we could regress to O(n^2) behavior trying to search for each new
+	// port we want to find.
+	//
+	// Using `:0` means an unused port is automatically picked for us (could be random,
+	// could be the next sequential unused port, we don't know). Ideally, we could ask for
+	// the next unused port after X to avoid a bunch of work. When using using `:0`, the
+	// pathological case that is O(n^2) is if OS hands back next lowest unused port
+	// sequentially which would mean we would have to probe and hold each listener until
+	// we finally got something new.
```
MadLittleMods, line 575: If you want to compare to the `:0` solution, see the implementation in https://github.com/matrix-org/complement/pull/880/changes/820b4286e37a0fa26033ae8c2f7975d5a398f5f4

I think this is the better way to go.

federation/server.go
MadLittleMods, line 624: As a result of this change, we can also remove all of the specialty cleanup code from `tests/msc3902/federation_room_join_partial_state_test.go`. See all of the logic around [`partialStateJoinResult.Destroy`](https://github.com/matrix-org/complement/blob/bc2b638b11f5b0a068e39ae51f7ac2782070d14f/tests/msc3902/federation_room_join_partial_state_test.go#L4399-L4428) and [`WithWaitForLeave`](https://github.com/matrix-org/complement/blob/bc2b638b11f5b0a068e39ae51f7ac2782070d14f/tests/msc3902/federation_room_join_partial_state_test.go#L126-L162) to see how messy this gets and all of the logic that we get to remove because of this new approach.

To avoid muddying the purpose and focus of this PR I'll tackle this in a follow-up.

erikjohnston (approved):

MadLittleMods:
Thanks for the review @erikjohnston 🐿️

federation/server.go
MadLittleMods: All of the clean up is happening in https://github.com/matrix-org/complement/pull/894 and https://github.com/matrix-org/complement/pull/895 :hole:

PR #881, opened 2026-06-15
Make `TestThreadReceiptsInSyncMSC4102` deterministic

erikjohnston:
Companion to https://github.com/element-hq/synapse/pull/19838

The test only caught the MSC4102 "prefer unthreaded receipt" bug when the clashing unthreaded and threaded receipts happened to be served in the same /sync response (where read-time dedup hides the threaded one), so it flaked.

Force the receipts into separate sync windows by waiting for each user to observe the unthreaded receipt before sending the clashing threaded one, then assert the threaded receipt never wins, both locally and over federation.

### DIFF
--- a/tests/csapi/thread_notifications_test.go
+++ b/tests/csapi/thread_notifications_test.go
@@ -320,8 +320,25 @@ func TestThreadedReceipts(t *testing.T) {
 	)
 }
 
-// Regression test for https://github.com/matrix-org/matrix-spec/issues/1727
-// Servers should always prefer the unthreaded receipt when there is a clash of receipts
+// Regression test for https://github.com/matrix-org/matrix-spec/issues/1727 (and
+// https://github.com/element-hq/synapse/pull/19838).
+//
+// Servers should always prefer the unthreaded receipt when there is a clash of
+// receipts for the same event, and that preference must be *durable*: it must
+// hold even when the unthreaded and threaded receipts are served in separate
+// `/sync` responses. This happens when the two land at different stream
+// positions, e.g. when they arrive over federation as separate EDUs. A
+// non-durable server that only dedupes at read-time will serve the threaded
+// receipt on its own in a later `/sync`, letting it incorrectly win.
+//
+// To test this deterministically (rather than relying on the receipts happening
+// to be served in the same `/sync` response), each user is made to observe the
+// unthreaded receipt down `/sync` *before* the clashing threaded receipt is
+// sent, so the two are guaranteed to fall in different sync windows. A marker
+// event sent after the threaded receipt then gives a deterministic point by
+// which the threaded receipt must have been delivered if it was going to be. We
+// assert it never is, both for the local user and for a remote user over
+// federation.
 func TestThreadReceiptsInSyncMSC4102(t *testing.T) {
 	runtime.SkipIf(t, runtime.Dendrite) // not supported
 	deployment := complement.Deploy(t, 2)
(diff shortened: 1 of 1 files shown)


tests/csapi/thread_notifications_test.go
```diff
@@ -320,8 +320,25 @@ func TestThreadedReceipts(t *testing.T) {
...
+// Regression test for https://github.com/matrix-org/matrix-spec/issues/1727 (and
+// https://github.com/element-hq/synapse/pull/19838).
+//
+// Servers should always prefer the unthreaded receipt when there is a clash of
+// receipts for the same event, and that preference must be *durable*: it must
+// hold even when the unthreaded and threaded receipts are served in separate
+// `/sync` responses. This happens when the two land at different stream
+// positions, e.g. when they arrive over federation as separate EDUs. A
+// non-durable server that only dedupes at read-time will serve the threaded
+// receipt on its own in a later `/sync`, letting it incorrectly win.
```
MadLittleMods, line 332: It's unclear to me that this is true for all cases.

As far as I can tell, for example, we could also tackle this problem at read-time. In Synapse, `get_linearized_receipts_for_room(...)`/`get_linearized_receipts_for_rooms(...)`, could de-duplicate if a previous unthreaded receipt existed. Even in Synapse, we store things as `(room_id, receipt_type, user_id, thread_id)` after all which accommodates holding onto both

Which also makes me question whether our approach in https://github.com/element-hq/synapse/pull/19838 is the best (not sure)

(also plays into some of the comments which make some assumptions as well)

tests/csapi/thread_notifications_test.go
```diff
@@ -354,23 +371,78 @@ func TestThreadReceiptsInSyncMSC4102(t *testing.T) {
...
+// must have been sent *after* the clashing threaded receipt, so that observing
+// it guarantees the receipt has already been delivered if it was going to be.
+func assertUnthreadedReceiptWins(t *testing.T, user *client.CSAPI, roomID, since, receiptUserID, clashEventID, clashThreadID, markerEventID string) {
+	t.Helper()
+	sawThreadedClash := false
+	// We can't use a `syncHasThreadedReadReceipt` check for this: MustSyncUntil
+	// drops a check once it passes, whereas we need to keep scanning every
+	// response for the forbidden receipt right up until the marker arrives. So
+	// we accumulate a flag by hand and use the marker as the (positive) stop
+	// condition.
```
MadLittleMods, line 425: We can still use `syncHasThreadedReadReceipt` but composed into our own check

https://github.com/matrix-org/complement/blob/bc2b638b11f5b0a068e39ae51f7ac2782070d14f/client/sync.go#L87-L89

tests/csapi/thread_notifications_test.go
```diff
@@ -354,23 +371,78 @@ func TestThreadReceiptsInSyncMSC4102(t *testing.T) {
...
+	marker := client.SyncTimelineHasEventID(roomID, markerEventID)
```
MadLittleMods, line 426:
```suggestion
	syncTimelineHasMarkerEventID := client.SyncTimelineHasEventID(roomID, markerEventID)
```

Although, I don't think it really matters to store and re-use this.

tests/csapi/thread_notifications_test.go
```diff
@@ -320,8 +320,25 @@ func TestThreadedReceipts(t *testing.T) {
...
+//
+// To test this deterministically (rather than relying on the receipts happening
+// to be served in the same `/sync` response), each user is made to observe the
+// unthreaded receipt down `/sync` *before* the clashing threaded receipt is
+// sent, so the two are guaranteed to fall in different sync windows. A marker
+// event sent after the threaded receipt then gives a deterministic point by
+// which the threaded receipt must have been delivered if it was going to be. We
+// assert it never is, both for the local user and for a remote user over
+// federation.
 func TestThreadReceiptsInSyncMSC4102(t *testing.T) {
```
MadLittleMods, line 342: [MSC4102](https://github.com/matrix-org/matrix-spec-proposals/pull/4102) only mentions that the de-duplication should happen when assembling EDU's and there is a unthreaded and threaded receipt in the same response.

It doesn't say anything about an unthreaded read receipt winning out over new threaded read receipts

tests/csapi/thread_notifications_test.go
```diff
@@ -354,23 +371,78 @@ func TestThreadReceiptsInSyncMSC4102(t *testing.T) {
 			},
 		},
 	})
-	// now send an unthreaded RR for event B and a threaded RR for event B and ensure we see the unthreaded RR
-	// down /sync. Non-compliant servers will typically send the last one only.
```
MadLittleMods, line 358: I think this test needs to stay trying to send an unthreaded and threaded read receipt in the same `/send` federation transaction.

But we can't control how a homeserver sends EDU's so I don't think this test can be made deterministic. We need to update it to accommodate the MSC4102 behavior or them being sent across two transactions.

tests/csapi/thread_notifications_test.go
erikjohnston: Is your point that actually you *could* correctly implement this at read time by checking for other receipts when pulling from the DB? Yes I suppose you could?

tests/csapi/thread_notifications_test.go
erikjohnston: > I think this test needs to stay trying to send an unthreaded and threaded read receipt in the same `/send` federation transaction.

Why?



> But we can't control how a homeserver sends EDU's so I don't think this test can be made deterministic. We need to update it to accommodate the [MSC4102](https://github.com/matrix-org/matrix-doc/pull/4102) behavior or them being sent across two transactions.

Sorry, what are you actually suggesting we do?

MadLittleMods: This is the whole point of the [MSC4102](https://github.com/matrix-org/matrix-doc/pull/4102) and we want the test to stress this behavior,

> When a server is combining receipts into an EDU, if there are multiple receipts for the same (user, event, receipt type), always choose the receipt which is unthreaded (has no `thread_id`) when aggregating into an EDU.
> 
> This change will apply to all `m.receipt` EDUs, which includes both CSAPI and Federation endpoints.
>
> *-- [MSC4102](https://github.com/matrix-org/matrix-doc/pull/4102)*

But we can't control how a homeserver emits EDU's so the best we can do is send them quickly in succession and see what it does. Combining into one EDU is fine (per MSC4102) or separating is fine (depends on how slow the server is to process things and when the request comes in).

We need to update the test to accommodate both expectations.

tests/csapi/thread_notifications_test.go
MadLittleMods: Yes, in which case this comment assumes too much

PR #892, opened 2026-07-13
Pin GitHub Actions and narrow permissions

anoadragon453:
Harden our GitHub Actions via the [zizmor](https://zizmor.sh/) GitHub Actions security static analysis tool.

`zizmor --fix=all .github` now passes.

This PR also adds `zizmor` as an action to this repository, so we don't regress the current config.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,16 +6,24 @@ on:
   pull_request:
   workflow_dispatch:
 
+permissions: {}
+
 jobs:
   check-signoff:
     if: "github.event_name == 'pull_request'"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
MadLittleMods: Non-pinned install

MadLittleMods (approved):

.github/workflows/ci.yaml
anoadragon453: Good catch. I'll pin it to latest ([v2.5.0](https://github.com/GoTestTools/gotestfmt/releases/tag/v2.5.0)).

PR #894, opened 2026-07-13
Remove messy `WithWaitForLeave` test clean-up in `TestPartialStateJoin`

MadLittleMods:
Remove messy `WithWaitForLeave` test clean-up in `TestPartialStateJoin`

Follow-up to https://github.com/matrix-org/complement/pull/880 which introduces a better way to avoid the test pollution and obsoletes the need for all of this cleanup logic. See the PR description there for a more complete context on why.

`WithWaitForLeave` was originally introduced in this context:

> Many of the faster joins test flakes are due to the homeserver under test failing to contact Complement homeservers after they have been torn down. When this happens, subsequent tests can fail if they use a Complement homeserver that happens to have the same `hostname:port` as one which the homeserver under test has previously marked as offline.
>
> *-- https://github.com/matrix-org/complement/pull/626*

But thanks to https://github.com/matrix-org/complement/pull/880 we no longer use the same `hostname:port` for engineered homeservers anymore.

And we can go even further with cleaning this up by removing the need for `Destroy(...)` altogether, see https://github.com/matrix-org/complement/pull/895

### Pull Request Checklist

### DIFF
--- a/tests/msc3902/federation_room_join_partial_state_test.go
+++ b/tests/msc3902/federation_room_join_partial_state_test.go
@@ -123,6 +123,74 @@ func (s *server) AddEDUHandler(eduHandler func(gomatrixserverlib.EDU) bool) func
 	}
 }
 
+// WithWaitForLeave runs the given action and, when the resulting leave is
+// expected to reach this server, waits for it. `leaveAction` is always run; the
+// wait is skipped when `user` had already left the room (per their own
+// homeserver, so the action produces no new leave) or when this server isn't in
+// the room (so the leave won't be federated to us).
+func (s *server) WithWaitForLeave(
+	t *testing.T, room *federation.ServerRoom, user *client.CSAPI, leaveAction func(),
+) {
+	userID := user.UserID
+	leaveChannel := make(chan gomatrixserverlib.PDU, 10)
+	removePDUHandler := s.AddPDUHandler(
+		func(e gomatrixserverlib.PDU) bool {
+			if membership, _ := e.Membership(); e.Type() == "m.room.member" &&
+				*e.StateKey() == userID &&
+				membership == "leave" {
+				leaveChannel <- e
+				return true
+			}
+			return false
+		},
+	)
+	defer removePDUHandler()
+
+	// We need to check if the user (on their homeserver) thinks they're in the
+	// room, before performing the `leaveAction` (to avoid races).
+	//
+	// If they are not in the room, then the `leaveAction` will not produce a
+	// new leave event and we should not wait for one.
+	//
+	// If they are in the room then the `leaveAction` will produce a new leave
+	// event. We then need to check if we expect this server receive the leave
+	// event by checking if this server is in the room. If they are, we wait, if
+	// not we can return immediately after the `leaveAction`.
+	userInRoom := userIsJoinedTo(t, user, room.RoomID)
+
+	leaveAction()
+
+	if !userInRoom {
+		// The user had already left, so the action produced no new leave and
+		// none is coming: don't wait.
+		t.Logf("%s is not joined to test room %s; not waiting for them to leave.", userID, room.RoomID)
+		return
+	}
+
+	if !s.isInRoom(room) {
+		// The homeserver only federates the leave to servers that are in the
+		// room. If we aren't, no leave PDU is coming to us, so don't block until
+		// the timeout.
+		t.Logf("%s is not in test room %s; not waiting for %s to leave.", s.ServerName(), room.RoomID, userID)
+		return
+	}
+
+	// Otherwise the action triggered the leave, which arrives as a PDU our
+	// handler matches. Wait on its channel rather than polling
+	// `room.CurrentState`: the room's current state is updated (by
+	// `room.AddEvent`) *before* the PDU callback runs, so returning on a
+	// `CurrentState` check could deregister our handler in the window before the
+	// callback fires, making the (expected) leave look unexpected to
+	// `HandleTransactionRequests`. This returns as soon as the leave arrives; the
+	// timeout is only a ceiling for declaring failure.
+	select {
+	case <-leaveChannel:
+		t.Logf("%s saw %s leave test room %s.", s.ServerName(), userID, room.RoomID)
+	case <-time.After(1 * time.Second):
+		t.Errorf("%s timed out waiting for %s to leave test room %s.", s.ServerName(), userID, room.RoomID)
+	}
+}
+
 // isInRoom reports whether this Complement server has a joined user in the room,
 // according to its own `ServerRoom` view. The server reliably tracks its own
 // users' membership (it created their join/leave events), so this answers "will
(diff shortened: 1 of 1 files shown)


tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -4465,23 +4381,6 @@ func (psj *partialStateJoinResult) Destroy(t *testing.T) {
...
 		psj.fedStateIdsRequestReceivedWaiter.Finish()
 	}
-
-	// Since the same deployment is being used across multiple tests, ensure that it
-	// has finished all federation activity before tearing down the Complement server.
-	// Otherwise the homeserver at the Complement's hostname:port combination may be
-	// considered offline and interfere with subsequent tests.
-	t.Log("Cleaning up after test...")
-
-	awaitPartialStateJoinCompletion(t, psj.ServerRoom, psj.User)
```
MadLittleMods, line 4475: LLM review likes to call out that `awaitPartialStateJoinCompletion(...)` is necessary in order for the server handlers in `handleStateIdsRequests(...)`/`handleStateRequests(...)` to finish before the test does but I don't think it really gives any guarantee of anything.

What it's trying to say is that calling `psj.fedStateIdsSendResponseWaiter.Finish()` wakes up the waiter so the handler [goroutine can keep running](https://github.com/matrix-org/complement/blob/f002aff99e240c994231a21985546b1c16d4173c/tests/msc3902/federation_room_join_partial_state_test.go#L4539-L4541) but we don't wait for that handler to finish here in `Destroy(...)`. The LLM claims that `awaitPartialStateJoinCompletion(...)` (requests `/members` on the homeserver) makes sure that the `/state_ids` or `/state` goroutine handlers finish because those requests are necessary to resync the room (partial -> full state) which is necessary in order to serve `/members`.

That seems like a very tenuous connection. I see the race condition it's explaining but using this is the wrong mechanism.

And the only thing it's trying to protect from is `t` being used after the test finishes which would panic.

In any case, all of this isn't necessary with the second step of this cleanup in a follow-up PR: https://github.com/matrix-org/complement/pull/895

And the tests still pass in the mean-time (theoretical flake)

tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -2529,13 +2457,7 @@ func TestPartialStateJoin(t *testing.T) {
 			server1.MustSendTransaction(t, deployment, deployment.GetFullyQualifiedHomeserverName(t, "hs1"), []json.RawMessage{leaveEvent.JSON()}, nil)
 			syncToken = awaitEventViaSync(t, alice, partialStateRoom.RoomID, leaveEvent.EventID(), syncToken)
 
-			leaveSharedRoom = func() {
-				server2.WithWaitForLeave(t, server2Room, alice, func() {
-					alice.MustLeaveRoom(t, roomID)
-				})
-			}
```
MadLittleMods, line 2536: This is in the function called `leaveSharedRoom` but the leave already happens above.

For context, this was added in https://github.com/matrix-org/complement/pull/627 which was addressing the flakes caused by homeservers still participating in rooms after each test.

Which as the PR description explains, we have a better solution for now.

tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -4455,8 +4373,6 @@ func beginPartialStateJoin(t *testing.T, server *server, serverRoom *federation.
 
 // Destroy cleans up the resources associated with the join attempt.
 // It is idempotent and must be called once the test is finished.
-// Specifically, it ensures that the partial state join completes and makes the joining user leave
-// the room.
 func (psj *partialStateJoinResult) Destroy(t *testing.T) {
```
MadLittleMods, line 4376: As a follow-up, we're removing this entire need for this function in https://github.com/matrix-org/complement/pull/895

anoadragon453 (commented):
I'm a little confused by the commit history. Did you mean to revert the first commit via the second @MadLittleMods?

MadLittleMods:
@anoadragon453 Yes. I tried removing everything, `WithWaitForLeave` and `Destroy` altogether, but it required a few more changes which I made in a follow-up PR instead, see details in https://github.com/matrix-org/complement/pull/895#discussion_r3574649551

This is a first step which just removes `WithWaitForLeave`

anoadragon453 (approved):

MadLittleMods:
Thanks for the review @anoadragon453 🐝

tests/msc3902/federation_room_join_partial_state_test.go
MadLittleMods: @realtyem shared a CI run where he encountered this exact goroutine panic. So not theoretical anymore but proof that this code didn't protect from anything (already flawed) :police_officer: 

:x: https://github.com/clokep/synapse/actions/runs/29368036552/job/87205014543
```
goroutine 6397 [running]:
net/http.(*conn).serve.func1()
	/opt/hostedtoolcache/go/1.25.0/x64/src/net/http/server.go:1943 +0xd3
panic({0x906240?, 0xc000b05110?})
	/opt/hostedtoolcache/go/1.25.0/x64/src/runtime/panic.go:783 +0x132
testing.(*common).log(0xc001618380, {0xc001387440?, 0x57?})
	/opt/hostedtoolcache/go/1.25.0/x64/src/testing/testing.go:1030 +0x1ac
testing.(*common).Logf(0xc001618380, {0x9ee810?, 0x1?}, {0xc001451938?, 0x9e4ad0?, 0x1e?})
	/opt/hostedtoolcache/go/1.25.0/x64/src/testing/testing.go:1191 +0x4f
github.com/matrix-org/complement/tests/msc3902.handleStateIdsRequests.func1({0xa9e460, 0xc001210b40}, 0xc000b04e70?)
	/home/runner/work/synapse/synapse/complement/tests/msc3902/federation_room_join_partial_state_test.go:4542 +0x205
net/http.HandlerFunc.ServeHTTP(0x0?, {0xa9e460?, 0xc001210b40?}, 0xc?)
	/opt/hostedtoolcache/go/1.25.0/x64/src/net/http/server.go:2322 +0x29
github.com/matrix-org/complement/federation.NewServer.func2.1({0xa9e460, 0xc001210b40}, 0xc00086f2c0)
	/home/runner/work/synapse/synapse/complement/federation/server.go:114 +0x122
net/http.HandlerFunc.ServeHTTP(0xc00086f180?, {0xa9e460?, 0xc001210b40?}, 0x4b9c29?)
	/opt/hostedtoolcache/go/1.25.0/x64/src/net/http/server.go:2322 +0x29
github.com/gorilla/mux.(*Router).ServeHTTP(0xc000bbe000, {0xa9e460, 0xc001210b40}, 0xc00086f040)
	/home/runner/go/pkg/mod/github.com/gorilla/mux@v1.8.1/mux.go:212 +0x1e2
net/http.serverHandler.ServeHTTP({0xc000d35740?}, {0xa9e460?, 0xc001210b40?}, 0x1?)
	/opt/hostedtoolcache/go/1.25.0/x64/src/net/http/server.go:3340 +0x8e
net/http.(*conn).serve(0xc001538f30, {0xa9f3c0, 0xc000be17d0})
	/opt/hostedtoolcache/go/1.25.0/x64/src/net/http/server.go:2109 +0x665
created by net/http.(*Server).Serve in goroutine 6376
	/opt/hostedtoolcache/go/1.25.0/x64/src/net/http/server.go:3493 +0x485
```

In the stack trace `/home/runner/work/synapse/synapse/complement/tests/msc3902/federation_room_join_partial_state_test.go:4542` points to https://github.com/matrix-org/complement/blob/edfe298b6a9be6511c60920cf75397bb6731beb4/tests/msc3902/federation_room_join_partial_state_test.go#L4542

which is exactly `t` usage after the test ends (causing a panic)

PR #895, opened 2026-07-13
Remove messy `Destroy` test clean-up in `TestPartialStateJoin`

MadLittleMods:
Remove messy `Destroy` test clean-up in `TestPartialStateJoin`

Follow-up to https://github.com/matrix-org/complement/pull/894 which removes another part of the clean-up code in this file and https://github.com/matrix-org/complement/pull/880 which introduces a better way to avoid the test pollution and obsoletes the need for all of this cleanup logic. See the PR description there for a more complete context on why.

`Destroy(...)` was originally introduced in this context:

> Since the partial state join tests reuse the same homeserver deployment, it is important that each test leaves the homeserver in an acceptable state for subsequent tests. In particular, we must allow in-progress partial-state joins to complete before tearing down Complement servers, otherwise the homeserver may mark Complement `hostname:port` combinations as offline and refuse to contact them in subsequent tests.
>
> *-- https://github.com/matrix-org/complement/pull/570*

But thanks to https://github.com/matrix-org/complement/pull/880 we no longer use the same `hostname:port` for engineered homeservers anymore.

---

Given the nature of these changes, this also fixes the goroutine panic that can happen when the waiter finishes after the test and `t` is used after the test finishes, see https://github.com/matrix-org/complement/pull/894#discussion_r3574583828

### Todo

 - [x] Wait for https://github.com/matrix-org/complement/pull/894 to merge

### Pull Request Checklist


tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -370,7 +302,6 @@ func TestPartialStateJoin(t *testing.T) {
 		defer cancel()
 		serverRoom := createTestRoom(t, server, alice.GetDefaultRoomVersion(t))
 		psjResult := beginPartialStateJoin(t, server, serverRoom, alice)
-		defer psjResult.Destroy(t)
```
MadLittleMods, line 305: Removing all of the `Destroy` usage since it's no longer necessary

tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -4537,7 +4360,26 @@ func handleStateIdsRequests(
 				requestReceivedWaiter.Finish()
 			}
 			if sendResponseWaiter != nil {
-				sendResponseWaiter.Waitf(t, 60*time.Second, "Waiting for /state_ids request")
+				select {
+				case <-sendResponseWaiter.Done():
+					// Happy-path now that we're done waiting, continue serving the request now
+				case <-req.Context().Done():
```
MadLittleMods, line 4376: Instead of relying on `Destroy` to finish the waiter to allow this to continue, we instead rely on engineered homeserver being closed and all of the in-flight requests being cancelled (`req.Context().Done()`).

We still have a 60 second timeout like before.

tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -4449,41 +4304,9 @@ func beginPartialStateJoin(t *testing.T, server *server, serverRoom *federation.
...
-	// has finished all federation activity before tearing down the Complement server.
-	// Otherwise the homeserver at the Complement's hostname:port combination may be
-	// considered offline and interfere with subsequent tests.
-	t.Log("Cleaning up after test...")
-
-	awaitPartialStateJoinCompletion(t, psj.ServerRoom, psj.User)
-
-	// The caller is about to tear down the Complement homeserver. Leave the room, so
-	// that the homeserver under test stops sending it presence updates.
-	psj.Server.WithWaitForLeave(
```
MadLittleMods: https://github.com/matrix-org/complement/pull/894 removes the `WithWaitForLeave(...)` part of `Destroy(...)`

This PR removes the need for finishing the waiters.

Which means `Destroy(...)` doesn't do anything anymore and we can remove it altogether. Less to think about.

tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -123,74 +123,6 @@ func (s *server) AddEDUHandler(eduHandler func(gomatrixserverlib.EDU) bool) func
 	}
 }
 
-// WithWaitForLeave runs the given action and, when the resulting leave is
-// expected to reach this server, waits for it. `leaveAction` is always run; the
-// wait is skipped when `user` had already left the room (per their own
-// homeserver, so the action produces no new leave) or when this server isn't in
-// the room (so the leave won't be federated to us).
-func (s *server) WithWaitForLeave(
```
MadLittleMods: The `WithWaitForLeave` changes in the diff disappear once https://github.com/matrix-org/complement/pull/894 is merged

This PR builds on that one so they are also included here for now.

tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -1284,7 +1265,29 @@ func TestPartialStateJoin(t *testing.T) {
...
+					// already completed and touching `t` after that panics.
+					return
+				case <-time.After(60 * time.Second):
+					// Sanity check so a wedged test fails loudly instead of blocking forever.
+					t.Fatalf(
+						"Timed out waiting for the test to finish the `sendResponseWaiter` while trying"+
+							"to serve /state_ids response for event %s", queryParams["event_id"],
+					)
+				}
+
```
MadLittleMods, line 1290: This whole handler is the same as `handleStateIdsRequests(...)` except that it's serving an invalid response.

It's very tempting to dry this out but it's prior art ⏩

tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -3712,8 +3685,9 @@ func TestPartialStateJoin(t *testing.T) {
 			psjResult := beginPartialStateJoin(t, server, serverRoom, alice)
 			// At the end of the test, keep Bob in the room. Have him make a /members
 			// call to ensure the resync has completed.
+			//
+			// FIXME: huh?
 			psjResult.User = bob
```
MadLittleMods: Just calling this out as I don't understand the logic/code

Something to figure out in the future

tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -4322,7 +4287,7 @@ type partialStateJoinResult struct {
 
 // beginPartialStateJoin has a test user attempt to join the given room.
 //
-// It returns a partialStateJoinResult, which must be Destroy'd on completion.
+// # It returns a partialStateJoinResult
```
MadLittleMods: Some weird auto-formatting in my editor keeps adding the `#` 🤔 

TODO: should be removed

MadLittleMods: This line appears to accidentally match godoc's legacy "implicit heading" heuristic :feelsgood: 

> Before Go 1.19, headings were identified implicitly by single-line paragraphs satisfying certain conditions, most notably the lack of any terminating punctuation.
>
> Gofmt reformats [lines treated as implicit headings](https://github.com/golang/proposal/blob/master/design/51082-godocfmt.md#headings) by earlier versions of Go to use # headings instead. If the reformatting is not appropriate—that is, if the line was not meant to be a heading—the easiest way to make it a paragraph is to introduce terminating punctuation such as a period or colon, or to break it into two lines.
>
> *-- https://go.dev/doc/comment#headings*

MadLittleMods: Added some punctuation to avoid this auto-formatting :white_check_mark:

tests/msc3902/federation_room_join_partial_state_test.go
```diff
@@ -4436,7 +4382,26 @@ func handleStateIdsRequests(
 				requestReceivedWaiter.Finish()
 			}
 			if sendResponseWaiter != nil {
-				sendResponseWaiter.Waitf(t, 60*time.Second, "Waiting for /state_ids request")
+				select {
+				case <-sendResponseWaiter.Done():
+					// Happy-path now that we're done waiting, continue serving the request now
+				case <-req.Context().Done():
```
MadLittleMods, line 4376: Instead of relying on `Destroy` to finish the waiter to allow this to continue, we instead rely on engineered homeserver being closed and all of the in-flight requests being cancelled (`req.Context().Done()`).

We still have a 60 second timeout like before.

This also fixes the goroutine panic that can happen when the waiter finishes after the test and `t` is used after the test finishes, see https://github.com/matrix-org/complement/pull/894#discussion_r3574583828

tests/msc3902/federation_room_join_partial_state_test.go
MadLittleMods: Re-commented on the diff -> https://github.com/matrix-org/complement/pull/895#discussion_r3590794645

tests/msc3902/federation_room_join_partial_state_test.go
MadLittleMods: Here is what an LLM said:

> `awaitPartialStateJoinCompletion` has the given user make a /members call, which blocks until the partial-state resync completes — a way to keep the homeserver from still trying to contact the Complement server after teardown. `beginPartialStateJoin` defaults `psjResult.User` to the joining user (alice), but in these two subtests alice leaves the room mid-test, so she couldn't be the one to make that /members call at cleanup time. Bob stays joined, so the tests swapped in bob: `psjResult.User = bob`. That's what the comment "At the end of the test, keep Bob in the room. Have him make a `/members` call..." was referring to.

And that it's "now dead code, orphaned by your own cleanup"

MadLittleMods: Removed the two places that did this special logic :white_check_mark: 

And we are also able to remove `partialStateJoinResult.User` since nothing uses it anymore :wastebasket: 

See aa75e580291b54bcec40adf4b4ffba09692d7fd5

erikjohnston (approved):

MadLittleMods:
Thanks for the review @erikjohnston 🦅

PR #903, opened 2026-07-17
Guard against homeservers stuck in a `/messages` pagination loop

gamesguru:
Fixes a couple 1 hour deadlock scenarios observed in Complement during a federation/timeline refactor of homeserver code.

Enhances checks and emits more helpful failure messages.

Additionally asserts no element is served twice (lessening the chances of future regressions or TOCTOU bugs).

---

### Pull Request Checklist

### DIFF
--- a/tests/csapi/room_messages_test.go
+++ b/tests/csapi/room_messages_test.go
@@ -427,7 +427,18 @@ func _sendAndTestMessageHistory(
 		testCase.numberOfMessagesToSend,
 	)
 	fromToken := ""
-	for {
+	// Guard against a buggy homeserver that never terminates pagination (e.g. it
+	// keeps returning an `end` token, possibly oscillating between a small set of
+	// tokens like `[-1, 1]` -> `[1, -1]` -> ...). Without this, the test would spin
+	// until the overall test timeout (which can be as long as 3600s) instead of
+	// failing fast with a useful error.
+	seenTokens := map[string]bool{fromToken: true}
+	maxIterations := testCase.numberOfMessagesToSend + 1
+	for iteration := 0; ; iteration++ {
+		if iteration >= maxIterations {
+			t.Fatalf("paginated %d times without reaching the start of the room (no `end` token) -- homeserver may be stuck in a pagination loop", iteration)
+		}
+
 		messageQueryParams := url.Values{
 			"dir":   []string{"b"},
 			"limit": []string{strconv.Itoa(testCase.messagesRequestLimit)},
@@ -472,6 +483,11 @@ func _sendAndTestMessageHistory(
 			break
 		}
 		fromToken = endTokenRes.Str
+
+		if seenTokens[fromToken] {
+			t.Fatalf("homeserver returned a pagination token (%s) we've already seen -- it appears to be stuck in loop or repeating elements", fromToken)
+		}
+		seenTokens[fromToken] = true
 	}
 
 	// Put them in chronological order to match the expected list


tests/csapi/room_messages_test.go
```diff
@@ -472,6 +483,11 @@ func _sendAndTestMessageHistory(
 			break
 		}
 		fromToken = endTokenRes.Str
+
+		if seenTokens[fromToken] {
+			t.Fatalf("homeserver returned a pagination token (%s) we've already seen -- it appears to be stuck in loop or repeating elements", fromToken)
```
MadLittleMods, line 488: I think it's valid for a homeserver to return the same pagination token if it doesn't have anything to return yet (still backfilling).

(we should get rid of this check)

tests/csapi/room_messages_test.go
```diff
@@ -427,7 +427,18 @@ func _sendAndTestMessageHistory(
...
+	// keeps returning an `end` token, possibly oscillating between a small set of
+	// tokens like `[-1, 1]` -> `[1, -1]` -> ...). Without this, the test would spin
+	// until the overall test timeout (which can be as long as 3600s) instead of
+	// failing fast with a useful error.
+	seenTokens := map[string]bool{fromToken: true}
+	maxIterations := testCase.numberOfMessagesToSend + 1
+	for iteration := 0; ; iteration++ {
+		if iteration >= maxIterations {
+			t.Fatalf("paginated %d times without reaching the start of the room (no `end` token) -- homeserver may be stuck in a pagination loop", iteration)
+		}
```
MadLittleMods, line 440: Having `maxIterations` can make sense 👍

tests/csapi/room_messages_test.go
```diff
@@ -427,7 +427,18 @@ func _sendAndTestMessageHistory(
 		testCase.numberOfMessagesToSend,
 	)
 	fromToken := ""
-	for {
+	// Guard against a buggy homeserver that never terminates pagination (e.g. it
+	// keeps returning an `end` token, possibly oscillating between a small set of
+	// tokens like `[-1, 1]` -> `[1, -1]` -> ...). Without this, the test would spin
+	// until the overall test timeout (which can be as long as 3600s) instead of
+	// failing fast with a useful error.
```
MadLittleMods, line 434:
```suggestion
	// Guard against a buggy homeserver that never terminates pagination (e.g. it keeps
	// returning an `end` token). Without this, the test could spin until the overall test
	// timeout instead of failing fast with a useful error.
	//
	// We chose `testCase.numberOfMessagesToSend + 1` as it allows the homeserver to
	// return events one by one with an extra request for good measure (no more events).
```

tests/csapi/room_messages_test.go
```diff
@@ -427,7 +427,18 @@ func _sendAndTestMessageHistory(
...
+	seenTokens := map[string]bool{fromToken: true}
+	maxIterations := testCase.numberOfMessagesToSend + 1
+	for iteration := 0; ; iteration++ {
```
MadLittleMods, line 437:
```suggestion
	for paginationAttempt := 0; ; paginationAttempt++ {
```

tests/csapi/room_messages_test.go
```diff
@@ -427,7 +427,18 @@ func _sendAndTestMessageHistory(
...
 	)
 	fromToken := ""
-	for {
+	// Guard against a buggy homeserver that never terminates pagination (e.g. it
+	// keeps returning an `end` token, possibly oscillating between a small set of
+	// tokens like `[-1, 1]` -> `[1, -1]` -> ...). Without this, the test would spin
+	// until the overall test timeout (which can be as long as 3600s) instead of
+	// failing fast with a useful error.
+	seenTokens := map[string]bool{fromToken: true}
+	maxIterations := testCase.numberOfMessagesToSend + 1
```
MadLittleMods, line 436:
```suggestion
	maxPaginationAttempts := testCase.numberOfMessagesToSend + 1
```

tests/csapi/room_messages_test.go
```diff
@@ -427,7 +427,18 @@ func _sendAndTestMessageHistory(
...
+	for iteration := 0; ; iteration++ {
+		if iteration >= maxIterations {
+			t.Fatalf("paginated %d times without reaching the start of the room (no `end` token) -- homeserver may be stuck in a pagination loop", iteration)
```
MadLittleMods, line 439:
```suggestion
		t.Fatalf(
			"paginated %d times without reaching the start of the room (no `end` token) "+
				"(saw %d out of %d expected events) -- homeserver may be stuck in a pagination loop",
			iteration,
			len(filterEventIDs(t, actualEventIDsFromRequest, eventIDs)),
			len(eventIDs),
		)
```

tests/csapi/room_messages_test.go
gamesguru: I agree it's allowed in the spec. However, having run this test against Synapse over 200 times and never seen a failure in this test, I would question what potential value allowing duplicate tokens has? Can you defend that?

Given the overwhelming empirical evidence that no homeserver duplicates tokens under this test, it seems a reasonable enforcement. (That could be totally invalid in other tests, but it strongly appears to hold here.)

My complement run failed for an unrelated flake:

```log
❌ TestRoomSummaryAllowedRoomIDs/restricted_room_includes_allowed_room_ids (0s)
      room_summary_test.go:60: CSAPI.MustDo GET http://127.0.0.1:32784/_matrix/client/v1/room_summary/%21HrEM4o3rDaGUCMUP:hs1 returned non-2xx code: 404 Not Found - body: {"errcode":"M_UNRECOGNIZED","error":"Unrecognized request"}
```

PR #909, opened 2026-08-03
Migrate docker module to moby

AndrewFerr:
Migrate from the deprecated github.com/docker/docker module to the current latest version of github.com/moby/moby, and make all changes necessary to adapt to its refactoring.

Signed-off-by: Andrew Ferrazzutti <andrewf@element.io>

### Pull Request Checklist


internal/docker/builder.go
```diff
@@ -466,11 +467,11 @@ func createNetworkIfNotExists(docker *client.Client, pkgNamespace, blueprintName
 	if err != nil {
 		return "", fmt.Errorf("%s: failed to create docker network. %w", blueprintName, err)
 	}
-	if nw.Warning != "" {
+	if len(nw.Warning) > 0 && nw.Warning[0] != "" {
```
reivilibre, line 470: wondering why we have to check `[0] != ""` as well. Is it actually necessary?
I don't mind and I don't know, just seems bizarre to me

reivilibre (approved):
Thanks!

internal/docker/builder.go
AndrewFerr: I don't know if it's necessary either, but figured it was cheap to guard against it just in case.

cmd/perftest/snapshot.go
```diff
@@ -5,8 +5,9 @@ import (
 	"encoding/json"
 	"time"
 
-	"github.com/docker/docker/api/types/container"
 	"github.com/matrix-org/complement/internal/docker"
```
MadLittleMods, line 8: For reference, what's spawning this change? (migrating docker module to moby)

AndrewFerr: The fact that `docker/docker` is no longer getting new releases & has some CVEs that AIUI are fixed only in the replacement `moby/` packages.

Granted, CVEs aren't much of an issue for integration tests, but this felt like a good catalyst for doing the migration.

MadLittleMods: What made you care about this though?

How did you notice?

AndrewFerr: One of our security scanners (Vanta) flagged the CVEs in all of our repos that use Complement.

PR #917, opened 2026-09-07
Fix missing time boundary guard in `TestJumpToDateEndpoint`

kegsay:
Whilst #872 tried to fix this, ~~the review process reintroduced the bug~~ for one specific scenario: when the initial room creation events are in the same millisecond as the event "before A". Without this extra sleep, `timeBeforeEventA` may actually be the "time exactly at the last initial creation events" which is Bad since it then messes up which events get returned.

### Pull Request Checklist

### DIFF
--- a/tests/room_timestamp_to_event_test.go
+++ b/tests/room_timestamp_to_event_test.go
@@ -362,7 +362,7 @@ func createTestRoom(t *testing.T, c *client.CSAPI) (roomID string, eventA, event
 	roomID = c.MustCreateRoom(t, map[string]interface{}{
 		"preset": "public_chat",
 	})
-
+	time.Sleep(tsBoundaryGuard) // ensure timeBeforeEventA is AFTER the initial creation events
 	timeBeforeEventA := time.Now()
 	time.Sleep(tsBoundaryGuard)
 	eventAID := c.SendEventSynced(t, roomID, b.Event{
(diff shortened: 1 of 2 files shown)


tests/room_timestamp_to_event_test.go
```diff
@@ -362,7 +362,7 @@ func createTestRoom(t *testing.T, c *client.CSAPI) (roomID string, eventA, event
 	roomID = c.MustCreateRoom(t, map[string]interface{}{
 		"preset": "public_chat",
 	})
-
+	time.Sleep(tsBoundaryGuard) // ensure timeBeforeEventA is AFTER the initial creation events
 	timeBeforeEventA := time.Now()
 	time.Sleep(tsBoundaryGuard)
```
MadLittleMods: > Fix regression introduced in #872

This isn't a regression.

> Whilst #872 tried to fix this, the review process reintroduced the bug for one specific scenario

I specifically pointed this out in review https://github.com/matrix-org/complement/pull/872#discussion_r3342425944 but it was never addressed and you approved/merged the PR.

tests/room_timestamp_to_event_test.go
```diff
@@ -362,7 +362,7 @@ func createTestRoom(t *testing.T, c *client.CSAPI) (roomID string, eventA, event
 	roomID = c.MustCreateRoom(t, map[string]interface{}{
 		"preset": "public_chat",
 	})
-
+	time.Sleep(tsBoundaryGuard) // ensure timeBeforeEventA is AFTER the initial creation events
```
MadLittleMods: We can copy the comment for how we explain this for `timeBeforeEventB`

jevolk (approved):

tests/room_timestamp_to_event_test.go
MadLittleMods: This was resolved but the PR title/description was never updated.

I've just made some light touches :shrug:

PR #921, opened 2026-09-15
bugfix: fix a federation connection leak which could cause tests to fail

kegsay:
If you have a low file descriptor (fd) limit (e.g. 256) then some test packages which do lots of federation requests from test -> server could result in the limit being reached due to there being a ton of HTTP transports each with a few idle connections because we did not reuse transports between HS names.

This would result in obscure errors like "Invalid request signature" because the `/key/server` request was failing.

### Pull Request Checklist


internal/docker/deployment.go
```diff
@@ -28,6 +29,25 @@ type Deployment struct {
...
+	if t, ok := d.transports.Load(hsName); ok {
+		return t.(*http.Transport)
+	}
+	t, _ := d.transports.LoadOrStore(hsName, &http.Transport{
+		TLSClientConfig: &tls.Config{
+			ServerName:         hsName,
+			InsecureSkipVerify: true,
+		},
+		MaxIdleConnsPerHost: 10,
+		IdleConnTimeout:     30 * time.Second,
```
MadLittleMods: Comment why

internal/docker/deployment.go
```diff
@@ -28,6 +29,25 @@ type Deployment struct {
...
+func (d *Deployment) transportFor(hsName string) *http.Transport {
+	if t, ok := d.transports.Load(hsName); ok {
+		return t.(*http.Transport)
+	}
+	t, _ := d.transports.LoadOrStore(hsName, &http.Transport{
+		TLSClientConfig: &tls.Config{
+			ServerName:         hsName,
+			InsecureSkipVerify: true,
+		},
+		MaxIdleConnsPerHost: 10,
```
MadLittleMods: Comment why

internal/docker/deployment.go
```diff
@@ -28,6 +29,25 @@ type Deployment struct {
...
+		IdleConnTimeout:     30 * time.Second,
+	})
```
MadLittleMods, line 55: To explain this for myself, in a concurrent situation where `LoadOrStore(...)` finds an existing entry, we're still creating a new `http.Transport` but it probably gets GC'd because nothing references it and it hasn't opened any connections yet.

Additionally, we have `IdleConnTimeout` which means any connections will timeout after 30 seconds and then `http.Transport` should be garbage collected.

internal/docker/deployment.go
```diff
@@ -28,6 +29,25 @@ type Deployment struct {
 	HS               map[string]*HomeserverDeployment
 	Config           *config.Complement
 	localpartCounter atomic.Int64
+	// HTTP transports used by RoundTripper, keyed by homeserver.
+	// Transports must be reused else we will leak some idle connections over the lifetime of the test run.
```
MadLittleMods: We should probably re-frame this to talk more about how we could more easily hit the file descriptor limit and run into obscure errors that don't point to the problem.

`IdleConnTimeout` will clean up idle connections over time

kegsay: https://github.com/matrix-org/complement/pull/921/changes/3c956b75f95f5ba51481635f589c3bd654e8f33f

internal/docker/deployment.go
kegsay: https://github.com/matrix-org/complement/pull/921/changes/3c956b75f95f5ba51481635f589c3bd654e8f33f

internal/docker/deployment.go
kegsay: https://github.com/matrix-org/complement/pull/921/changes/3c956b75f95f5ba51481635f589c3bd654e8f33f

MadLittleMods (approved):

